PR #3, opened 2019-09-30
Feature/postgres cicd

poggdogg:
Numbering the pull request


poggdogg:
Handled the unpopulated token and null cases from Git

PR #4, opened 2019-10-04
Web api sprint 1

mardiel:
Updated to use dotnet core 2.2, and postgres 9.6 - the docker-compose should be working too.


prime-dotnet-webapi/ApiDbContext.cs
```diff
@@ -0,0 +1,201 @@
...
+                .IsUnique(false);
+
+            modelBuilder.Entity<Address>()
+                .HasIndex("EnrolleeId", "AddressType")
+                .HasName("IX_EnrolleeId_AddressType")
+                .IsUnique(true);
+            #endregion
+
+            #region Relationships
+            // modelBuilder.Entity<MailingAddress>(entity =>
```
itchison: Is there a reason for the commented out code here?

prime-dotnet-webapi/Controllers/EnrolmentController.cs
```diff
@@ -0,0 +1,181 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+
+using Prime.Models;
+using Prime.Services;
+
+namespace Prime.Controllers
+{
+    [Produces("application/json")]
+    [Route("api/v1/[controller]")]
```
itchison: This is fine for now, but we should have a bigger conversation about api and version in the pathing

prime-dotnet-webapi/ApiDbContext.cs
mardiel: This was accidentally left in the commit - I have removed it now.

prime-dotnet-webapi/Controllers/EnrolmentController.cs
mardiel: After discussion the API versioning will be handled by the routing in OpenShift, so it does not need to be in the code.  From a dotnet routing perspective, there will be no version in the endpoint URL.

PR #5, opened 2019-10-07
Enrolment workflow

mtpultz:
Initial enrolment workflow for Sprint 1


prime-angular-frontend/src/app/config/config.service.ts
```diff
@@ -0,0 +1,168 @@
...
+    return [...this.configuration.provinces];
+  }
+
+  /**
+   * Load runtime configuration.
+   *
+   * @returns {Promise<Config>}
+   * @memberof ConfigService
+   */
+  public async load(): Promise<Config> {
```
itchison: Can't wait to hook this up to services so we don't have these hardcoded values here.

PR #7, opened 2019-10-09
PRIME-93 - Setting up API round trips

mtpultz:


prime-angular-frontend/src/app/features/enrolment/shared/services/enrolment-state.service.ts
```diff
@@ -22,26 +29,37 @@ export class EnrolmentStateService {
...
-    this.contactForm.patchValue(this.getRawEnrolment().enrollee);
-    // TODO: working to populate the forms of an existing enrolment
-    // this.professionalInfoForm.patchValue(this.getRawEnrolment());
-    // this.selfDeclarationForm.patchValue(this.getRawEnrolment());
-    // this.pharmaNetAccessForm.patchValue(this.getRawEnrolment());
+  /**
+   * Store the enrolment JSON and populate the enrolment form.
+   */
+  public set enrolment(enrolment: Enrolment) {
+    this.enrolmentId = enrolment.id;
```
mardiel: I think this is missing the line:
    this.enrolleeId = enrolment.enrollee.id;
so that the enrolleeId gets send out.

PR #8, opened 2019-10-09
Feature/root user fixes

poggdogg:
We had to make some changes to handle the lack of root access at runtime.  This change updates docker files to match the need.
Also breaks out the docker-compose for easier interpretation by OpenShift CLI.
Next step after this is CI/CD.


prime-dotnet-webapi/docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+    restart: always
+    build:
+      context: .
+    environment:
+      DB_CONNECTION_STRING: "host=postgres;port=5432;database=postgres;username=postgres;password=postgres"      
+      JWT_SIGNING_KEY: "SdqlA0Y5e6HMAK1wm6iHHHz2u37XGtlo"      
+      ASPNETCORE_HTTPS_PORT: "5001"
+      ASPNETCORE_ENVIRONMENT: "Development"
+    ports:
+      - "5000:80"
```
mardiel: Will this need to map 5000 to port 8080 now, not port 80?

poggdogg: We cannot map to port 80 locally on a container.  We can redirect the public port 80 to a container on 8080 or some other high-range port but not directly to anything low-range (like 1024 or lower).  It's a bad practice from a network perspective.

prime-dotnet-webapi/Dockerfile
```diff
@@ -17,7 +17,8 @@ RUN dotnet publish -c Release -o out
 FROM mcr.microsoft.com/dotnet/core/aspnet:2.2 AS runtime
 COPY --from=build $WORKDIR .
 
-EXPOSE 8080
+EXPOSE 8080 5001
+
 USER 1001
 ENV ASPNETCORE_ENVIRONMENT Development
```
mardiel: You might want to add it here too:
ENV ASPNETCORE_URLS: http://+:8080

prime-dotnet-webapi/docker-compose.yml
mardiel: Yes, so we probably need to have the container running on port 8080, not port 80 - since OpenShift won't allow port 80.

PR #23, opened 2019-10-16
Feature/web api auditing

mardiel:


prime-dotnet-webapi/Controllers/EnrolmentsController.cs
```diff
@@ -176,13 +178,13 @@ private bool BelongsToEnrollee(Enrolment enrolment)
             //if user is not ADMIN, check that user belongs to the enrolment
             if (!belongsToEnrollee)
             {
-                var PrimeUserId = PrimeUtils.PrimeUserId(User);
-                belongsToEnrollee = PrimeUserId != null
+                Guid PrimeUserId = PrimeUtils.PrimeUserId(User);
+                belongsToEnrollee = !Guid.Empty.Equals(PrimeUserId)
```
itchison: I dislike this pattern - Is this there a better way?

prime-dotnet-webapi/Controllers/LookupsController.cs
```diff
@@ -14,8 +11,9 @@ namespace Prime.Controllers
     [Produces("application/json")]
     [Route("api/[controller]")]
     [ApiController]
-    //User needs at least the ADMIN or ENROLMENT role to use this controller
-    //[Authorize(Roles = PrimeConstants.PRIME_ADMIN_ROLE + "," + PrimeConstants.PRIME_ENROLMENT_ROLE)]
+    // User needs at least the ADMIN or ENROLMENT role to use this controller
+    // FIXME - add this back once there are OAuth tokens
```
itchison: Can we communicate with the team that we're using FIXME as something to look for?

prime-dotnet-webapi/Models/IAuditable.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+namespace Prime.Models
+{
+    public interface IAuditable
```
itchison: nice work here!

prime-dotnet-webapi/Models/BaseAuditable.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using System.ComponentModel.DataAnnotations;
+using Newtonsoft.Json;
+
+namespace Prime.Models
+{
+    public class BaseAuditable : IAuditable
```
itchison: Where do we set these values?

prime-dotnet-webapi/appsettings.json
```diff
@@ -1,7 +1,7 @@
 {
   "Jwt": {
     "Audience": "prime-web-api",
-    "WellKnown": "https://sso-dev.pathfinder.gov.bc.ca/auth/realms/v4mbqqas/.well-known/openid-configuration"
+    "WellKnown": "http://localhost:8080/auth/realms/prime/.well-known/openid-configuration"
```
itchison: Is this correct? How does this work in non localhost environments?

mardiel: I changed this back to the dev instance - it is still not hooked up yet anyway.

prime-dotnet-webapi/Models/BaseAuditable.cs
mardiel: These get set by the ApiDbContext as part of the SaveChanges and SaveChangesAsync methods.

prime-dotnet-webapi/Controllers/LookupsController.cs
mardiel: I can change these to TODO, as a standard.

prime-angular-frontend/src/app/features/enrolment/shared/services/enrolment-state.service.ts
```diff
@@ -192,7 +192,7 @@ export class EnrolmentStateService {
       hasCertification: [null, [FormControlValidators.requiredBoolean]],
       certifications: this.fb.array([]),
       isDeviceProvider: [null, [FormControlValidators.requiredBoolean]],
-      deviceProviderNumber: ['', []],
+      deviceProviderNumber: ['', [Validators.pattern('[0-9]+'), Validators.minLength(5), Validators.maxLength(5)]],
```
james-hollinger: If we can make our own validators, we should probably make an exactLenth validator of some kind. I could see this being used again.

prime-dotnet-webapi/Controllers/EnrolmentsController.cs
mardiel: I changed the order of the check to make this a little easier to read.  The issue is that Guid cannot be null, so they use a special Guid.Empty to represent that.  I also added some comments to the code to help explain it better.

prime-angular-frontend/src/app/features/enrolment/shared/services/enrolment-state.service.ts
mardiel: Yes, this was just my first cut at this - I would think Martin can do a better job of this.

PR #27, opened 2019-10-22
Added static prefix for device provider

mtpultz:


mardiel:
I have also removed the device provider number from the data model, as it is not needed.

PR #31, opened 2019-10-23
bugfix/pr-based-builds

poggdogg:
working a little more on multibranch


poggdogg:
I changed some of the parameters around in Jenkins and the templates to be more friendly to branches and master/develop.  Update your PR with these changes.

PR #35, opened 2019-10-24
Provisioner Enrolments

mtpultz:


prime-angular-frontend/src/app/modules/provision/pages/enrolments/enrolments.component.html
```diff
@@ -0,0 +1,115 @@
...
+<table mat-table
+       [dataSource]="dataSource"
+       class="w-100">
+
+  <!-- Applied Date Column -->
+  <ng-container matColumnDef="appliedDate">
+    <th mat-header-cell
+        *matHeaderCellDef> Application Date </th>
+    <td mat-cell
+        *matCellDef="let enrolment"> {{ enrolment.appliedDate | date: 'MMM d, y' }} </td>
```
itchison: Can we constant the date formatters? Doesn't have to be now, but maybe on our backlog.

prime-angular-frontend/src/app/modules/provision/shared/enums/enrolment-status.enum.ts
```diff
@@ -0,0 +1,8 @@
+export enum EnrolmentStatus {
```
itchison: Are we going to grab this from lookups?

prime-angular-frontend/src/app/modules/provision/pages/enrolments/enrolments.component.html
mtpultz: Definitely, I can make our own DatePipe and have it set to any default format easily.  The Angular DatePipe has lots of formatting options, but has no global default I can set.

mtpultz: Definitely, I can make our own DatePipe and have it set to any default format easily.  The Angular DatePipe has lots of formatting options, but has no global default I can set.

prime-angular-frontend/src/app/modules/provision/shared/enums/enrolment-status.enum.ts
mtpultz: We already do, but there is a need where the code for accepted or declined is required when updating a enrolment status.  The API takes a query parameter of status code, and I either hardcore it or use an enum.  The only other solution would be a more specific endpoint that is updateEnrolmentAccepted, updateEnrolmentDenied, etc instead of a single generic endpoint.

PR #47, opened 2019-10-29
summary

AnaisHebert:


prime-angular-frontend/src/app/modules/enrolment/shared/guards/enrolment.guard.ts
```diff
@@ -45,6 +45,15 @@ export class EnrolmentGuard implements CanActivate, CanActivateChild, CanLoad {
     return this.canActivate(next, state);
   }
 
+  public navigate(routePath: string, navPath: string) {
```
mtpultz: Guards provide a limited API of usable methods for the router so the navigate method could be private

PR #56, opened 2019-10-31
Updated country and province to be codes

mtpultz:


mtpultz:
The backend changes look good! :)

PR #57, opened 2019-10-31
Fix bug in Authentication guard

james-hollinger:
When hitting a route in our app while un-authenticated, we save your URL so the app can redirect back to your original destination after logging in. The auth guard does not handle relative urls; keycloak is appending the relative url to the keycloak base url (instead of the app base url).


james-hollinger:
Refactored the guard a little while I was here. Also added a navigation to the frontpage after keycloak login error

PR #63, opened 2019-11-01
Adding Automatic Adjudication

mardiel:


prime-dotnet-webapi/Services/DefaultAutomaticAdjudicationService.cs
```diff
@@ -0,0 +1,213 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Http;
+using Prime.Models;
+
+namespace Prime.Services
+{
+    public class DefaultAutomaticAdjudicationServiceService : BaseService, IAutomaticAdjudicationService
```
james-hollinger: Is ServiceService intentional?

prime-dotnet-webapi/Services/DefaultAutomaticAdjudicationService.cs
```diff
@@ -0,0 +1,213 @@
...
+                    // make sure there is a current status
+                    if (currentStatus == null)
+                    {
+                        throw new InvalidOperationException($"Could not process enrolment rule, current status was missing for Enrolment.Id={enrolment.Id}.");
+                    }
+
+                    // for every item returned in the results, add the reason to the current status
+                    foreach (var item in results)
+                    {
+                        if (currentStatus.EnrolmentStatusReasons == null)
```
james-hollinger: Looks like we could pull this out of the foreach loop, just for clarity.

prime-dotnet-webapi/Services/DefaultAutomaticAdjudicationService.cs
```diff
@@ -0,0 +1,213 @@
...
+
+        /// check to see if any of the addresses are outside of BC
+        private class AddressRule : BaseAutomaticAdjudicationRule
+        {
+            public override ICollection<StatusReason> ProcessRuleInternal(Enrolment enrolment)
+            {
+                var result = new List<StatusReason>(0);
+                // check to see if any of the addresses are outside of BC
+                // if there is not code (or address), use BC, so we don't get a mis-match
+                if (string.Compare(Province.BRITISH_COLUMBIA_CODE, (enrolment.Enrollee?.PhysicalAddress.ProvinceCode ?? Province.BRITISH_COLUMBIA_CODE), StringComparison.OrdinalIgnoreCase) != 0
```
james-hollinger: Perhaps for clarity it might be better to do something like:
```
var provinceCodes = new[] { enrolment.Enrollee.PhysicalAddress?.ProvinceCode, enrolment.Enrollee?.MailingAddress?.ProvinceCode };
                if (provinceCodes.Any(p => p != null && !p.Equals(Province.BRITISH_COLUMBIA_CODE, StringComparison.InvariantCultureIgnoreCase)))
                {
                    result.Add(new StatusReason { Code = StatusReason.ADDRESS_CODE });
                }
```

prime-dotnet-webapi/Services/DefaultAutomaticAdjudicationService.cs
mardiel: Oops, removed the typo.

prime-dotnet-webapi/Services/DefaultAutomaticAdjudicationService.cs
mardiel: I pulled it out of the loop - I had this both ways before as I didn't want to create it is there were no reasons, but I don't think it will really hurt to make an empty list.

prime-dotnet-webapi/Services/DefaultAutomaticAdjudicationService.cs
mardiel: Yes, this makes sense - this check it a little funky, since the data may not be present all the time.

PR #72, opened 2019-11-04
Hotfix/dotnet container not running

poggdogg:
Reverting to vanilla


poggdogg:
Only affect local dev

poggdogg:
Uppercase 'P' should have been lower on 'prime.dll'

PR #81, opened 2019-11-06
Route Guards

mtpultz:
* Restored and refactored guards to use updates to AuthService


prime-angular-frontend/src/app/core/guards/base.guard.ts
```diff
@@ -1,33 +1,86 @@
...
 import { Observable } from 'rxjs';
 
+import { LoggerService } from '@core/services/logger.service';
+import { AuthService } from '@auth/shared/services/auth.service';
+
 @Injectable({
   providedIn: 'root'
 })
-export class BaseGuard implements CanActivate, CanActivateChild, CanLoad {
+export class BaseGuard implements CanLoad, CanActivate, CanActivateChild {
```
mardiel: Should this be abstract?

PR #103, opened 2019-11-13
Basic Email Service

james-hollinger:
Basic email service, currently not used in the app by anything. This also adds a new docker image: Mailhog. This is a local email server, catching and displaying emails for testing. It has also been deployed into dev.

Also includes an unrelated bug fix for re-mapping port 8080 to allow the frontend to work both in docker & local angular serve (for keycloak redirects), and a fix to windows git preventing changing the linux script files line endings.


james-hollinger:
> 
> 
> Getting 403 as expected until 200 is handled

The testing code that generated the 403 has been removed.

PR #110, opened 2019-11-18
Adding MailHog to tools space

poggdogg:
- Created a container
- Opened port 1025, 25, web
- Need to add mail server info to the API


openshift/compositions/mailhog-deploymentconfig.yaml
```diff
@@ -0,0 +1,43 @@
...
+      creationTimestamp: null
+      labels:
+        io.kompose.service: mailhog-server
+    spec:
+      containers:
+      - image: ' '
+        name: mailhog-server
+        ports:
+        - containerPort: 1025
+        - containerPort: 1025
```
james-hollinger: Is this duplication intentional?

poggdogg: Pretty sure it is - Kompose/Kubernetes built this source template.  I think it needs it for proper mapping (I'm opening 25 and 1025 to the same port on the container)

james-hollinger: Ok, that makes sense. It is in the DC file as well, can I assume its the same there?

PR #120, opened 2019-11-20
After submit work flow

AnaisHebert:


prime-angular-frontend/src/app/modules/enrolment/pages/access-agreement/access-agreement.component.ts
```diff
@@ -116,5 +121,12 @@ export class AccessAgreementComponent implements OnInit {
           this.enrolmentStateService.enrolment = enrolment;
         }
       });
+
+    this.enrolmentService.enrolment$
```
mtpultz: Accessed as an observable, but can access the enrolment as JSON from the enrolment service

PR #140, opened 2019-11-27
HealthCheck

mtpultz:
* Added healthcheck to .NET


mtpultz:
Researched queries commonly used for health checks and the default query `"SELECT 1;"` appears to be a common starting point for simply checking the connection.  When the health check is being used in OpenShift I can check whether this is enough to alleviate the lag of the initial request on a new deployment, and build on it further if needed ie. `"SELECT COUNT(*) FROM \"Enrolment\";"`, etc.  

@poggdogg  I need a bit of direction regarding setting up the health check to run after a build.  I'm not sure if I should be using the `HEALTHCHECK` directive provided by Docker in OpenShift ie. `HEALTHCHECK CMD curl --fail http://localhost:5000/healthcheck || exit`, or just a plain curl, and whether it would live in `Dockerfile` or `openshift.dockerfile`.

PR #168, opened 2019-12-10
Adjudicator Comments

mtpultz:


prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -324,5 +328,65 @@ public async Task<ActionResult<EnrolmentStatus>> CreateEnrolmentStatus(int enrol
...
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiCreatedResponse<AdjudicatorNote>), StatusCodes.Status201Created)]
+        public async Task<ActionResult<AdjudicatorNote>> CreateAdjudicatorNote(int enrolleeId, AdjudicatorNote adjudicatorNote)
+        {
+            var enrollee = await _enrolleeService.GetEnrolleeAsync(enrolleeId);
+
+            if (enrollee == null)
+            {
+                this.ModelState.AddModelError("Enrollee", "Could not create an adjudicator note, the passed in Enrollee cannot be null.");
+                return BadRequest(new ApiBadRequestResponse(this.ModelState));
```
james-hollinger: Probaly a 404 instead of 400?

mtpultz: Yah, I thought of that when I copied these from the other endpoints.  Should we update all the endpoints to use 404 not found instead of 400.  Seems like we need to go through the current API and do some scrubbing.

mtpultz:
Hi @AnaisHebert yah definitely!  Once they decide whether they want to upload documents and we're provided with the UX we can run through and refine into some general components.

PR #177, opened 2019-12-13
Zap Scanning

poggdogg:


openshift/zap.pod.yaml
```diff
@@ -0,0 +1,44 @@
...
+    metadata: 
+      name: "${NAME}"
+      creationTimestamp: 
+      labels: 
+        run: "zap"
+        app: "${APP}"
+    spec: 
+      containers: 
+      - name: "zap"
+        image: "172.50.0.2:5000/openshift/jenkins-slave-zap"
```
itchison: What is this?

poggdogg: This is a template for a ZAP pod  provided by the pathfinder project.  We didn't end up using it - in fact I thought I deleted it.

poggdogg: Delorted...

PR #181, opened 2019-12-17
Refactoring and cleanup of enrollee controller + service

james-hollinger:
Refactorings:
- Removed random generation of enrollee Ids from enrollee service mock. Replaced with new utility method on base service mock class. Removed unnecessary constants.
- Removed a lot of unnecessary casts and unused variables from tests. Minor syntactic changes for readability.
- BelongsToEnrollee is now an extension method on ClaimsPrincipal rather than in the enrollee controller. Renamed to CanAccess.
- Added EnrolleeUserIdExistsAsync method to the enrollee service. Interface and mock have been updated. Renamed EnrolleeExists to EnrolleeExistsAsync.
- Moved CanAccess up in the execution sequence for some controller methods, generally right after checking if the enrollee object is null. Check auth as soon as possible. 
- Capitallized setAllPharmaNetStatusesFalseAsync.
- Refactored GetAvailableStatuses

Functional changes:
- Added CanAccess to CreateEnrollee. Before, enrollees could create enrollee objects with arbitrary user ids.  
- In the notes section, removed some console.writeLines and Uncommented some code that looked like they were missed. Slightly changed wording of error messages.
- Made EnrolleeExists async.

### DIFF
--- a/prime-dotnet-webapi/Extensions/ClaimsPrincipalExensions.cs
+++ b/prime-dotnet-webapi/Extensions/ClaimsPrincipalExensions.cs
@@ -2,6 +2,8 @@ using System;
 using System.Linq;
 using System.Security.Claims;
 
+using Prime.Models;
+
 namespace Prime
 {
     public static class ClaimsPrincipalExensions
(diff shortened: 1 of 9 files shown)


prime-dotnet-webapi/Extensions/ClaimsPrincipalExensions.cs
```diff
@@ -2,6 +2,8 @@
 using System.Linq;
 using System.Security.Claims;
 
+using Prime.Models;
+
 namespace Prime
 {
     public static class ClaimsPrincipalExensions
```
AnaisHebert: Is this supposed to be Extensions?

james-hollinger: Need Prime.Models because CanAccess has Enrollee as a parameter

PR #183, opened 2019-12-19
College Licence Auto-Adjudication Rule

james-hollinger:
Also some refactoring


prime-dotnet-webapi/Models/Certification.cs
```diff
@@ -44,5 +44,9 @@ public class Certification : BaseAuditable, IEnrolleeNavigationProperty
 
         [JsonIgnore]
         public Practice Practice { get; set; }
+
+        [NotMapped]
+        [JsonIgnore]
+        public string FullLicenceNumber { get { return $"{College?.Prefix}-{LicenseNumber}"; } }
```
itchison: I feel there are some inconsistencies with licence vs license in our code.

prime-dotnet-webapi/Services/DefaultEnrolleeService.cs
```diff
@@ -318,109 +318,89 @@ public Task<EnrolmentStatus> CreateEnrolmentStatusAsync(int enrolleeId, Status s
...
+
+                case Status.DECLINED_TOS_CODE:
+                    await SetAllPharmaNetStatusesFalseAsync(enrolleeId);
+                    createdEnrolmentStatus.PharmaNetStatus = true;
+                    break;
+            }
 
-                return createdEnrolmentStatus;
+            var created = await _context.SaveChangesAsync();
+            if (created < 1)
```
itchison: This reads weird doesn't it?What is greater than 1?

james-hollinger: this returns the number of records created when saving the database. It's existing code that has moved in the file

prime-dotnet-webapi/Models/Certification.cs
james-hollinger: I agree, I defaulted to the "correct" spelling of licence for Canada here without thinking. I feel like there are 3 options:

1. Change this to FullLicenseNumber to be consistent, if wrong
2. Change LicenseNumber to LicenceNumber on the Certification only
3. Change every instance of License to Licence (where grammatically appropriate)

This PR already has a fair bit of refactoring in it, so whats a little bit more? On the other hand, I would like to get this feature in, it's a high profile one. I'm leaning towards 2.

james-hollinger: I have done #1 for consistency

PR #185, opened 2019-12-20
Assign Privileges

AnaisHebert:
On acceptance of TOA, all user privileges are removed and updated with their privileges based on their college certificate licenses


prime-dotnet-webapi/Services/DefaultPrivilegeService.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            ICollection<Certification> certifications = _enrolleeDb.Certifications;
+            if (certifications != null)
+            {
+                ICollection<DefaultPrivilege> defaultPrivileges = new List<DefaultPrivilege>();
+                foreach (var cert in certifications)
+                {
+                    ICollection<DefaultPrivilege> result = this.GetDefaultPrivilegesForLicenseCode(cert.LicenseCode);
+
+                    foreach (var p in result)
```
itchison: I feel like you could filter out and then loop. But I'm ok with this approach.

AnaisHebert:
> I think it should be licence code and not license code.
> 
> In Britain and Canada, licence is the noun and license is the verb.

Okay I think we should change this in the backend in a separate PR.

PR #201, opened 2020-01-07
Change table names to singular

james-hollinger:


mtpultz:
Do we need to explicitly set the Table annotation when we're using singular?

AnaisHebert:
> Do we need to explicitly set the Table annotation when we're using singular?

The table annotations are used for the relationships as well in some way that I cant exactly remember and are necessary in some cases, so I think we should just continue to have them

PR #202, opened 2020-01-07
Dynamic Access Agreement

mtpultz:


prime-dotnet-webapi/Configuration/GlobalClauseConfiguration.cs
```diff
@@ -0,0 +1,20 @@
...
+
+public class GlobalClauseConfiguration : IEntityTypeConfiguration<GlobalClause>
+{
+    private readonly Guid SYSTEM_USER = Guid.Empty;
+
+    private readonly DateTime SEEDING_DATE = DateTime.Now;
+
+    public void Configure(EntityTypeBuilder<GlobalClause> builder)
+    {
+        #region GlobalClauseSeed
```
james-hollinger: #region is a directive for visual studio to collapse the section for easier reading, it has not effect on the code. I assume you added these lines to be consistent with the other code, but we are removing them in another PR.

mtpultz: K I'll remove them

prime-dotnet-webapi-tests/Mocks/EnrolleeServiceMock.cs
```diff
@@ -157,6 +157,17 @@ public Task<bool> IsEnrolleeInStatusAsync(int enrolleeId, short statusCodeToChec
             return Task.FromResult(false);
         }
 
+        public Task<bool> IsEnrolleeInStatusAsync(int enrolleeId, ICollection<short> statusCodesToCheck)
```
james-hollinger: Rather than make a new method that accepts a collection (in addition to the method that accepts a single status code), I would recommend updating the method signature to `public Task<bool> IsEnrolleeInStatusAsync(int enrolleeId, params short[] statusCodesToCheck)`

prime-dotnet-webapi/Configuration/GlobalClauseConfiguration.cs
mtpultz: Removed

prime-dotnet-webapi-tests/Mocks/EnrolleeServiceMock.cs
mtpultz: I was wondering if you'd say that :)  Definitely, and I can quickly refactor it's use in the EnrolleeController since that's the only place it is used

prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -453,5 +451,36 @@ public async Task<ActionResult<EnrolmentCertificateNote>> UpdateEnrolmentCertNot
...
+        public async Task<ActionResult<TermsOfAccess>> GetTermsOfAccess(int enrolleeId)
+        {
+            if (!await _enrolleeService.EnrolleeExistsAsync(enrolleeId))
+            {
+                return NotFound(new ApiResponse(404, $"Enrollee not found with id {enrolleeId}"));
+            }
+
+            // Prevent access to the enrollee's current terms of access based on status
+            var statuses = new[] { Status.IN_PROGRESS_CODE, Status.DECLINED_CODE, Status.DECLINED_TOS_CODE };
+            if (await _enrolleeService.IsEnrolleeInStatusAsync(enrolleeId, statuses))
```
james-hollinger: If the above params suggestion is implemented, this line can change to `if (await _enrolleeService.IsEnrolleeInStatusAsync(enrolleeId, Status.IN_PROGRESS_CODE, Status.DECLINED_CODE, Status.DECLINED_TOS_CODE))` if you want. Its longer but doesn't involve creating an unnecessary array first. Your call.

prime-dotnet-webapi-tests/Mocks/EnrolleeServiceMock.cs
mtpultz: Did as you suggested and updated the method to use `params`

prime-dotnet-webapi/Controllers/EnrolleesController.cs
mtpultz: Updated and added as params

prime-dotnet-webapi/Models/TermsOfAccessLicenseClassClause.cs
```diff
@@ -0,0 +1,22 @@
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+using Newtonsoft.Json;
+
+namespace Prime.Models
+{
+    [Table("TermsOfAccessLicenseClassClauseXref")]
```
james-hollinger: We are not putting "Xref" in the names of our Xref tables. I'm sure there is a coffee shop argument to be had here.

prime-dotnet-webapi/Services/DefaultEnrolleeService.cs
```diff
@@ -426,20 +434,21 @@ public bool IsStatusChangeAllowed(Status startingStatus, Status endingStatus)
...
+
             if (enrollee == null)
             {
                 return false;
             }
 
             var currentStatusCode = enrollee.CurrentStatus?.StatusCode;
 
-            return statusCodeToCheck.Equals(currentStatusCode);
+            return statusCodesToCheck.Contains(currentStatusCode.Value);
```
james-hollinger: if currentStatusCode is null, calling currentStatusCode.Value will throw an error. Its probably better to do an explicit null check and return false or do something like `var currentStatusCode = enrollee.CurrentStatus?.StatusCode ?? -1;`

prime-dotnet-webapi/Models/TermsOfAccessLicenseClassClause.cs
mtpultz: Sounds good, removed.  Nuked, paved, and pushed

prime-dotnet-webapi/Services/DefaultEnrolleeService.cs
mtpultz: Nice catch, definitely.  I like your suggestion I think I might have done a null check out of the box, but this is cleaner

prime-dotnet-webapi/Models/LicenseClassClause.cs
```diff
@@ -0,0 +1,29 @@
...
+        [Key]
+        public int Id { get; set; }
+
+        [Required]
+        public string Clause { get; set; }
+
+        [Required]
+        public DateTime EffectiveDate { get; set; }
+
+        public List<TermsOfAccessLicenseClassClause> TermsOfAccessLicenseClassClauses { get; set; }
```
james-hollinger: I think you mentioned having problems with circular references in this PR? This is why. When you JSONify your TOA, it JSONifys all the LicenceClassClauses through the xref table/navigation property, each of which have a ref back to the Xref table, which is circular. you should [JsonIgnore] the navigation property here.

mtpultz: Nice thanks James!  That was super frustrating

PR #210, opened 2020-01-08
Quick Fix for Front-end

mtpultz:


prime-angular-frontend/src/app/modules/enrolment-certificate/shared/models/enrolment-certificate.model.ts
```diff
@@ -13,4 +13,6 @@ export interface EnrolmentCertificate {
   licensePlate: string;
   privileges: Privilege[];
   organizations: Organization[];
+  // TODO should be an enum
+  organizationTypes: number;
```
james-hollinger: Err I think this is just a mistake with naming. In the backend, Organization is the xref, orginizationtypes is the lookup table.

james-hollinger: So line 15 should just be ` organizationTypes: Organization[];`
I think.

mtpultz: @james-hollinger seems like it on EnrolmentCertificate now

PR #218, opened 2020-01-09
Fixing Tests, Add Access Token Tests

james-hollinger:


prime-dotnet-webapi-tests/Mocks/EnrolleeServiceMock.cs
```diff
@@ -146,24 +146,12 @@ public bool IsStatusChangeAllowed(Status startingStatus, Status endingStatus)
             return availableStatuses.Contains(endingStatus);
         }
 
-        public Task<bool> IsEnrolleeInStatusAsync(int enrolleeId, short statusCodeToCheck)
+        public Task<bool> IsEnrolleeInStatusAsync(int enrolleeId, params short[] statusCodesToCheck)
```
hogmann: Been trying to search why this is with no avail, but what does `params` do in the case of passing things into methods?

PR #219, opened 2020-01-09
tweaking templates for prod readiness

poggdogg:


poggdogg:
To properly test, I need to progress through all environments.

PR #236, opened 2020-01-15
Privilege Types

AnaisHebert:
Add Privilege Type lookup to groups
change PrivilegeGroup to lookup table
Send both to front end with lookups
Separate privileges on front end based on privilege type for both certificates


docker-compose.yml
```diff
@@ -37,20 +37,20 @@ services:
...
-    restart: always
-    ports:
-      - "8080:8080"
-      - "4200:8080"
-    expose:
-      - "8080"
-      - "4200"
-    networks:
-      - primenet
+  # nginx-angular:
```
james-hollinger: I'm going to assume this was accidentally committed :P

AnaisHebert: oops yup, let me remove that

prime-dotnet-webapi/Models/EnrolmentCertificate.cs
```diff
@@ -35,9 +36,17 @@ public static EnrolmentCertificate Create(Enrollee enrollee)
...
+                EnrolmentCertificateNote = enrollee.EnrolmentCertificateNote,
+                Transactions = enrollee.AssignedPrivileges
+                                .Select(ap => ap.Privilege)
+                                .Where(p => p.PrivilegeGroup.PrivilegeTypeCode == 2),
+                UserType = enrollee.AssignedPrivileges
+                            .Select(ap => ap.Privilege)
+                            .Where(p => p.PrivilegeGroupCode == 4).FirstOrDefault(),
+                CanHaveOBOs = enrollee.AssignedPrivileges
+                            .Select(ap => ap.Privilege)
+                            .Where(p => p.PrivilegeGroupCode == 5).FirstOrDefault(),
```
james-hollinger: Don't really like the magic numbers here, but its probably fine.
There isn't an obvious ordering for these last two, I assume SingleOrDefault() is more appropriate.

AnaisHebert: K changed it to SingleorDefault. Agree about the 'magic numbers'. Do you have a suggestion for a better way?

AnaisHebert: Prime Constant?

james-hollinger: You could do for example what we've done with the status reasons with some named constants on the model. That's still not a flawless solution since those have to be maintained if the data changes but it at least names the magic numbers.

AnaisHebert: k I made that change!

prime-dotnet-webapi/Models/PrivilegeType.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+using Newtonsoft.Json;
+
+namespace Prime.Models
+{
+    [Table("PrivilegeTypeLookup")]
+    public class PrivilegeType : BaseAuditable, ILookup<short>
+    {
+        public readonly static short PHARMANET_TRANSACTIONS_TYPE = 2;
```
james-hollinger: A little nit-picky, but might as well drop the _TYPE from this. The use of it will read PrivilegeType.PHARMANET_TRANSACTIONS_TYPE which is redundant. 
Same for the group constants.

prime-dotnet-webapi/Models/EnrolmentCertificate.cs
```diff
@@ -35,9 +36,17 @@ public static EnrolmentCertificate Create(Enrollee enrollee)
...
+                EnrolmentCertificateNote = enrollee.EnrolmentCertificateNote,
+                Transactions = enrollee.AssignedPrivileges
+                                .Select(ap => ap.Privilege)
+                                .Where(p => p.PrivilegeGroup.PrivilegeTypeCode == PrivilegeType.PHARMANET_TRANSACTIONS_TYPE),
+                UserType = enrollee.AssignedPrivileges
+                            .Select(ap => ap.Privilege)
+                            .Where(p => p.PrivilegeGroupCode == PrivilegeGroup.USER_TYPE_GROUP).SingleOrDefault(),
+                CanHaveOBOs = enrollee.AssignedPrivileges
+                            .Select(ap => ap.Privilege)
+                            .Where(p => p.PrivilegeGroupCode == PrivilegeGroup.CAN_HAVE_OBOS_GROUP).SingleOrDefault(),
```
james-hollinger: A little nit-picky, but might as well drop the _TYPE and _GROUP from these constants. It's a little redundant.

PR #243, opened 2020-01-16
Limits Conditions Clauses from notes

AnaisHebert:
* Change TermsofAccess to AccessTerm in models and tables for pluralization reasons
* Change LimitsAndConditions to LimitsConditions
* Limits and conditions no longer many to many
* At request for TOA, get most recent note for limits and conitions from adjudicator, and copy value to limits and conditions clause for TOA


prime-dotnet-webapi/ApiDbContext.cs
```diff
@@ -42,11 +42,11 @@ public ApiDbContext(DbContextOptions<ApiDbContext> options, IHttpContextAccessor
         public DbSet<AccessAgreementNote> AccessAgreementNotes { get; set; }
         public DbSet<EnrolmentCertificateNote> EnrolmentCertificateNotes { get; set; }
 
-        public DbSet<TermsOfAccess> TermsOfAccess { get; set; }
+        public DbSet<AccessTerm> AccessTerm { get; set; }
```
james-hollinger: Should this be pluralized to AccessTerms?

AnaisHebert: yes, good catch

PR #267, opened 2020-01-20
Feature/stop ie

hogmann:
require the banner to be designed.


hogmann:
To run locally in a mode for ie11 to work run `ng serve --configuration es5`

PR #283, opened 2020-01-23
Static TOA (RU/OBO)

mtpultz:


prime-angular-frontend/src/app/modules/enrolment/pages/access-agreement/access-agreement.component.ts
```diff
@@ -144,7 +154,7 @@ export class AccessAgreementComponent extends BaseEnrolmentPage implements OnIni
   public ngOnInit() {
     this.enrolment = this.enrolmentService.enrolment;
     this.isInitialEnrolment = this.enrolment.progressStatus !== ProgressStatus.FINISHED;
-    this.enrolmentResource.getTermsOfAccess(this.enrolment.id)
+    this.enrolmentResource.getAccessTerm(this.enrolment.id)
       .subscribe(
         (termsOfAccess: AccessTerm) => this.termsOfAccess = termsOfAccess,
```
hogmann: might be out of scope but this line hurts me as it uses an old variable that matches the old model name.

PR #287, opened 2020-01-24
Feature/expiry date renewal

hogmann:
Added migrations for expiry date in the database to the access term table as well as propagating it to the front end and displaying it on the provisioner link and the pharmanet page.

as well as starting a README and throwing somethings inside for API


prime-dotnet-webapi/Services/AccessTermService.cs
```diff
@@ -64,12 +64,14 @@ public async Task CreateEnrolleeAccessTermAsync(Enrollee enrollee)
 
         public async Task SetAcceptedDateForAccessTermAsync(Enrollee enrollee)
         {
-            var termsOfAccess = await _context.AccessTerms
+            var accessTerm = await _context.AccessTerms
                 .Where(toa => toa.EnrolleeId == enrollee.Id)
                 .OrderByDescending(toa => toa.AcceptedDate)
                 .FirstOrDefaultAsync();
```
james-hollinger: If accessTerm is null, this method blows up on the next line. This should either have a null guard here or this could be FirstAsync() so it throws if there are no access terms. Theres a coffee shop argument here about weather it should silently return or if it should throw. Personally, I feel that calling this method on an enrollee with no TOAs is a programmer error, not a user or run-time error, so I would lean toward this failing loudly not silently.

prime-dotnet-webapi/Services/AccessTermService.cs
```diff
@@ -64,12 +64,14 @@ public async Task CreateEnrolleeAccessTermAsync(Enrollee enrollee)
 
         public async Task SetAcceptedDateForAccessTermAsync(Enrollee enrollee)
```
james-hollinger: Probably should rename this method, it doesn't do what it says any more. Perhaps something like AcceptCurrentAccessTerm

prime-dotnet-webapi/Services/AccessTermService.cs
```diff
@@ -64,12 +64,14 @@ public async Task CreateEnrolleeAccessTermAsync(Enrollee enrollee)
...
-            var termsOfAccess = await _context.AccessTerms
+            var accessTerm = await _context.AccessTerms
                 .Where(toa => toa.EnrolleeId == enrollee.Id)
                 .OrderByDescending(toa => toa.AcceptedDate)
                 .FirstOrDefaultAsync();
 
-            termsOfAccess.AcceptedDate = DateTime.Now;
+            accessTerm.AcceptedDate = DateTime.Now;
+            // Add an Expiry Date of one year in the future.
+            accessTerm.ExpiryDate = DateTime.Now.AddYears(1);
```
james-hollinger: What should happen if this method gets called when the most recent access term is already accepted? Programatically this makes total sense but accepting the TOA is akin to signing a legal document so changing this date seems off to me.

hogmann: So the case is that a user has accepted a the most recent TOA, and somehow a user receives a TOA that was created before that one and accepts it?

Are you alluding to we should check that if a accessTerm already has a accepted + expiry date we should not override them?

prime-dotnet-webapi/Services/AccessTermService.cs
hogmann: also i just changed the working to reflect the new accessTerm but can do a quick change for the var from toa -> at so that it doesn't look off

prime-dotnet-webapi/Services/AccessTermService.cs
james-hollinger: This method could (theoretically) be called from anywhere in our code. It _currently_ is only called when the enrollee goes to the ACCEPTED_TOA status, but even then we aren't checking that they actually have a non-accepted TOA to "sign", we are 100% trusting in the status change process. Which feels a little iffy to me when it represents signing a legal document.

I feel like if the newest TOA has an accepted date (is already signed) this should be another loud error.

hogmann: Alright i see where you are coming from. Should i move those thoughts into a technical backlog tickets as that seems out of the scope of this ticket. Adding an expiry date and having it displayed on the front end. This would be more changing the logic on how the code handles the TOA.

hogmann: https://bcgovmoh.atlassian.net/browse/PRIME-320?atlOrigin=eyJpIjoiNDdkY2JmMTZmNjMyNDRjZTlhZWRhN2FiMTU2MzhkMWEiLCJwIjoiaiJ9 for reference

PR #314, opened 2020-01-30
Email Provisioner Access Link

james-hollinger:


prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -7,33 +10,100 @@ namespace Prime.Services
 {
     public class EmailService : BaseService, IEmailService
     {
-        public EmailService(
-            ApiDbContext context, IHttpContextAccessor httpContext)
+        private const string PRIME_EMAIL = "prime@gov.bc.ca";
```
itchison: I'm 50-50 if this should be an env variable

james-hollinger: Would you want it to be different between dev/test/prod?

itchison: No, and it's why I'm 50-50. But this email is highly likely to change.

PR #317, opened 2020-01-31
Reduce Info on Provisioner Link Certificate

AnaisHebert:


prime-dotnet-webapi/Models/EnrolmentCertificate.cs
```diff
@@ -20,9 +20,6 @@ public sealed class EnrolmentCertificate
         public IEnumerable<OrganizationType> OrganizationTypes { get; set; }
 
         public EnrolmentCertificateNote EnrolmentCertificateNote { get; set; }
```
james-hollinger: this line should also be removed I assume

AnaisHebert: oh yep

PR #339, opened 2020-02-06
Template tweaks and HTTPS everywhere

poggdogg:


poggdogg:
Fixed a bunch of edge-case OC stuff (immutable errors).

Added secure URLs everywhere

PR #347, opened 2020-02-07
Feature/local api docker

hogmann:


prime-dotnet-webapi/entrypoint.sh
```diff
@@ -1,7 +1,18 @@
 #!/bin/bash
 echo "Running the migrations..."
 #psql -d postgres -f databaseMigration.sql
+
+PG_IS_READY=$(pg_isready -h $DB_HOST -U ${POSTGRESQL_USER} -d ${POSTGRESQL_DATABASE})
+until [[ $PG_IS_READY == *"$DB_HOST"* && $PG_IS_READY == *"accepting connections"* ]]
```
itchison: It'd be great if we had an exit after X number of attempts.

hogmann: That would probably be best, can add that in.

PR #370, opened 2020-02-12
enrolment history for a given access term

hogmann:


prime-angular-frontend/src/app/modules/enrolment/pages/access-terms/access-terms.component.html
```diff
@@ -64,7 +78,7 @@
         <span>View Terms of Access</span>
       </button>
       <button mat-menu-item
-              [routerLink]="['enrolments', accessTerm.id]">
+              [routerLink]="[accessTerm.id, 'enrolment']">
```
mtpultz: If you use the BaseEnrolmentPage you can use EnrolmentRoutes in the template so we don't have to find strings of the route in the templates if the routes change.

prime-angular-frontend/src/app/modules/enrolment/pages/access-agreement-history/access-agreement-history.component.ts
```diff
@@ -29,6 +30,10 @@ export class AccessAgreementHistoryComponent implements OnInit {
     this.getAccessTerm();
   }
 
+  public routeTo() {
+    this.router.navigate(['../'], { relativeTo: this.route.parent });
```
mtpultz: If you use the BaseEnrolmentPage you can use get routeTo for free through inheritance

PR #381, opened 2020-02-13
Feature/login page vii

hogmann:


hogmann:
> 👍 Should the link for how to set up a mobile card go somewhere?

I think you asked me this on a past version, and yes it should go somewhere. But no one has told me where yet.

PR #383, opened 2020-02-14
Clauses Preview On Load

AnaisHebert:


prime-angular-frontend/src/app/modules/adjudication/pages/limits-conditions-clauses/limits-conditions-clauses.component.html
```diff
@@ -127,7 +127,7 @@
     <div class="row mb-4">
       <div class="col-sm-12 col-lg-6">
         <editor class="mb-4"
-                initialValue=""
+                #editor
```
mtpultz: Doesn't look like this is used in the controller

PR #404, opened 2020-02-19
Feature/email fail gracefully back end

hogmann:


prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -292,6 +295,20 @@ public async Task<ActionResult<EnrolmentStatus>> CreateEnrolmentStatus(int enrol
 
             var enrolmentStatus = await _enrolleeService.CreateEnrolmentStatusAsync(enrolleeId, status);
 
+
+            // Enrollee just left manual adjudication, inform the enrollee
+            if (enrollee.CurrentStatus.StatusCode == Status.UNDER_REVIEW_CODE)
```
james-hollinger: Does this work? I would expect the current status to be updated. You would need to capture the old status before calling CreateEnrolmentStatusAsync.

james-hollinger: Does this work? I would expect the current status to be updated already on line 296. You would need to capture the old status before calling CreateEnrolmentStatusAsync.

hogmann: valid point, i will double check this.

PR #420, opened 2020-02-21
Business Events database and status changes captured

AnaisHebert:


prime-dotnet-webapi/Models/BusinessEvent.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.ComponentModel.DataAnnotations.Schema;
+using System.Text.Json.Serialization;
+
+namespace Prime.Models
+{
+    [Table("BusinessEvent")]
+    public class BusinessEvent : BaseAuditable
+    {
+        [Key]
+        public int? Id { get; set; }
```
james-hollinger: Don't think we should have nullable indexes any more

AnaisHebert: oh yeah good catch

prime-dotnet-webapi/Models/BusinessEvent.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public int EnrolleeId { get; set; }
+
+        [JsonIgnore]
+        public Enrollee Enrollee { get; set; }
+
+        public int? AdminId { get; set; }
```
james-hollinger: Admin Id is not nullable

james-hollinger: Wait you're right, not all events have Admin Ids

PR #441, opened 2020-03-02
API Resource

mtpultz:
* Added API resource service
* Added API HTTP response model
* Added API HTTP error response model
* Updated error handler and interceptor


poggdogg:
ERROR in src/test/mocks/mock-config.service.ts:15:19 - error TS2345: Argument of type 'HttpClient' is not assignable to parameter of type 'ApiResource'.
--
  | Type 'HttpClient' is missing the following properties from type 'ApiResource': config, logger, http, handleResponse, and 2 more.
  |  
  | 15     super(config, http);
  | ~~~~

PR #448, opened 2020-03-03
Feature/admin view only

hogmann:


prime-angular-frontend/src/app/modules/adjudication/pages/enrolments/enrolments.component.ts
```diff
@@ -73,10 +75,6 @@ export class EnrolmentsComponent implements OnInit {
     return (currentStatusCode === EnrolmentStatus.UNDER_REVIEW);
   }
 
-  public isSuperAdmin(): boolean {
```
AnaisHebert: ERROR in src/app/modules/adjudication/pages/enrolments/enrolments.component.html(178,15): Property 'isSuperAdmin' does not exist on 

this is till used on the template i believe

hogmann: Whoops, this is why I'm supposed to build prod after changes

prime-dotnet-webapi/AuthenticationSetup.cs
```diff
@@ -90,8 +90,10 @@ IHostEnvironment environment
             services.AddAuthorization(options =>
             {
                 options.AddPolicy(PrimeConstants.USER_POLICY, policy => policy.Requirements.Add(new PrimeUserRequirement()));
+                //options.AddPolicy(PrimeConstants.ACCESS_POLICY, policy => policy.RequireRole().a)
```
AnaisHebert: Should this commented line be here?

hogmann: It should not be there.

PR #451, opened 2020-03-03
Revamp API Status Codes

james-hollinger:
API response had status codes in the body, which is unnecessary. Refactored API response models for simplicity.


hogmann:
No migrations is what I like to see!

PR #460, opened 2020-03-04
Fix Lookup Service

james-hollinger:


prime-dotnet-webapi/ApiDbContext.cs
```diff
@@ -233,31 +233,4 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             #endregion
         }
     }
-
-    public static class ApiDbContextExtensions
-    {
-        public static IQueryable Set(this DbContext context, Type T)
```
hogmann: But HOW WITH WE SETTTTTTTTT!!! oh wait it already has this method... carry on.

PR #483, opened 2020-03-10
Refactoring status

james-hollinger:
_Lots of stuff_


prime-dotnet-webapi-tests/ModelFactories/EnrolleeFactory.cs
```diff
@@ -56,13 +56,13 @@ public EnrolleeFactory()
             RuleFor(x => x.EnrolleeProfileVersions, f => null);
             Ignore(x => x.AccessTerms); // Awaiting finialization of access terms
 
-            RuleSet("status.under_review", (set) =>
+            RuleSet("status.submitted", (set) =>
```
AnaisHebert: Why do we have submitted instead of under review here?

prime-dotnet-webapi/Services/BusinessEventService.cs
```diff
@@ -71,8 +76,12 @@ public async Task<BusinessEvent> CreateAdminClaimEventAsync(int enrolleeId, stri
         }
 
 
-        private BusinessEvent CreateBusinessEvent(int BusinessEventTypeCode, int enrolleeId, string description, int? adminId = null)
+        private async Task<BusinessEvent> CreateBusinessEvent(int BusinessEventTypeCode, int enrolleeId, string description)
```
AnaisHebert: Is every business event adding an admin id now?

prime-dotnet-webapi-tests/ModelFactories/EnrolleeFactory.cs
james-hollinger: This creates enrollees with a certain set of statuses, rather than just one status. In this case, it's "active" then "under review". I called it "submitted" to try to distinguish it from the name of a status, but its not a perfect naming.

prime-dotnet-webapi/Services/BusinessEventService.cs
james-hollinger: The method pulls the userId from the JWT token via the HTTP context and looks up if they are an admin. If the creator of the event is an admin, it adds the admin ID, else its null. I wanted to avoid having to pass down admin Ids everywhere, since most services don't actually need that info to do the action.

AnaisHebert: Okay thats awesome, just wanted to make sure the same functionality was there. That is much nicer in that case!

james-hollinger: It will mean that we cant deliberately choose not to add the admin Id to certain events caused by admin users, but I don't think we have any need for that currently.

PR #493, opened 2020-03-14
Event Update

mtpultz:
* Added Enrollee Created event
* Updated Disclaim event
* Changed location of Submitted event

### DIFF
--- a/prime-dotnet-webapi/Services/SubmissionService.cs
+++ b/prime-dotnet-webapi/Services/SubmissionService.cs
@@ -79,6 +79,7 @@ namespace Prime.Services
             enrollee.AddEnrolmentStatus(StatusType.UnderReview);
 
             await _enroleeProfileVersionService.CreateEnrolleeProfileVersionAsync(enrollee);
+            await _businessEventService.CreateStatusChangeEventAsync(enrollee.Id, "Submitted");
 
             if (await _automaticAdjudicationService.QualifiesForAutomaticAdjudication(enrollee))
             {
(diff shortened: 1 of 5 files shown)


prime-dotnet-webapi/Services/SubmissionService.cs
```diff
@@ -120,8 +120,11 @@ private async Task ProccessToaAsync(Enrollee enrollee, bool accept)
                 await _accessTermService.AcceptCurrentAccessTermAsync(enrollee);
                 await _privilegeService.AssignPrivilegesToEnrolleeAsync(enrollee.Id, enrollee);
                 await _businessEventService.CreateStatusChangeEventAsync(enrollee.Id, "Accepted TOA");
-                await _enrolleeService.UpdateEnrolleeAdjudicator(enrollee.Id);
-                await _businessEventService.CreateAdminClaimEventAsync(enrollee.Id, "Admin disclaimed after TOA accepted");
+                if (enrollee.AdjudicatorId !== null)
+                {
+                    await _enrolleeService.UpdateEnrolleeAdjudicator(enrollee.Id);
+                    await _businessEventService.CreateAdminClaimEventAsync(enrollee.Id, "Admin disclaimed after TOA accepted");
```
james-hollinger: Should we be raising this event at all? its not really a business event that someone was automatically removed

mtpultz: Yah good question.  I don't know.  I assume it was asked for as an event last sprint.  I'm just updating it since it's triggered even when the enrollee hasn't been claimed by an adjudicator.

mtpultz: @itchison what do you think do we need this business event?

PR #495, opened 2020-03-17
Health Checks

mtpultz:
* Turned on health checks


mtpultz:
@poggdogg I've uncommented the database connection and dbcontext health checks for this branch

PR #513, opened 2020-03-19
PRIME-538 - Updated checkbox label

mtpultz:


itchison:
Please update with jira ticket number

PR #519, opened 2020-03-19
PRIME-557 Auto Adjudicate RN LPN RPN + fix backend tests

AnaisHebert:


prime-dotnet-webapi-tests/Mocks/EmailServiceMock.cs
```diff
@@ -14,7 +14,7 @@ public override void SeedData()
             // no data to seed, as it is done in the base class
         }
 
-        public Task SendProvisionerLinkAsync(string provisionerEmail, EnrolmentCertificateAccessToken token, string ccEmail)
+        public Task SendProvisionerLinkAsync(string provsionerName, string provisionerEmail, EnrolmentCertificateAccessToken token, string ccEmail)
```
mtpultz: Fnoud a tpoy on provisionerName

prime-dotnet-webapi/Migrations/20200319180139_ValidateLicenceFlag.cs
```diff
@@ -0,0 +1,331 @@
...
+namespace Prime.Migrations
+{
+    public partial class ValidateLicenceFlag : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DeleteData(
+                table: "DefaultPrivilege",
+                keyColumns: new[] { "PrivilegeId", "LicenseCode" },
+                keyValues: new object[] { 18, 16 });
```
james-hollinger: Could you make this an UpdateData rather than a drop and re-add?

AnaisHebert: It is removing one privilege and replacing it with another so i dont think it needs to be an update in this case.

prime-dotnet-webapi/Services/AutomaticAdjudicationService.cs
```diff
@@ -134,38 +134,41 @@ protected override async Task<bool> ProcessRuleInternal(Enrollee enrollee)
 
                 foreach (var cert in enrollee.Certifications)
                 {
-                    PharmanetCollegeRecord record = null;
-                    try
+                    if (cert.License.Validate)
```
james-hollinger: Could you please flip the logic and make this
`if(!cert.Licencse.Validate)
{
    continue;
}`
it makes it easier to read in my opinion

james-hollinger: or, alternatively, you could use
`foreach (var cert in enrollee.Certifications.Where(c => c.License.Validate))`

AnaisHebert: > or, alternatively, you could use
> `foreach (var cert in enrollee.Certifications.Where(c => c.License.Validate))`

I like this better the other ways feels awkward to me

PR #527, opened 2020-03-23
PRIME-567-most-recent-toa

hogmann:
attached to the enrollee object that the enrollee has signed the most recent ToA when the call is made for the admin side of thing. Not sure what verbage to use so just made it super simple for now.


prime-dotnet-webapi/Models/Enrollee.cs
```diff
@@ -178,6 +181,7 @@ public IEnumerable<ValidationResult> Validate(ValidationContext validationContex
 
         public bool AlwaysManual { get; set; }
 
+
```
james-hollinger: ??

PR #534, opened 2020-03-25
PRIME-593 - Infographic to next steps page

hogmann:
The svg are not in the right format for it to suite responsive. Will need some more made to better suite it. Will make not in ticket and stand-up

They wanted to try and sneak this in for tomorrow, works but info graphic is not fully responsive at some resolutions due to the svg's  being wonky.


prime-angular-frontend/src/app/modules/enrolment/pages/pharmanet-enrolment-summary/pharmanet-enrolment-summary.component.ts
```diff
@@ -47,7 +47,7 @@ export class PharmanetEnrolmentSummaryComponent extends BaseEnrolmentPage implem
     private fb: FormBuilder
   ) {
     super(route, router);
-    this.showProgressBar = false;
+    this.showProgressBar = true;
```
mtpultz: I don't think we want this to default show the progress bar.  The progress bar is only used on initial enrolment and in this case only displayed after you accept a TOA to show 100%.  It gets set based on route state data in ngOnInit that is only passed from the access agreement view.  By setting it in the constructor it is possible you'll see it a bounce in the screen content as it appears and then disappears when the view is revisited post acceptance.

prime-angular-frontend/src/app/modules/enrolment/pages/pharmanet-enrolment-summary/pharmanet-enrolment-summary.component.html
```diff
@@ -2,31 +2,14 @@
...
-        </ng-container>
-        <ng-container #alertContent
-                      class="alert-content">
-          Renew your PRIME enrolment by {{ enrolment?.expiryDate | formatDate }}
-        </ng-container>
-      </app-alert>
-    </div>
+    <app-page-header>Next Steps to get PharmaNet</app-page-header>
+
+    <app-progress-indicator [currentRoute]="EnrolmentRoutes.PHARMANET_ENROLMENT_SUMMARY"></app-progress-indicator>
```
mtpultz: Looks like the `ngIf` was removed from the progress indicator

PR #535, opened 2020-03-25
PRIME-615 - Updated colleges to prevent multiples

mtpultz:


prime-angular-frontend/src/app/modules/enrolment/pages/regulatory/regulatory.component.ts
```diff
@@ -32,19 +28,20 @@ export class RegulatoryComponent extends BaseEnrolmentProfilePage implements OnI
...
-    this.licenses = this.configService.licenses;
   }
 
   public get certifications(): FormArray {
     return this.form.get('certifications') as FormArray;
   }
 
+  public get selectedCollegeCodes(): number[] {
+    return this.certifications.value
+      .map((certification: CollegeCertification) => +certification.collegeCode);
```
hogmann: I think i asked somewhere else but what does that "+" do?

mtpultz: @hogmann it forces a cast to integer so TypeScript types don't get angry, but should only be used when you absolutely know that what you're getting is a number in a string, otherwise you'll get NaN.

PR #548, opened 2020-03-28
PRIME-514 - Angular 9

mtpultz:


prime-angular-frontend/Dockerfile
```diff
@@ -19,37 +19,36 @@ ENV KEYCLOAK_CLIENT_ID $KEYCLOAK_CLIENT_ID
...
-    ng build --prod && \ 
+RUN npm install @angular/cli -g && \
+    npm install && \
+    npm audit fix && \
+    ng build --prod && \
     echo "NPM packages installed..."
 
-# RUN npm audit fix --only=prod && \ 
-#     npm install @angular-devkit/build-angular@0.803.24 \ 
+# RUN npm audit fix && \
```
hogmann: Not sure if this is needed commented out here as we are including it above.

prime-angular-frontend/openshift.dockerfile
```diff
@@ -36,15 +36,14 @@ COPY nginx.template.conf /etc/nginx/nginx.template.conf
...
+RUN cat /usr/src/app/src/environments/environment.prod.ts && \
+    npm install @angular/cli  -g --silent && \
+    npm install && \
+    npm audit fix && \
+    ng build --prod && \
     echo "NPM packages installed..."
 
-# RUN npm audit fix --only=prod && \ 
-#     npm install @angular-devkit/build-angular@0.803.24 \ 
+# RUN npm audit fix --only=prod && \
```
hogmann: Same here

prime-angular-frontend/Dockerfile
mtpultz: Yah, probably not, but it's Paul's code and I don't know if he had something in mind when he commented it out.  @poggdogg do we need this commented out code in the Dockerfile?

prime-angular-frontend/openshift.dockerfile
mtpultz: @poggdogg do we need this commented out code in the Dockerfile?

mtpultz:
@poggdogg I've updated the Dockerfile and OpenShift template.  I've removed our temporary patch and turned on audit fix in the other environments since I'd rather find out there's an issue with it updating something before we push to prod, or even test.

prime-angular-frontend/Dockerfile
poggdogg: Was commented out during troubleshooting of angular 8 - please delete.

PR #550, opened 2020-03-31
PRIME-609-MongoDB

AnaisHebert:
Mongo DB storing 'Locations' collection. 
Migrations running at startup.
API.
Store EnrolleeProfileVersion history in both databases.


prime-dotnet-webapi-tests/Utils/TestUtils.cs
```diff
@@ -16,6 +16,7 @@
 using System.Net.Http.Headers;
 using Microsoft.EntityFrameworkCore.Internal;
 using Newtonsoft.Json.Linq;
+using Prime.Auth;
```
james-hollinger, line 19: This line is already in the file, line 10

prime-dotnet-webapi/Models/MongoModels/Locations.cs
```diff
@@ -0,0 +1,20 @@
+using Mongo.Migration.Documents;
+using Mongo.Migration.Documents.Attributes;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+
+namespace Prime.Models
+{
+    [StartUpVersion("0.0.2")]
+    [CollectionLocation("Locations", "prime")]
+    public class Locations : IDocument
```
james-hollinger, line 10: We have a site model called "location" now, so this naming is unfortunate. Is it possible to re-name this model?
Could we also change the namespace to something to make clear this is a POC model?

prime-dotnet-webapi/Services/MongoServices/LocationService.cs
```diff
@@ -0,0 +1,40 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Options;
+using MongoDB.Driver;
+using Prime.Models;
+using Prime.Services;
+
+namespace Prime.Services
+{
+    public class LocationService : ILocationService
```
itchison, line 10: I am a little worried about the "POC" implementation coming into our code like this. It's not live or being used. There is a chance we move this into a different parent subfolder (not in prime-dotnet-webapi) as well. So it's a little early to have these implementation items here. I like that we took the POC to the point of getting data to work via our services.. but maybe that could stay behind in a branch instead.

AnaisHebert: Do you want me to remove the insert of the EnrolleeProfileVersion into the Mongo database as well?

AnaisHebert:
> Seems like the tests are broken on your branch; they never complete when run.

Should be working now

PR #554, opened 2020-04-01
PRIME-148 Adding document manager

poggdogg:
Initial attempt


poggdogg:
Yes, I've been asking devs to take this for a spin to see if there was anything I missed before proceeding with the openshift implementation.

Does it work locally for you? Is there any other functionality required?

itchison:
I still see some left overs from MDS/Core in the code base. Also, looks like it might have been lifted directly from MDS and not other repo. I'm also curious if there's a way for us to bring this in so that we can keep in synch with the other repo. ie - if they fix something we can then also receive that fix.

For now, looks good and is a great first step.

poggdogg:
Paul D was talking about making it a separate repo once it's polished.  I'll ping him about this.
For now I'll get to work ont the openshift template

PR #565, opened 2020-04-13
PRIME-652 feature flagging

hogmann:


james-hollinger:
Added roles to dev and test keycloak

PR #575, opened 2020-04-14
PRIME-698 - Email Razor Port

mtpultz:


prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -140,67 +163,16 @@ private async Task Send(string from, IEnumerable<string> to, IEnumerable<string>
...
-            body += "Thank you,<br><br>";
-
-            body += "PRIME<br><br>";
-            //TODO: Is this stored as an attribute in the backend?
-            body += "1-844-39PRIME<br><br>";
-            //TODO: Is this stored as an attribute in the backend?
-            body += "<a href='mailto:PRIMEsupport@gov.bc.ca' target='_top'>PRIMEsupport@gov.bc.ca</a>";
-
-            return body;
+            EmailProps emailProps = new EmailProps(enrollee.FirstName, enrollee.LastName, token.FrontendUrl, _certificateService.GetMaxViews(), _certificateService.GetExpiryDays());
```
james-hollinger, line 174: These props are created identically in these two methods, can just move them out into the send provisioner link method

james-hollinger: When you do that, these metods just pass the props into the razor converter, so these methods are a little unnessisary

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -8,18 +8,41 @@
...
 namespace Prime.Services
 {
+    public class EmailProps
+    {
+        public string FirstName;
+        public string LastName;
+        public string TokenUrl;
+        public int MaxViews;
+        public int ExpiryDays;
+        public string ProvisionerName;
```
james-hollinger, line 18: I dont know if razor requires these to be fields vs properties, but I would prefer to see these as properties

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -8,18 +8,41 @@
 
 namespace Prime.Services
 {
+    public class EmailProps
```
james-hollinger, line 11: Props is a rather Javascript term :P
Razor wants to render a _model_ even though we are really just smushing arguments into an object. How about something like EmailBody?

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -8,18 +8,41 @@
...
+    {
+        public string FirstName;
+        public string LastName;
+        public string TokenUrl;
+        public int MaxViews;
+        public int ExpiryDays;
+        public string ProvisionerName;
+
+        public EmailProps(string firstName, string lastName, string tokenUrl, int maxViews, int expiryDays, string provisionerName = null)
```
james-hollinger, line 20: why not pass in the access token and get the props from the enrollee, rather then pass them in individually?

james-hollinger: or a static generation method FromToken or somesuch

prime-dotnet-webapi/Services/EmailService.cs
mtpultz: The only difference is there is no provisionerName in one of them

prime-dotnet-webapi/Services/EmailService.cs
mtpultz: EmailBody suggests structure and these are just parameters for binding.  What about EmailParams?

prime-dotnet-webapi/Services/EmailService.cs
mtpultz: Doesn't matter just need a model for binding so I can update these to be properties.

prime-dotnet-webapi/Services/EmailService.cs
mtpultz: Yep, I can make it take a token as a param instead.  I literally copy pasted the properties to make the constructor since that wasn't the point of the ticket.

prime-dotnet-webapi/Services/EmailService.cs
mtpultz: I've updated it to be EmailParams

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -8,18 +8,37 @@
...
     {
         private const string PRIME_EMAIL = "no-reply-prime@gov.bc.ca";
-        private readonly IEnrolmentCertificateService _certificateService;
+        private readonly IRazorConverterService _razorConverterService;
 
         public EmailService(
             ApiDbContext context,
             IHttpContextAccessor httpContext,
-            IEnrolmentCertificateService enrolmentCertificateService)
+            IEnrolmentCertificateService enrolmentCertificateService,
```
james-hollinger, line 37: Missed this line

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -71,8 +91,8 @@ public async Task SendProvisionerLinkAsync(string[] recipients, EnrolmentCertifi
 
             string subject = "New Access Request";
             string emailBody = (string.IsNullOrEmpty(provisionerName))
-                ? this.GetClinicManagerEmailBody(token.Enrollee, token)
-                : this.GetVendorEmailBody(token.Enrollee, token, provisionerName);
+                ? await _razorConverterService.RenderViewToStringAsync("/Views/Emails/VendorEmail.cshtml", new EmailParams(token, provisionerName))
+                : await _razorConverterService.RenderViewToStringAsync("/Views/Emails/OfficeManagerEmail.cshtml", new EmailParams(token));
```
james-hollinger, line 95: How about
`string view = string.IsNullOrEmpty(provisionerName) ? "/Views/Emails/VendorEmail.cshtml" : "/Views/Emails/OfficeManagerEmail.cshtml";
string emailBody = _razorConverterService.RenderViewToStringAsync(view, new EmailParams(token, provisionerName);)`

james-hollinger: Also looks like the logic is reversed from the old code?

PR #576, opened 2020-04-15
PRIME-682 - GPID Improvements

james-hollinger:
GPIDs have characters in them that are not HL7 compliant. This PR makes the GPID character set the following:

ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.,?!@#$%*

and adds a uniqueness check to GPIDs.

Also added a unique index on HPDIDs and GPIDs to the Enrollee table.


prime-dotnet-webapi/Configuration/EnrolleeConfiguration.cs
```diff
@@ -9,10 +9,13 @@ public class EnrolleeConfiguration : IEntityTypeConfiguration<Enrollee>
         public void Configure(EntityTypeBuilder<Enrollee> builder)
         {
             builder
-                .HasKey(a => new { a.Id });
```
james-hollinger, line 12: Dont need the key in the configuration, we have the [Key] attribute on the model

PR #588, opened 2020-04-20
PRIME-705 - Site Registration API and Angular Workflows

mtpultz:


prime-dotnet-webapi/Services/SiteService.cs
```diff
@@ -0,0 +1,102 @@
...
+            if (site == null)
+            {
+                return;
+            }
+
+            _context.Sites.Remove(site);
+            await _context.SaveChangesAsync();
+        }
+
+        public async Task<Site> GetSiteNoTrackingAsync(int siteId)
```
james-hollinger, line 85: Why both getsite and getsitenotracking?

mtpultz: This is how enrollee is implemented, and I thought we did this so we could get an untracked version of the current model.  So if it needs to be changed we should change it in both places.

mtpultz: Honestly, I threw the initial API together very quickly on Monday since we had an 8 point ticket that hadn't been started that included the entire front- and back-end to get the site registration's workflow in place.  I don't think I have the ability to answer questions on implementation at the moment, but feel like any suggestions on implementation changes would be welcome.

prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -0,0 +1,251 @@
...
+using Prime.Models;
+using Prime.Models.Api;
+using Prime.Services;
+
+namespace Prime.Controllers
+{
+    [Produces("application/json")]
+    [Route("api/[controller]")]
+    [ApiController]
+    [Authorize(Policy = AuthConstants.USER_POLICY)]
```
james-hollinger, line 17: Rather than check for the feature flag in each endpoint, this can be updated to be
     [Authorize(Policy = AuthConstants.USER_POLICY, Roles = AuthConstants.FEATURE_SITE_REGISTRATION)]

prime-dotnet-webapi/Extensions/ClaimsPrincipalExtensions.cs
```diff
@@ -32,6 +32,20 @@ public static bool CanEdit(this ClaimsPrincipal User, Enrollee enrollee)
             return !PrimeUserId.Equals(Guid.Empty) && PrimeUserId.Equals(enrollee.UserId);
         }
 
+        /// <summary>
+        /// Returns true if the logged in user is an admin, or if the user has the same UserId as the record
+        /// </summary>
+        public static bool PartyCanEdit(this ClaimsPrincipal User, Party party)
```
james-hollinger, line 38: I feel like CanEdit could be updated to take in an interface object for both Enrollee and Party. Something like
`public interface IUserBoundModel
    {
        Guid UserId { get; set; }
    }`

prime-dotnet-webapi/Extensions/ClaimsPrincipalExtensions.cs
```diff
@@ -50,12 +64,31 @@ public static bool HasAdminView(this ClaimsPrincipal User)
...
             Claim assuranceLevelClaim = User?.Claims?.SingleOrDefault(c => c.Type == AuthConstants.ASSURANCE_LEVEL_CLAIM_TYPE);
 
             int assuranceLevel;
             return Int32.TryParse(assuranceLevelClaim?.Value, out assuranceLevel) && assuranceLevel == level;
         }
+
+        /// <summary>
+        /// Returns true if the logged in user is an admin, or if the user has the same UserId as the record
+        /// </summary>
+        public static bool CanEditSite(this ClaimsPrincipal User, Party party)
```
james-hollinger, line 83: this seems to be a duplicate

prime-dotnet-webapi/Models/SiteRegistration/Location.cs
```diff
@@ -19,25 +19,25 @@ public class Location : BaseAuditable
 
         public string DoingBusinessAs { get; set; }
 
-        public int AddressId { get; set; }
+        public int? PhysicalAddressId { get; set; }
```
james-hollinger, line 22: shouldn't need the Address id, ef can handle it for you.

prime-dotnet-webapi/Services/PartyService.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        public async Task<Party> GetPartyAsync(int partyId)
+        {
+            IQueryable<Party> query = this.GetBasePartyQuery();
+
+            var entity = await query
+                .SingleOrDefaultAsync(e => e.Id == partyId);
+
+            return entity;
```
james-hollinger, line 108: I assume this is broken up for puting in breakpoints but this can just be
`return await GetBasePartyQuery()
            .SingleOrDefaultAsync(p => p.Id == partyId);`

prime-dotnet-webapi/Services/PartyService.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            _context.Parties.Remove(party);
+            await _context.SaveChangesAsync();
+        }
+
+
+        private IQueryable<Party> GetBasePartyQuery()
+        {
+            return _context.Parties;
```
james-hollinger, line 98: this doesnt include the address, probably should?

prime-dotnet-webapi/Services/PartyService.cs
```diff
@@ -0,0 +1,112 @@
...
+            var created = await _context.SaveChangesAsync();
+            if (created < 1)
+            {
+                throw new InvalidOperationException("Could not create party.");
+            }
+
+            return party.Id;
+        }
+
+        public async Task<int> UpdatePartyAsync(int partyId, Party party)
```
james-hollinger, line 64: From a quick reading I dont think this method will update the address at all. Its not in the include when fetching the model and SetValues() does not update child objects.

prime-dotnet-webapi/Services/PartyService.cs
```diff
@@ -0,0 +1,112 @@
...
+namespace Prime.Services
+{
+    public class PartyService : BaseService, IPartyService
+    {
+        private readonly IBusinessEventService _businessEventService;
+
+        public PartyService(
+            ApiDbContext context,
+            IHttpContextAccessor httpContext,
+            IBusinessEventService businessEventService)
```
james-hollinger, line 18: Doesn't look like this is being used

prime-dotnet-webapi/Services/PartyService.cs
james-hollinger: This method isn't being used much in the service

prime-dotnet-webapi/Startup.cs
```diff
@@ -56,12 +56,15 @@ public void ConfigureServices(IServiceCollection services)
...
             services.AddScoped<IRazorConverterService, RazorConverterService>();
+            services.AddScoped<ISiteService, SiteService>();
+            services.AddScoped<IPartyService, PartyService>();
 
             services.AddControllers()
                 .AddNewtonsoftJson(options =>
                 {
                     options.SerializerSettings.Converters.Add(new StringEnumConverter(new KebabCaseNamingStrategy(), false));
                     options.SerializerSettings.Converters.Add(new EmptyStringToNullJsonConverter());
+                    options.SerializerSettings.ReferenceLoopHandling = Newtonsoft.Json.ReferenceLoopHandling.Ignore;
```
james-hollinger, line 67: What problem were you seeing that made this necessary?

AnaisHebert: It was because of the old json import so things weren't json ignoring properly. removed now.

prime-dotnet-webapi/Services/PartyService.cs
```diff
@@ -0,0 +1,118 @@
...
+
+            return party.Id;
+        }
+
+        public async Task<int> UpdatePartyAsync(int partyId, Party party)
+        {
+            var currentParty = await GetBasePartyQuery()
+                .SingleAsync(e => e.Id == partyId);
+
+            _context.Entry(party).CurrentValues.SetValues(party);
```
james-hollinger, line 62: This line looks iffy to me, should it be Entry(currentParty)?

prime-dotnet-webapi/Services/PartyService.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                return await _context.SaveChangesAsync();
+            }
+            catch (DbUpdateConcurrencyException)
+            {
+                return 0;
+            }
+        }
+
+        public void UpdatePartyAddress(Party current, Party updated)
```
james-hollinger, line 76: Dont think this method handles the case where the new party's address is null

prime-dotnet-webapi/Models/IUserBoundModel.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace Prime.Models
+{
+    public interface IUserBoundModel
+    {
+        Guid UserId { get; set; }
+    }
+}
```
james-hollinger, line 9: Admins also implement this interface.

prime-dotnet-webapi/Services/PartyService.cs
AnaisHebert: Yeah it does this adds it if null otherwise it updates it
` if (current.PhysicalAddress == null)
  {
        current.PhysicalAddress = updated.PhysicalAddress;
   }`

prime-angular-frontend/src/app/modules/common/services/form-utils.service.ts
```diff
@@ -62,4 +75,25 @@ export class FormUtilsService {
     }
     return false;
   }
+
+  /**
+   * @description
+   * Get all the errors contained within a form.
+   *
+   */
+  public getFormErrors(form: FormGroup | FormArray): { [key: string]: any; } | null {
```
AnaisHebert, line 84: awesome

prime-dotnet-webapi/Migrations/20200427215637_SiteRegistrationUpdates.cs
```diff
@@ -0,0 +1,565 @@
...
+
+            migrationBuilder.AddColumn<int>(
+                name: "SiteId",
+                table: "BusinessEvent",
+                nullable: true);
+
+            migrationBuilder.InsertData(
+                table: "Vendor",
+                columns: new[] { "Id", "CreatedTimeStamp", "CreatedUserId", "Email", "Name", "UpdatedTimeStamp", "UpdatedUserId" },
+                values: new object[,]
```
itchison, line 155: These are fine. I see lots of value in storing these email addresses (ie other integrations, automation, etc)

PR #589, opened 2020-04-21
Feature/prime 692 decline

hogmann:
Currently just need to hook up the guard, but may be taking the wrong route at which point i should route to the "Declined" Screen so ill ask some advise in the morning.


prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -434,8 +434,19 @@ public async Task<IEnumerable<HpdidLookup>> HpdidLookupAsync(IEnumerable<string>
             return await _context.Enrollees
                 .Include(e => e.AccessTerms)
                 .Where(e => hpdids.Contains(e.HPDID))
+                .Where(e => e.CurrentStatus.StatusCode != (int)StatusType.Declined)
```
james-hollinger, line 437: Status has an IsType method

prime-angular-frontend/src/app/app.routes.ts
```diff
@@ -5,4 +5,5 @@ export class AppRoutes {
   public static PAGE_NOT_FOUND = 'page-not-found';
   public static HELP = 'help';
   public static DEFAULT = '**';
+  public static ACCESS_DECLINED = 'access-declined';
```
mtpultz, line 8: Feels like this is left over from your refactor, and now exists in the EnrolmentRoutes

prime-angular-frontend/src/app/modules/adjudication/shared/components/adjudication-container/adjudication-container.component.ts
```diff
@@ -231,6 +231,65 @@ export class AdjudicationContainerComponent extends AbstractComponent implements
       });
   }
 
+  public onDeclineEnrollee(enrolleeId: number) {
+    const data: DialogOptions = {
+      title: 'Decline Enrollee',
+      message: 'When declined the enrollee will not have access to PRIME and their Terms of Access revoked,'
+        + ' Are you sure you want to lock this enrollee ? ',
```
mtpultz, line 238: You can get multiple lines using backticks and don't have to concatenate.  Also, might be worth updating "and their Terms of Access revoked" to "and their Terms of Access will be revoked"

prime-angular-frontend/src/app/modules/enrolment/pages/access-declined/access-declined.component.html
```diff
@@ -0,0 +1,25 @@
...
+
+  <app-page-header>PharmaNet Enrolment</app-page-header>
+
+  <app-enrolment-progress-indicator *ngIf="isInitialEnrolment"
+                                    message="Enrolment Declined"
+                                    [currentRoute]="EnrolmentRoutes.ACCESS_DECLINED"></app-enrolment-progress-indicator>
+
+  <div class="mb-5">
+    <app-alert type="danger">
+      <ng-container #alertContent
```
mtpultz, line 11: Can you update the alert to have a title like the access-locked page "Your PRIME enrolment has been declined", and it the content of the alert looks like it should be "Your application for PharmaNet access has not be approved."

prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -434,8 +434,19 @@ public async Task<IEnumerable<HpdidLookup>> HpdidLookupAsync(IEnumerable<string>
...
                 .Select(e => HpdidLookup.FromEnrollee(e))
                 .ToListAsync();
         }
+
+        public async Task<Enrollee> GetEnrolleeForUserIdNotDeclinedAsync(Guid userId)
```
james-hollinger, line 442: I think I would rather see an optional parameter on getenrolleeforuserid with excludeDeclined = false or something like that.

PR #595, opened 2020-04-25
PRIME-570 - Resend Reminder Notification

mtpultz:


prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -459,5 +462,31 @@ public async Task<ActionResult<IEnumerable<BusinessEvent>>> getEnrolleeBusinessE
 
             return Ok(ApiResponse.Result(events));
         }
+
+        // GET: api/Enrollees/5/reminder
+        /// <summary>
+        /// Send an enrollee a reminder email.
+        /// </summary>
+        /// <param name="enrolleeId"></param>
+        [HttpGet("{enrolleeId}/reminder", Name = nameof(sendEnrolleeReminderEmail))]
```
james-hollinger, line 471: Should probably be a POST

mtpultz: Sure, I can make it a POST if you feel that verb better suits the creation of an email, but it isn't creating anything in a collection, and there is no body so didn't seem necessary.

PR #600, opened 2020-04-29
PRIME-556 - Remove Provisioner Link Max Views

james-hollinger:
Removes max views for a provisioner link.
Also moved some parameters around. Vendor emails are in the database now, and getting the emails in now in the email service.


james-hollinger:
> 
> 
> Im assuming you this will be brought back, that is why you left all the code in?

From the ticket is seems we are removing this limitation to gather more imperical data for what the limit should be in the future.

PR #609, opened 2020-05-08
PRIME-730 - Current TOA Update

AnaisHebert:


prime-dotnet-webapi/Services/AccessTermService.cs
```diff
@@ -200,6 +203,32 @@ public async Task<bool> IsCurrentByEnrolleeAsync(Enrollee enrollee)
...
+        /// Under Review, Locked, Declined -> Current TOA status will read: NA
+        /// Required TOA -> Current TOA status: Pending
+        /// Editable -> Current TOA status: Is their signed TOA the most current version and not expired
+        /// </summary>
+        public async Task<string> GetCurrentTOAStatusAsync(Enrollee enrollee)
+        {
+            var currentStatus = enrollee.CurrentStatus.StatusCode;
+            var isCurrent = await this.IsCurrentByEnrolleeAsync(enrollee);
+
+            if (currentStatus == (int)StatusType.RequiresToa)
```
james-hollinger, line 217: EnrolmentStatus has an IsType method

prime-dotnet-webapi/Services/AccessTermService.cs
```diff
@@ -200,6 +203,32 @@ public async Task<bool> IsCurrentByEnrolleeAsync(Enrollee enrollee)
...
+            var currentStatus = enrollee.CurrentStatus;
+            var isCurrent = await this.IsCurrentByEnrolleeAsync(enrollee);
+
+            if (currentStatus.IsType(StatusType.RequiresToa))
+            {
+                return "Pending";
+            }
+            else if (currentStatus.IsType(StatusType.Editable))
+            {
+                var accessTerm = await GetMostRecentAcceptedEnrolleesAccessTermAsync(enrollee.Id);
```
james-hollinger, line 223: Could we move the isCurrent call into here to avoid hitting the database if we don't have to?

AnaisHebert: yup!

PR #613, opened 2020-05-11
PRIME-730 - Current TOA update 2

AnaisHebert:
Update Rules for current TOA status as requested in Jira


prime-dotnet-webapi/Services/AccessTermService.cs
```diff
@@ -205,28 +205,38 @@ public async Task<bool> IsCurrentByEnrolleeAsync(Enrollee enrollee)
...
 
-            if (currentStatus.IsType(StatusType.RequiresToa))
+            if (currentStatus.IsType(StatusType.Locked) || currentStatus.IsType(StatusType.Declined))
             {
-                return "Pending";
+                return "NA";
+            }
+            else if (currentStatus.IsType(StatusType.RequiresToa))
+            {
+                toaStatus = "Pending";
```
james-hollinger, line 225: could this just return "Pending"?

prime-dotnet-webapi/Services/AccessTermService.cs
```diff
@@ -205,28 +205,38 @@ public async Task<bool> IsCurrentByEnrolleeAsync(Enrollee enrollee)
...
             }
             else if (currentStatus.IsType(StatusType.Editable))
             {
                 var accessTerm = await GetMostRecentAcceptedEnrolleesAccessTermAsync(enrollee.Id);
                 var isCurrent = await this.IsCurrentByEnrolleeAsync(enrollee);
 
-                return (accessTerm != null && isCurrent && accessTerm.ExpiryDate > DateTimeOffset.Now)
-                ? "Yes"
-                : "No";
+                if (accessTerm?.ExpiryDate > DateTimeOffset.Now)
```
james-hollinger, line 232: I'm not sure what the business wants if you're in editable but you have no accepted access terms. Before it would show "no", now it shows "---"

prime-dotnet-webapi/Services/AccessTermService.cs
```diff
@@ -205,28 +205,38 @@ public async Task<bool> IsCurrentByEnrolleeAsync(Enrollee enrollee)
...
+        /// Locked, Declined -> Current TOA status: NA
         /// Required TOA -> Current TOA status: Pending
-        /// Editable -> Current TOA status: Is their signed TOA the most current version and not expired
+        /// Editable (AND before their renewal date) -> Current TOA status: Is their signed TOA the most current version
+        /// Editable (AND on/after their renewal date) -> Current TOA status: --
         /// </summary>
         public async Task<string> GetCurrentTOAStatusAsync(Enrollee enrollee)
         {
             var currentStatus = enrollee.CurrentStatus;
+            var toaStatus = "";
```
james-hollinger, line 217: This code now returns "" instead of "--" when under review

AnaisHebert: Yeah the front end handles it with the default pipe now instead of setting it to "--" on the backend

PR #615, opened 2020-05-12
PRIME-727 - Review Status Reasons

hogmann:


prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-review-status/components/review-status-content/review-status-content.component.html
```diff
@@ -12,4 +12,30 @@ <h3 class="mb-4">This Enrollee was manually adjudicated for the following reason
       </div>
     </div>
   </div>
+  <div class="col-6 pl-0">
```
mtpultz, line 15: I think this ticket was looking for a way to combine previous statuses with adjudicator notes for displaying the history of statuses with descriptions like in the mockups.

mtpultz:
It looks like when in Requires TOA that it doesn't display the manual review reasons.

hogmann:
Alright all the api logic is now in there for linking admin + note + status changes. It is still in a rough table but with the similar styling to how it would be show above

prime-angular-frontend/src/app/modules/adjudication/shared/components/review-status-content/review-status-content.component.ts
```diff
@@ -1,16 +1,154 @@
...
+    + ' of a health profession in any jurisdiction, that involved improper access to, collection,'
+    + ' use, or disclosure of personal information?';
+  private pharmanetQ = 'Have you ever had your access to an electronic health record system,'
+    + ' electronic medical record system, pharmacy or laboratory record system,'
+    + ' or any similar health information system, in any jurisdiction, suspended or cancelled?';
+  private disciplinaryQ = 'Have you ever been disciplined or fired by an employer, or had a contract for your services terminated,'
+    + ' for a matter that involved improper access to, collection, use, or disclosure of personal information?';
+
+
+  @Input() set enrollee(value: Enrolment) {
```
mtpultz, line 50: Another option instead of using a method with an input decorator is to use the OnChanges hook ngOnChanges then you only need the one input and setter or getter, and ngOnChanges can manage updates/operations for any binding updates.

mtpultz:
@hogmann  Looks great!  I added a comment as an FYI only.

prime-angular-frontend/src/app/modules/adjudication/shared/components/review-status-content/review-status-content.component.ts
hogmann: I did this as the actual first pass but ran into problems, and decided to go this route as it worked. Put it on my note pad as on another pass to update.

PR #618, opened 2020-05-19
PRIME-740 -  CHES Email Service

AnaisHebert:
Currently sends every email to Anais' until ready to merge into dev.
useCHES var is currently set to true in dev. It will be set to false everywhere but prod once testing is complete


prime-dotnet-webapi/Services/CHESApiService.cs
```diff
@@ -0,0 +1,173 @@
...
+        private static async Task<HttpClient> InitHttpClientAsync()
+        {
+            // if (PrimeConstants.ENVIRONMENT_NAME == "local")
+            // {
+            //     return null;
+            // }
+
+            HttpClient client = new HttpClient();
+
+            if (accessToken == "")
```
james-hollinger, line 35: JWT tokens expire, often in about 5 minutes. This should probably fetch a new token every time, or manage token expiry.

PR #620, opened 2020-05-19
PRIME-676 - Business Licence Upload Frontend

AnaisHebert:
Blocked by https://bcgovmoh.atlassian.net/browse/PRIME-695


openshift/document-manager.bc.yaml
```diff
@@ -0,0 +1,154 @@
...
+          - name: DB_NAME
+            valueFrom:
+              secretKeyRef:
+                key: database-name
+                name: "postgresql-secret"
+          - name: CACHE_REDIS_PASS
+            valueFrom:
+              secretKeyRef:
+                key: database-admin-password
+                name: "postgresql-secret"
```
james-hollinger, line 98: Is it intentional that the redis cache has the same password as the postgres database?

hogmann: @poggdogg

poggdogg:
Pulled out my templates - this is good for now.  I will build templates in another PR

openshift/document-manager.bc.yaml
james-hollinger: After some discussion, it seems so.

PR #628, opened 2020-05-21
PRIME-726 - De-brand prime phone number

AnaisHebert:


prime-angular-frontend/src/environments/environment.prod.template.ts
```diff
@@ -4,7 +4,7 @@ export const environment = {
   apiEndpoint: '/api/v1',
   loginRedirectUrl: '$REDIRECT_URL',
   prime: {
-    displayPhone: '1-844-39PRIME',
+    displayPhone: '1-844-397-7463',
```
mtpultz, line 7: We'll want to drop `displayPhone` from the environment variables.  It only exists for providing the human readable vanity portion of an anchor in the PRIME phone component.  So it will mean dropping it out of the component and removing the mode input binding and vanity related methods and logic and just straight up using the phone number.

mtpultz: Or leave the vanity number unchanged in the environment files in case they want it back, but make the default mode `normal` in the component.

PR #630, opened 2020-05-25
PRIME-671 - Landing Page (Version 2)

mtpultz:
* Added newer updates to the new updates


mtpultz:
> Can you remove the second prime phone number in brackets now that we have removed the display number

PR #631, opened 2020-05-25
PRIME-666 - Visual Flag when Enrollee is set to always manual

AnaisHebert:


prime-angular-frontend/src/app/modules/adjudication/shared/components/adjudication-container/adjudication-container.component.ts
```diff
@@ -359,6 +355,34 @@ export class AdjudicationContainerComponent extends AbstractComponent implements
     }
   }
 
+  public onToggleFlag(enrollee: HttpEnrollee) {
```
mtpultz, line 358: I would probably make this `onToggleManualAdj`, and similarly for the event emitter `toggleManualAdj` so a developer in the future doesn't have to trace through the code to find out what the flag is referring since flag is only a visual/written cue for users in the UI and not subject of the event on an enrollee.

PR #633, opened 2020-05-27
PRIME-761 - Add organization type to site

AnaisHebert:


prime-angular-frontend/src/app/modules/site-registration/shared/models/organization.model.ts
```diff
@@ -4,6 +4,7 @@ export interface Organization {
   id?: number;
   signingAuthorityId?: number;
   signingAuthority: Party;
+  organizationTypeCode?: number;
```
mtpultz, line 7: Is the field optional?

AnaisHebert: No its not. we just have the rest of our id's as optional so i figured it needed to be that way to allow us to fill out from page to page?

PR #640, opened 2020-06-02
PRIME-259 - BCeID

james-hollinger:
BCeID has now been enabled in Dev only. BCeID requires two-factor authentication using GoogleAuth or FreeOTP (only FreeOTP ahs been tested). Currently, BCeIDs are automatically granted the prime_user role and have access to the application normally. A new auto-adjudication rule will flag any identity providers that are not BC Services Card to manual review (in addition to flagging any IDPs that have an assurance level of 2 or lower).

At this point, the assurance level requirement has been removed from our API's Auth policies, due to BCeID having an assurance level of 1.


prime-dotnet-webapi/Configuration/StatusReasonConfiguration.cs
```diff
@@ -23,6 +23,7 @@ public override ICollection<StatusReason> SeedData
                     new StatusReason { Code = 11, Name = "Contact Address or Identity Address not in British Columbia", CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
                     new StatusReason { Code = 12, Name = "Admin has flagged the applicant for manual adjudication", CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
                     new StatusReason { Code = 13, Name = "User does not have high enough identity assurance level", CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
+                    new StatusReason { Code = 14, Name = "User authenticated with a method other than BC Services Card", CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
```
hogmann, line 26: Dang will update my PR as i as well added a rule number #14 !

PR #671, opened 2020-06-16
PRIME-782 - Download and upload signed Organization Agreement

AnaisHebert:


prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -254,5 +259,96 @@ public async Task<ActionResult<Organization>> SubmitOrganizationRegistration(int
...
+            var licences = await _organizationService.GetSignedAgreementsAsync(organization.Id);
+
+            return Ok(ApiResponse.Result(licences));
+        }
+
+        // GET: api/Organizations/download-organization-agreement
+        /// <summary>
+        /// Download the organization agreement document.
+        /// </summary>
+        [HttpGet("download-organization-agreement", Name = nameof(DownloadOrganizationAgreement))]
```
james-hollinger, line 333: REST endpoints don't generally have verbs in them. How about GET Organizations/organization-agreement-document ?

james-hollinger: I could also see this being on the document controller, so it would be GET document/organization-agreement

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -254,5 +261,102 @@ public async Task<ActionResult<Organization>> SubmitOrganizationRegistration(int
...
+        // Get: api/organizations/5/signed-agreement
+        /// <summary>
+        /// Gets the signed agreement for a organization.
+        /// </summary>
+        /// <param name="organizationId"></param>
+        [HttpGet("{organizationId}/signed-agreement", Name = nameof(GetSignedAgreement))]
+        [ProducesResponseType(typeof(ApiBadRequestResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiResultResponse<Organization>), StatusCodes.Status201Created)]
```
james-hollinger, line 311: This response code seems wrong

prime-dotnet-webapi/Controllers/OrganizationsController.cs
mtpultz: That's a great idea James, and I think in the next sprint we should definitely create a document controller.

PR #679, opened 2020-06-22
PRIME-720 admin download of self declaration

hogmann:


hogmann:
We can wait to review this once 38 is in develop

PR #691, opened 2020-06-30
Fixing PVC

poggdogg:
The front-end is getting a server error 500 as a result of getting a 401 on the API (not authorized) when you try to post a document to:
https://pr-691.pharmanetenrolment.gov.bc.ca/site-registration/organizations/1/sites/1/business-licence

We need to find out why the API can't post to document-manager


ajdeziel:
Is this good to go for review?

poggdogg:
Fixed local build

PR #693, opened 2020-07-01
PRIME-771 Manually Triggering a reminder email in Admin isn't adding the event onto Events Log

mtpultz:


prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -554,7 +554,10 @@ public async Task<ActionResult> sendEnrolleeReminderEmail(int enrolleeId)
                 return NotFound(ApiResponse.Message($"Enrollee not found with id {enrolleeId}"));
             }
 
+            var admin = await _adminService.GetAdminForUserIdAsync(User.GetPrimeUserId());
+            var username = admin.IDIR.Replace("@idir", "");
             await _emailService.SendReminderEmailAsync(enrollee);
+            await _businessEventService.CreateEmailEventAsync(enrollee.Id, $"Email reminder sent to Enrollee by {username}");
```
james-hollinger, line 560: This wont _currently_ work with readonly admins since they dont create Admin objects in our DB. This will work once we merge in our Auth changes. (also, should readonly admins be able to hit this endpoint?)

PR #696, opened 2020-07-01
PRIME-795 Issuing PRIME Verifiable Credential Upon TOA Acceptance 

mtpultz:


prime-dotnet-webapi/Migrations/20200625182734_Credential.cs
```diff
@@ -0,0 +1,67 @@
...
+                name: "Credential",
+                columns: table => new
+                {
+                    Id = table.Column<int>(nullable: false)
+                        .Annotation("Npgsql:ValueGenerationStrategy", NpgsqlValueGenerationStrategy.IdentityByDefaultColumn),
+                    CreatedUserId = table.Column<Guid>(nullable: false),
+                    CreatedTimeStamp = table.Column<DateTimeOffset>(nullable: false),
+                    UpdatedUserId = table.Column<Guid>(nullable: false),
+                    UpdatedTimeStamp = table.Column<DateTimeOffset>(nullable: false),
+                    SchemaId = table.Column<string>(nullable: true),
```
itchison, line 26: are these really nullable true?

AnaisHebert: image
I think its just not set as specifically required, but its not set to nullable true specifically either

PR #697, opened 2020-07-02
Add back KEYCLOAK_TOKEN_URL env var

ajdeziel:
KEYCLOAK_TOKEN_URL was removed from Document Manager's OpenShift deploy config templates, and are now being added back to assess whether this was the cause of the 401 errors.


poggdogg:
Fixed docman local and cloud

PR #698, opened 2020-07-02
PRIME-740 Use CHES for Email

AnaisHebert:
Emails with send through smtp on develop and test


prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -172,8 +182,8 @@ public async Task SendSiteRegistrationAsync(Site site)
                 ("SiteRegistrationReview.pdf", await _razorConverterService.RenderViewToStringAsync("/Views/SiteRegistrationReview.cshtml", site)),
                 ("BusinessLicence.pdf", await _razorConverterService.RenderViewToStringAsync(businessLicenceTemplate, businessLicenceDoc))
             }
-            .Select(content => (Filename: content.Filename, Content: _pdfService.Generate(content.HtmlContent)))
-            .Select(pdf => new Attachment(new MemoryStream(pdf.Content), pdf.Filename, "application/pdf"));
+            .Select(content => (Filename: content.Filename, Content: _pdfService.Generate(content.HtmlContent)));
+            // .Select(pdf => new Attachment(new MemoryStream(pdf.Content), pdf.Filename, "application/pdf"));
```
hogmann, line 186: What is the reason for the commented out line?

hogmann: Is it something that we need later or something that isn't really needed? As it looks like all the functions changed for what they accept as attachments so could potentially just remove this line.

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -195,26 +205,32 @@ public async Task<IEnumerable<string>> GetPharmaNetProvisionerNamesAsync()
...
             if (!to.Any())
             {
                 throw new ArgumentException("Must specify at least one \"To\" email address.");
             }
 
+            // TODO temporarily only send emails to Anais just in case
+            var tempTo = new List<string>();
+            tempTo.Add("anais.hebert@nttdata.com");
+
+            to = tempTo;
```
hogmann, line 232: Testing code that needs to be removed.

AnaisHebert: Just here temporarily to make sure no one accidently sends a real email to someone until testing is done

openshift/dotnet-webapi.dc.yaml
```diff
@@ -102,6 +102,21 @@ objects:
                     secretKeyRef:
                       key: DOCUMENT_MANAGER_CLIENT_SECRET
                       name: "document-manager-secret"
+                - name: OPENID_API_URL
```
james-hollinger, line 105: How about CHES_TOKEN_URL? in a vacuum this env variable is not very descriptive

openshift/dotnet-webapi.dc.yaml
```diff
@@ -102,6 +102,21 @@ objects:
                     secretKeyRef:
                       key: DOCUMENT_MANAGER_CLIENT_SECRET
                       name: "document-manager-secret"
+                - name: OPENID_API_URL
+                  valueFrom:
+                    secretKeyRef:
+                      key: OPENID_API_URL
+                      name: "ches-secrets"
+                - name: PRIME_SERVICE_CLIENT
```
james-hollinger, line 110: same here, could this be CHES_CLIENT_SECRET or something similar?

openshift/dotnet-webapi.dc.yaml
```diff
@@ -184,6 +199,11 @@ objects:
                     secretKeyRef:
                       key: KEYCLOAK_TOKEN_URL
                       name: "keycloak" 
+                - name: useCHES
```
james-hollinger, line 202: CHES_ENABLED seems like a better name to me

prime-dotnet-webapi/PrimeConstants.cs
```diff
@@ -23,6 +23,13 @@ public static class PrimeConstants
         public readonly static string PHARMANET_SSL_CERT_FILENAME = Environment.GetEnvironmentVariable("PHARMANET_SSL_CERT_FILENAME");
         public readonly static string PHARMANET_SSL_CERT_PASSWORD = Environment.GetEnvironmentVariable("PHARMANET_SSL_CERT_PASSWORD");
 
+        // CHES Email API
+        public readonly static string USE_CHES = Environment.GetEnvironmentVariable("useCHES") ?? "true";
```
james-hollinger, line 27: These constants dont have to be a string, they can be a bool. mail server port is an int, for example.
The default should probably be false, as well.
how about:

public readonly static bool CHES_ENABLED = bool.Parse(Environment.GetEnvironmentVariable("CHES_ENABLED") ?? "false");

prime-dotnet-webapi/Services/Clients/Access Tokens/ClientCredentials.cs
```diff
@@ -4,5 +4,5 @@ namespace Prime.Services.Clients
 {
     public class DocumentManagerClientCredentials : ClientCredentialsTokenRequest { }
 
-    public class ChesClientCredentials : ClientCredentialsTokenRequest { }
+    public class CHESClientCredentials : ClientCredentialsTokenRequest { }
```
james-hollinger, line 7: The C# naming conventions for acronyms longer than two characters is to have a leading capital only. See HttpClient.

prime-dotnet-webapi/Services/Clients/Email/CHESClient.cs
```diff
@@ -0,0 +1,148 @@
...
+using System.Net.Mail;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Newtonsoft.Json;
+
+namespace Prime.Services
+{
+    public class CHESClient : BaseService, ICHESClient
```
james-hollinger, line 16: Since this is meant only to do web requests and doesn't need to talk to the database, you dont need BaseService or the DI in the constructor for it.

prime-dotnet-webapi/PrimeConstants.cs
AnaisHebert: yes the default will be false as soon as testing is done

prime-dotnet-webapi/Startup.cs
```diff
@@ -142,6 +143,21 @@ protected void ConfigureClients(IServiceCollection services)
...
             });
+
+            services.AddTransient<CHESBearerTokenHandler>()
+            .AddHttpClient<ICHESClient, CHESClient>(client =>
+            {
+                client.BaseAddress = new Uri(PrimeConstants.CHES_API_URL.EnsureTrailingSlash());
+            })
+            .AddHttpMessageHandler<CHESBearerTokenHandler>();
+
+            services.AddHttpClient<IAccessTokenClient, AccessTokenClient>();
```
james-hollinger, line 154: This is already added on line 138, you dont need to add it again

prime-dotnet-webapi/Services/Clients/Email/CHESClient.cs
```diff
@@ -0,0 +1,148 @@
...
+            }
+
+            var requestParams = new CHESEmailRequestParams(from, to, subject, body, chesAttachments);
+
+            var requestContent = new StringContent(JsonConvert.SerializeObject(requestParams), Encoding.UTF8, "application/json");
+
+            HttpResponseMessage response = null;
+            try
+            {
+                response = await _client.PostAsync(PrimeConstants.CHES_API_URL + "/email", requestContent);
```
james-hollinger, line 52: the base url of the client is set up in Startup.cs, so you can use a relative URL here rather than specify the entire URL again. Try:

_client.PostAsync("email", requestContent); 

(note: no leading slash, "/email" wont work)

prime-dotnet-webapi/PrimeConstants.cs
```diff
@@ -23,6 +23,13 @@ public static class PrimeConstants
...
+        public readonly static string CHES_API_URL = Environment.GetEnvironmentVariable("CHES_API_URL") ?? "https://ches-master-9f0fbe-dev.pathfinder.gov.bc.ca/api/v1";
+        public readonly static string CHES_SERVICE_CLIENT_ID = "PRIME_SERVICE_CLIENT";
+        public readonly static string CHES_SERVICE_CLIENT_SECRET = Environment.GetEnvironmentVariable("PRIME_SERVICE_CLIENT") ?? "88e123a6-80cb-46a0-96d3-e2edae076ae7";
```
james-hollinger, line 30: Environment.GetEnvironmentVariable("PRIME_SERVICE_CLIENT") => Environment.GetEnvironmentVariable("CHES_CLIENT_SECRET")

prime-dotnet-webapi/Services/Clients/Email/CHESClient.cs
```diff
@@ -0,0 +1,146 @@
...
+                throw new Exception("Error occurred when calling CHES Email API. Try again later.", ex);
+            }
+        }
+
+        public async Task<bool> HealthCheckAsync()
+        {
+            HttpResponseMessage response = null;
+            try
+            {
+                response = await _client.GetAsync(new Uri("health"));
```
james-hollinger, line 71: Relative URLs are specified using a string, so this should just be 
`response = await _client.GetAsync("health");`

prime-dotnet-webapi/Services/Clients/Email/SmtpEmailClient.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Net.Http;
+
+namespace Prime.Services
+{
+    public class SmtpEmailClient : ISmtpEmailClient
+    {
+        private static HttpClient _client;
+        public SmtpEmailClient(HttpClient httpClient)
+        {
+            _client = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
```
james-hollinger, line 16: This doesn't actually use the http client, so you don't need to DI it in here.

AnaisHebert: So would I just add this as a a service instead of a client?

PR #700, opened 2020-07-03
Feature/prime 720 self decl download

hogmann:
@poggdogg @ajdeziel Do to the shift with the front end needing access to the document manager we will need an environment variable for the url to get injected to the front end code for the manager


ajdeziel:
@hogmann Do you have the URL somewhere? I can update the branch now if you have it.

PR #707, opened 2020-07-09
PRIME-879 - Enable Comunity Pharmacy in Dev/Test

james-hollinger:
The Community Pharmacy feature flag is now enabled to all users in DEV and TEST.

Kecyloak documentation has been updated,


documentation/Keycloak settings/realm-export-dev.json
```diff
@@ -128,11 +128,18 @@
         "clientRole": false,
```
james-hollinger, line 128: Client Roles:

PR #710, opened 2020-07-09
PRIME-895 API Unit Test

mtpultz:


ajdeziel:
Based on the convo @mtpultz and @ajdeziel had in the Tech Team convo in Zoom, we will flip on SonarQube quality gating to determine its impact on our ability to complete work during the sprint as it is currently not in use. We currently have 0 conditions that are imposed, therefore if SonarQube picks out any code quality issues, the build and deployment of the branch is allowed to continue.

If quality gating begins to impede us, we will turn it off and open another ticket to rectify major issues being raised by SonarQube.

PR #713, opened 2020-07-14
PRIME-833 - Save Files Sent to HIBC

hogmann:
Files sent to HIBC for site registration are now being sent to the document manager as well.

Document manger has been updated with an endpoint allowing the API to send a file directly (rather than using the TUS protocol)


prime-angular-frontend/src/environments/environment.ts
```diff
@@ -7,7 +7,7 @@ export const environment = {
   version: '1.0.0',
   apiEndpoint: 'http://localhost:5000/api',
   loginRedirectUrl: 'http://localhost:4200',
-  documentManagerUrl: 'http://localhost:6001',
+  documentManagerUrl: 'http://localhost:6001/',
```
mtpultz, line 10: I feel like we want to be consistent on how format our URLs and not include a `/` at the end.  If I imagine myself as a developer using this environment variable I would add it when used since when I read this `location.replace(`${this.config.documentManagerUrl}documents/downloads/${token}`);` and don't see the `/` it looks incorrect

hogmann: I agree with you. @ajdeziel @poggdogg the way the document manager is injected it contains the trailing slash. Is this something we can easily change?

james-hollinger: As an FYI, if you want to use a relative URL in an Http Client in C#, the Base Url must have a trailing slash. Hence the extension method EnsureTrailingSlash().

james-hollinger: I wish Javascript had a built in path.join

PR #730, opened 2020-07-20
Spiking GitLab runner

poggdogg:


poggdogg:
Technically possible, business improbable.  GitLab has to be used to manage the CICD runners

PR #741, opened 2020-07-23
Replace REDIRECT_URL with VANITY_URL

ajdeziel:
Fix-a-roo for Document Manager! Replaced `SUFFIX` and `REDIRECT_URL` with `VANITY_URL`.


james-hollinger:
Need the ephemeral DC file as well

openshift/document-manager.dc.yaml
```diff
@@ -265,7 +265,7 @@ objects:
               key: DOCUMENT_MANAGER_CLIENT_SECRET
               name: "document-manager-secret"
         - name: DOCUMENT_MANAGER_URL
-          value: "https://document-manager${SUFFIX}.pharmanetenrolment.gov.bc.ca/"
+          value: "https://${VANITY_URL}/api/docman/"
```
james-hollinger, line 268: The document manager does not have a VANITY_URL parameter

james-hollinger: (hence why the URL we are seeing in file uploads is https://${VANITY_URL}/api ...)

poggdogg: Correct, it needed to be added to the pipeline.  This is done.

ajdeziel:
I've tested PR-741 through the front-end, and was able to submit a site successfully without error. This PR is now ready for review. 👍

ajdeziel:
This work has been merged in as part of #742. This PR will now be closed as a result.

PR #757, opened 2020-08-05
PRIME-933 middle names.

hogmann:
As a note this will not retroactively add given names. Doesn't appear to be in the scope of the ticket, but might need to be thought about.


prime-dotnet-webapi-tests/ModelFactories/EnrolleeFactory.cs
```diff
@@ -21,6 +21,7 @@ public EnrolleeFactory()
             RuleFor(x => x.FirstName, f => f.Name.FirstName());
             RuleFor(x => x.MiddleName, f => f.Name.FirstName());
             RuleFor(x => x.LastName, f => f.Name.LastName());
+            RuleFor(x => x.GivenNames, f => f.Name.FirstName());
```
james-hollinger, line 24: Thank you for updating the tests

prime-dotnet-webapi/Models/Enrollee.cs
```diff
@@ -31,6 +31,9 @@ public class Enrollee : BaseAuditable, IValidatableObject, IUserBoundModel
         [Required]
         public string LastName { get; set; }
 
+        [Required]
+        public string GivenNames { get; set; }
```
james-hollinger, line 35: I assumed we were replacing MiddleName with GivenNames

hogmann: Was going to ask about that, Not sure if we ever enter any data into the middle name field.

mtpultz: Yah, we're replace middle name with given names.  The given names will only appear in the admin UI, and status quo during enrolment like you have it.

hogmann: 👍

prime-angular-frontend/src/app/shared/pipes/certificate.pipe.ts
```diff
@@ -21,9 +21,8 @@ export class CertificatePipe implements PipeTransform {
 
   private getFullName(certificate: any) {
```
mtpultz, line 22: Note for another future ticket so that one of us remembers, but we have I believe 3 pipes that do full name

PR #758, opened 2020-08-05
PRIME-911 Updating Site Information (show pencil icons for where we want to allow updates)

AnaisHebert:


prime-angular-frontend/src/app/modules/site-registration/shared/guards/site.guard.ts
```diff
@@ -40,9 +38,71 @@ export class SiteGuard extends BaseGuard {
...
+    // Otherwise, prevent the route from resolving
+    return false;
+  }
+
+  private manageSubmittedSiteRouting(routePath: string, site: Site) {
+    return this.manageRouting(routePath, SiteRoutes.SITE_MANAGEMENT, site);
+  }
+
+  private manageRouting(routePath: string, defaultRoute: string, site: Site): boolean {
+    let childRoute = routePath.includes('remote-users')
```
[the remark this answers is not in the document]
mtpultz: Also, I might have commented why we're checking for inclusion of the `remote-users` even though it is in `editRegistrationRouteOrder` since another developer might not go an scrutinize the routing module and make the connection.  Alternatively, you could skip this check entirely and add `remote-user` to `editRegistrationRouteOrder ` and rename it to `editRegistrationRouteAccess` since the doesn't control order after the registration has been submitted.

prime-angular-frontend/src/app/modules/site-registration/site-registration.routes.ts
```diff
@@ -69,6 +78,18 @@ export class SiteRoutes {
     ];
   }
 
+  // Used to indicate the routes and order for sites for editing
+  public static editSiteRegistrationRouteOrder(): string[] {
```
mtpultz, line 82: Could be `editRegistrationRouteAccess` which could include `remote-user`.  The method comment would need a small update indicating access not order, and a short comment above `remote-user` on why it was included to the list as a child route of `remote-users`.

prime-angular-frontend/src/app/modules/site-registration/shared/guards/site.guard.ts
```diff
@@ -40,9 +38,67 @@ export class SiteGuard extends BaseGuard {
...
+    return false;
+  }
+
+  private manageSubmittedSiteRouting(routePath: string, site: Site) {
+    return this.manageRouting(routePath, SiteRoutes.SITE_MANAGEMENT, site);
+  }
+
+  private manageRouting(routePath: string, defaultRoute: string, site: Site): boolean {
+
+    const childRoute = routePath.includes('?')
```
mtpultz, line 68: You actually don't need to check for the `?` if you use `.split('?')` and it doesn't exist you'll have an array with a single index, which when you chain it with `.shift()` results in you have the same thing you started with so can one-liner it with no ternary

prime-dotnet-webapi/ViewModels/SiteRegistration/SiteUpdateModel.cs
```diff
@@ -0,0 +1,40 @@
...
+        public Party PrivacyOfficer { get; set; }
+
+        public int? TechnicalSupportId { get; set; }
+
+        public Party TechnicalSupport { get; set; }
+
+        public int? OrganizationTypeCode { get; set; }
+
+        [JsonIgnore]
+        public OrganizationType OrganizationType { get; set; }
```
james-hollinger, line 31: If this is JsonIgnored, the FE can't send this up, right? Do we need this on the update model?

PR #766, opened 2020-08-07
PRIME-880 - TOA Updates

james-hollinger:
There is now a TOA for Comunity Pharmacists. Almost everything to do with TOAs has been refactored after the decision that we don't need dynamic TOA. Many performance improvements.

Several new Extension methods, including .ThrowIfNull for simpler null handling, and .If() on IQueryable for conditional filtering.

Migration and tests to come.


prime-dotnet-webapi/ApiDbContext.cs
```diff
@@ -162,35 +165,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
-
-            modelBuilder.ApplyConfiguration(new GlobalClauseConfiguration());
-            modelBuilder.ApplyConfiguration(new UserClauseConfiguration());
-            modelBuilder.ApplyConfiguration(new LicenseClassClauseConfiguration());
-            modelBuilder.ApplyConfiguration(new LicenseClassClauseMappingConfiguration());
-
-            modelBuilder.ApplyConfiguration(new BusinessEventTypeConfiguration());
-
-            modelBuilder.ApplyConfiguration(new VendorConfiguration());
+            modelBuilder.ApplyConfigurationsFromAssembly(typeof(ApiDbContext).Assembly);
```
james-hollinger, line 168: Rather than loading every file explicitly, we can use ApplyConfigurationsFromAssembly() to load all configuration files inheriting from IEntityTypeConfiguration.

prime-dotnet-webapi/Configuration/AccessTerms/AgreementConfiguration.cs
```diff
@@ -0,0 +1,39 @@
+using System;
+using System.Linq;
+using System.Collections.Generic;
+
+using Prime.Models;
+using Prime.Configuration.Resources;
+
+namespace Prime.Configuration.Agreements
+{
+    public static class AgreementConfiguration
```
james-hollinger, line 10: Despite being named like a configuration file, its not actually an IEntityTypeConfiguration. It just holds the seed data for all the agreements in one place.

prime-dotnet-webapi/Configuration/AccessTerms/AgreementConfiguration.cs
```diff
@@ -0,0 +1,39 @@
...
+                    new OboAgreement                 { Id = 5, Text = "obo-access-terms-v3.html", EffectiveDate = DateTimeOffset.Parse("2020-03-10 00:00:00"), CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
+                    new RegulatedUserAgreement       { Id = 6, Text =  "ru-access-terms-v3.html", EffectiveDate = DateTimeOffset.Parse("2020-05-07 00:00:00"), CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
+                    new OboAgreement                 { Id = 7, Text = "obo-access-terms-v4.html", EffectiveDate = DateTimeOffset.Parse("2020-05-07 00:00:00"), CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
+                    new RegulatedUserAgreement       { Id = 8, Text =  "ru-access-terms-v4.html", EffectiveDate = DateTimeOffset.Parse("2020-06-03 00:00:00"), CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
+                    new CommunityPharmacistAgreement { Id = 9, Text =  "com-pharm-terms-v1.html", EffectiveDate = SEEDING_DATE, CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
```
james-hollinger, line 27: Seed data has the filename as placeholder text now. This should let us use the agreements in tests without worrying about huge amounts of string data. The Configure() methods for each type of agreement call LoadText() when configuring the database to load the text from the file when needed.

PR #772, opened 2020-08-10
PRIME-949 Preventing the 504 Gateway time out for larger documents that are uploaded

mtpultz:


prime-dotnet-webapi/Models/DocumentAccessToken.cs
```diff
@@ -0,0 +1,22 @@
...
+        [Key]
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        public Guid Id { get; set; }
+
+        public Guid DocumentGuid { get; set; }
+
+        [NotMapped]
+        public string DownloadUrl
+        {
+            get => $"{PrimeConstants.BACKEND_URL}/file-download/{Id}";
```
mtpultz, line 19: Is there an alternative to using `PrimeConstants` for getting the API URL?

prime-dotnet-webapi/PrimeConstants.cs
```diff
@@ -6,6 +6,7 @@ public static class PrimeConstants
     {
         public readonly static string ENVIRONMENT_NAME = Environment.GetEnvironmentVariable("OC_APP") ?? "local";
         public readonly static string FRONTEND_URL = Environment.GetEnvironmentVariable("FRONTEND_URL") ?? "localhost:4200";
+        public readonly static string BACKEND_URL = Environment.GetEnvironmentVariable("BACKEND_URL") ?? "localhost:5000";
```
mtpultz, line 9: image

prime-dotnet-webapi/Views/Emails/UpdateRemoteUsersEmail.cshtml
```diff
@@ -51,3 +57,11 @@
     </div>
   }
 }
+
+<h2 class="mb-2">Site Information</h2>
+
+<p>
```
mtpultz, line 63: Will create a Razor component so we're not duplicating text

prime-dotnet-webapi/Controllers/DocumentAccessController.cs
```diff
@@ -0,0 +1,85 @@
...
+            var response = await _documentManagerClient.GetFileAsync(documentAccessToken.DocumentGuid);
+
+            if (response == null)
+            {
+                return NotFound();
+            }
+
+            Response.Headers.Add("Content-Disposition", response.Content.Headers.ContentDisposition.ToString());
+
+            return File(await response.Content.ReadAsStreamAsync(), "application/octet-stream");
```
mtpultz, line 59: Doesn't download immediately when the browser opens up.  Looking at force download, but doesn't prevent review since it works.

prime-dotnet-webapi/Models/DocumentAccessToken.cs
hogmann: From what i gathered from when i had to use it, it is just the way it is.

AnaisHebert: hmm not that i know of

hogmann:
Just testing out the PR branch and the link has localhost in it.
image

mtpultz:
> Just testing out the PR branch and the link has localhost in it.
> image

Yah, waiting on an environment variable

PR #781, opened 2020-08-18
PRIME-973 Remote Practitioners changes

AnaisHebert:


prime-angular-frontend/src/app/modules/site-registration/shared/guards/site.guard.ts
```diff
@@ -65,7 +65,13 @@ export class SiteGuard extends BaseGuard {
 
   private manageRouting(routePath: string, defaultRoute: string, site: Site): boolean {
 
-    const childRoute = routePath.split('?').shift().split('/').pop();
+    let childRoute = routePath.includes('remote-users')
```
mtpultz, line 68: We'll want to use `SiteRoutes.REMOTE_USERS` instead of hardcoding the path

prime-angular-frontend/src/app/modules/site-registration/pages/site-management/site-management.component.html
```diff
@@ -73,36 +56,20 @@
...
+                      [menu]="siteMenu"
+                      [menuOutletContext]="{ site: site }"
+                      [properties]="getSiteProperties(site)"
+                      actionButtonTitle="Update Remote Practitioners"
+                      [actionDisabled]="!site.completed"
+                      (action)="viewSiteRemoteUsers(site)"></app-summary-card>
 
   </ng-container>
 
   <ng-template #property
```
mtpultz, line 70: Looks like this `ng-template` isn't used anymore in `SiteManagement`

prime-angular-frontend/src/app/modules/site-registration/shared/components/summary-card/summary-card.component.ts
```diff
@@ -1,12 +1,36 @@
...
   selector: 'app-summary-card',
   templateUrl: './summary-card.component.html',
   styleUrls: ['./summary-card.component.scss']
 })
 export class SummaryCardComponent implements OnInit {
-  constructor() { }
+  @Input() public icon: string;
+  @Input() public title: string;
+  @Input() public menu: Template;
+  @Input() public menuOutletContext: object;
```
mtpultz, line 13: We might want to make this typing `object | null` it's quite possible that you could pass in a `menu` template reference, but not supply a `menuOutletContext` since the template is scoped to the parent.  We could then set a default in the constructor of `null`.

mtpultz: The only other issue I find with `object` is that it means any non-primitive types so by saying `object` instead of `{ [key:string]: any }` which is equivalent you're sort of saying I'm okay with an object literal or an array.  While `{ [key:string]: any }` allows the same, but you're sort of saying I'm more likely expecting an object literal.

prime-angular-frontend/src/app/modules/site-registration/shared/components/summary-card/summary-card.component.ts
```diff
@@ -1,12 +1,35 @@
...
+
+  @Output() public action: EventEmitter<void>;
+  @Output() public remove: EventEmitter<void>;
+
+  constructor() {
+    this.action = new EventEmitter<void>();
+    this.remove = new EventEmitter<void>();
+  }
+
+  public onClick() {
```
mtpultz, line 25: `onClick` doesn't really provide context other than you clicked something in the DOM, but the element it references uses `actionButtonTitle`, `actionDisabled`, and the event handler itself invokes the `action` event emitter.  I'd probably call this event handler `onAction`

PR #785, opened 2020-08-19
PRIME-981 Site Reg - Organization Information Page replace OrgBook reference with BC Corporate Registry

mtpultz:


mtpultz:
> Just looking at the PRIME-981 ticket. OrgBook needs to be replaced in two places while I only see one change here.

Yah, the other must have been removed in the word smithing updates from last sprint.

PR #795, opened 2020-08-26
Remove Privileges

james-hollinger:
Privileges on Enrollees is not used or displayed currently and requires extra database hits to populate.

This PR is to remove the Privileges property in a single PR so it can be reverted later if need be.


james-hollinger:
> 
> 
> 😥

😿

PR #796, opened 2020-08-26
PRIME-118 Address Validation

AnaisHebert:


prime-dotnet-webapi/Services/Clients/AddressAutocomplete/AddressAutocompleteClient.cs
```diff
@@ -0,0 +1,186 @@
...
+            // Auth header and api-key are injected in Startup.cs
+            _client = client;
+            _logger = logger;
+        }
+
+        public async Task<IEnumerable<AddressAutocompleteFindResponse>> Find(string searchTerm, string lastId)
+        {
+            //Build the url
+            var url = "Find/v2.10/json3ex.ws?";
+            url += "Key=" + System.Web.HttpUtility.UrlEncode(PrimeConstants.ADDRESS_VALIDATION_API_KEY);
```
AnaisHebert: @james-hollinger How do i set a get variable key in the startup httpclient setup?

mtpultz:
Is there a reason we chose to invoke their API through our server versus directly from the web client?

AnaisHebert:
> Is there a reason we chose to invoke their API through our server versus directly from the web client?

No, should i not have done it that way?

mtpultz:
> > Is there a reason we chose to invoke their API through our server versus directly from the web client?
> 
> No, should i not have done it that way?

Unfortunately by passing it through our API we introduce latency to the autocomplete

prime-dotnet-webapi/Services/Clients/AddressAutocomplete/AddressAutocompleteClient.cs
```diff
@@ -0,0 +1,186 @@
...
+            }
+
+            _logger.LogInformation("GET autocomplete find {@JObject}", JsonConvert.SerializeObject(body));
+
+            return autocompleteResponse;
+        }
+
+        public async Task<IEnumerable<AddressAutocompleteRetrieveResponse>> Retrieve(string Id)
+        {
+            //Build the url
```
mtpultz, line 67: Probably don't need this comment

prime-dotnet-webapi/Services/Clients/AddressAutocomplete/AddressAutocompleteClient.cs
```diff
@@ -0,0 +1,186 @@
...
+            var url = "Retrieve/v2.11/json3ex.ws?";
+            url += "Key=" + System.Web.HttpUtility.UrlEncode(PrimeConstants.ADDRESS_AUTOCOMPLETE_API_KEY);
```
mtpultz, line 67: Could use interpolation instead of concatentation

prime-dotnet-webapi/Services/Clients/AddressAutocomplete/AddressAutocompleteClient.cs
james-hollinger: You could dependancy inject a singleton object holding the key, something like

services.AddSingleton(new AddressValidationCredentials { Key = PrimeConstants.ADDRESS_VALIDATION_API_KEY });

in Startup.cs, and then you could change the constructor to
public AddressAutocompleteClient( HttpClient client, ILogger<AddressAutocompleteClient> logger, AddressValidationCredentials credentials)

PR #799, opened 2020-08-28
Fix Migration Designer File

james-hollinger:
Quick fix to the migration files so weird stuff stops appearing.
Also fix backend tests


AnaisHebert:
> 🥳

🦝🦝

PR #800, opened 2020-08-28
Feature/prime 987 hours of operation open at the close

hogmann:


prime-angular-frontend/src/app/modules/site-registration/shared/services/site-form-state.service.ts
```diff
@@ -261,12 +270,49 @@ export class SiteFormStateService extends AbstractFormState<Site> {
 
   private buildHoursOperationForm(): FormGroup {
     return this.fb.group({
-      businessDays: this.fb.array(
-        [],
-        [Validators.required])
+      businessDays: this.fb.array([
+        this.fb.group({
+          startTime: [null, [Validators.required]],
```
mtpultz, line 275: We won't want to set any of these fields as required in the form state service

prime-angular-frontend/src/app/modules/site-registration/shared/services/site-form-state.service.ts
```diff
@@ -261,12 +270,49 @@ export class SiteFormStateService extends AbstractFormState<Site> {
 
   private buildHoursOperationForm(): FormGroup {
     return this.fb.group({
-      businessDays: this.fb.array(
-        [],
-        [Validators.required])
+      businessDays: this.fb.array([
```
mtpultz, line 273: We'll probably want to have validation checking that `atLeaseOne` set of hours has been entered by the user.  It doesn't need to be on the form, but should be checked at least `onSubmit` to prevent submission.  Currently we can submit without choosing hours.

prime-angular-frontend/src/app/modules/site-registration/pages/hours-operation/hours-operation.component.ts
```diff
@@ -52,30 +72,31 @@ export class HoursOperationComponent implements OnInit, IPage, IForm {
...
-    this.hasNoHours = false;
+  public getWeekDay(num: number) {
+    return WeekDay[num];
+  }
 
-    this.formUtilsService.formArrayPush(this.businessDays, businessDay);
-    const sorted = this.businessDays.value.sort(this.sortConfigByDay());
-    this.businessDays.patchValue(sorted);
+  public hasDay(index: number): boolean {
+    return (this.businessDays.value[index].startTime !== null) ? true : false;
```
mtpultz, line 90: We could pass in the group instead of the index, and check the value.  Something like `return !!group.get('startTime').value;`

prime-angular-frontend/src/app/core/resources/site-resource.service.ts
```diff
@@ -91,14 +86,11 @@ export class SiteResource {
   public updateSite(site: Site): NoContent {
     // TODO separate this out into a proper adapter
     if (site.businessHours?.length) {
+      console.log(site.businessHours)
```
mtpultz, line 89: Oops missed `console.log` :)

prime-angular-frontend/src/app/modules/site-registration/shared/services/site-form-state.service.ts
```diff
@@ -161,9 +163,16 @@ export class SiteFormStateService extends AbstractFormState<Site> {
     }
 
     if (site.businessHours?.length) {
-      const array = this.hoursOperationForm.get('businessDays') as FormArray;
-      array.clear(); // Clear out existing indices
-      this.formUtilsService.formArrayPush(array, site.businessHours);
+      const array = [];
+      for (let index = 0; index < 7; index++) {
```
mtpultz, line 167: Feels like we can drop this loop and use method chaining.  Something like `const businessDays = [...new Array(7)].reduce((days, dayOfWeek) => ..., []);`

prime-angular-frontend/src/app/modules/site-registration/pages/hours-operation/hours-operation.component.html
```diff
@@ -15,22 +15,66 @@
...
                       class="alert-content">
-          No hours of operation have been added.
+          No hours of operation have been chosen.
         </ng-container>
       </app-alert>
 
+      <ng-container formArrayName="businessDays"
+                    *ngFor="let businessDay of businessDays.controls; let i = index;">
+        <div class="row">
+          <span class="col-md-2 col-4 mt-3">{{ WeekDay[i] }}</span>
```
hogmann, line 30: that is a lot cleaner was of doing it!

PR #804, opened 2020-08-29
PRIME 987 - Hours of Operation

mtpultz:
* Updates to fix 24 hours
* Updated validations to prevent invalid hours length
* Unrelated fix for deactivate guards being triggered on patch
* Fix for address pipe formatting of postal codes


mtpultz:
> LGTM, though what does mark as pristine do?

When you patch a form it's state is marked as dirty, and we use a trigger on the canDeactivate lifecycle hook when a form is dirty so you can't leave a view even if you haven't updated it without triggering the canDeactivate model saying you're about to loss data.  Pristine is the initial state of the form (opposite of dirty) so after we programmatically update the form model we're reseting it's state to unedited.  Unlike the enrollee module we set on every page in site registration.

PR #809, opened 2020-09-01
PRIME-1025 - File Upload Label Includes Allowed File Types

james-hollinger:
First (smaller) part of ticket 851 adds the allowed file types to the idle label of Filepond. Should handle any list of allowed file types, and does not have any extra text if no allowed filetypes are supplied (in case we allow all file types in the future somewhere).

### DIFF
--- a/prime-angular-frontend/src/app/shared/components/document-upload/document-upload/document-upload.component.ts
+++ b/prime-angular-frontend/src/app/shared/components/document-upload/document-upload/document-upload.component.ts
@@ -54,7 +54,7 @@ export class DocumentUploadComponent implements OnInit {
 
     this.filePondOptions = {
       className: `prime-filepond-${this.componentName}`,
-      labelIdle: 'Click to Browse or Drop files here',
+      labelIdle: this.getIdleText(fileValidateTypeLabelExpectedTypesMap),
       fileValidateTypeLabelExpectedTypesMap,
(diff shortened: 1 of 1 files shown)


prime-angular-frontend/src/app/shared/components/document-upload/document-upload/document-upload.component.ts
```diff
@@ -78,6 +78,23 @@ export class DocumentUploadComponent implements OnInit {
...
+
+    const types = Object.values(allowedFileTypesMap);
+    let typeText = types.pop();
+
+    if (types.length > 0) {
+      const allButLast = types.join(', ');
+      typeText = `${allButLast} or ${typeText}`;
+    }
+
+    return `${baseText} Files must be ${typeText}`;
```
hogmann, line 95: To follow punctuation should we throw a '.' at the end.

james-hollinger: I was thinking about that too. if we did, the label would read (for example)

`Click to Browse or Drop files here. Files must be .pdf.`

that second period looks a little off

james-hollinger: But I just noticed that the label didn't have a period at the end before, so im gonna remove that from the base label.

PR #810, opened 2020-09-02
API Environment Cleanup

james-hollinger:
Rename + cleanup the constants file, since it now contains no constants.


mtpultz (approved):
image

PR #812, opened 2020-09-02
PRIME-843 Site Adjudication Actions

hogmann:


prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -385,5 +388,69 @@ public async Task<ActionResult> sendRemoteUsersEmail(int siteId)
...
+        /// <param name="siteId"></param>
+        [HttpPost("{siteId}/approval", Name = nameof(ApproveSite))]
+        [ProducesResponseType(typeof(ApiBadRequestResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        public async Task<ActionResult<Site>> ApproveSite(int siteId)
+        {
+            if (!User.IsAdmin())
```
james-hollinger, line 404: rather than checking this in the method, you can add the 
`[Authorize(Policy = AuthConstants.ADMIN_POLICY)]`
header to this method.

prime-angular-frontend/src/app/core/services/utils.service.ts
```diff
@@ -133,4 +133,27 @@ export class UtilsService {
   public downloadToken(token: string): void {
     window.location.assign(`${this.config.documentManagerUrl}/documents/downloads/${token}`);
   }
+
+  /**
+   * @description
+   * Opens a mailto href from your current window
```
mtpultz, line 139: The comment could be a bit more descriptive like: "Open an email using the user's default mail client."

prime-angular-frontend/src/app/core/services/utils.service.ts
```diff
@@ -133,4 +133,27 @@ export class UtilsService {
...
+   */
+  public mailTo(recipient: string, subject: string = null, body: string = null) {
+    let href = 'mailto:';
```
mtpultz, line 142: When there's not recipient the email won't be sent so we could move the assignment of `href` within the if-statement and use interpolation of mailto and recipient.

prime-angular-frontend/src/app/modules/adjudication/shared/components/site-registration-container/site-registration-container.component.ts
```diff
@@ -81,6 +82,71 @@ export class SiteRegistrationContainerComponent implements OnInit {
...
 
+  public onApprove(siteId: number) {
+    const data: DialogOptions = {
+      title: 'Approve Site Registration',
+      message: 'Are you sure you want to approve this Registration?',
+      actionText: 'Approve Site Registration',
+      component: NoteComponent
+    };
+
+    let note: string;
```
mtpultz, line 93: We could reduce the use of closures and keep the observable pure by piping `note` through the stream

prime-angular-frontend/src/app/modules/adjudication/shared/components/site-registration-container/site-registration-container.component.ts
```diff
@@ -81,6 +82,71 @@ export class SiteRegistrationContainerComponent implements OnInit {
...
+
+  public onDecline(siteId: number) {
+    const data: DialogOptions = {
+      title: 'Decline Site Registration',
+      message: 'Are you sure you want to Decline this Site Registration?',
+      actionText: 'Decline Site Registration',
+      component: NoteComponent
+    };
+
+    let note: string;
```
mtpultz, line 125: We could reduce the use of closures and keep the observable pure by piping note through the stream

prime-dotnet-webapi/Services/SiteService.cs
```diff
@@ -416,6 +436,28 @@ public async Task<BusinessLicenceDocument> GetLatestBusinessLicenceAsync(int sit
...
+            };
+
+            _context.SiteRegistrationNotes.Add(SiteRegistrationNote);
+
+            var created = await _context.SaveChangesAsync();
+            if (created < 1)
+            {
+                throw new InvalidOperationException("Could not create site registration note.");
+            }
+            // TODO: Business events for sites?
```
mtpultz, line 456: Seems like a good idea to start collecting data on sites from the start.  I would ask Jason what he might want to see in metabase.

prime-dotnet-webapi/Services/SiteService.cs
```diff
@@ -332,6 +332,26 @@ public async Task DeleteSiteAsync(int siteId)
...
+        public async Task<Site> ApproveSite(int siteId)
+        {
+            var site = await this.GetBaseSiteQuery()
+                .SingleOrDefaultAsync(s => s.Id == siteId);
+
+            if (site != null && site.ApprovedDate == null)
+            {
+                site.ApprovedDate = DateTimeOffset.Now;
+
+                this._context.Update(site);
```
james-hollinger, line 344: shouldn't need this line

prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -385,5 +388,65 @@ public async Task<ActionResult> sendRemoteUsersEmail(int siteId)
...
+                this.ModelState.AddModelError("note", "site registration notes can't be null or empty.");
+                return BadRequest(ApiResponse.BadRequest(this.ModelState));
+            }
+
+            var admin = await _adminService.GetAdminAsync(User.GetPrimeUserId());
+
+            var createdSiteRegistrationNote = await _siteService.CreateSiteRegistrationNoteAsync(siteId, note, admin.Id);
+
+            return CreatedAtAction(
+                nameof(CreateSiteRegistrationNote),
```
james-hollinger, line 446: if there is no location where we can get the note, this should just return 200 OK. This is creating a redirect url to get the note that points back at this action to create a note.

prime-dotnet-webapi/Services/SiteService.cs
hogmann: Why is that? Just going off other methods that update that have this line. As well as not having it there would make me believe context would not know of the update. Unless the "site" when we get it from the search is a reference to the original object and the context tracks changes to that object?

james-hollinger: When you pull an object from the database, entity framework tracks changes you make to the object. It already knows the object is updated. The few places we use context.Update in the code base are generally because we have a situation where a child object is being either added, modified, or removed because of the update and we want to make it explicit. There are also a few uses of context.update that are also not needed and we could remove them if we wanted to.

james-hollinger: se, for example, in the Submissionservice:
`public async Task UpdateAlwaysManualAsync(int enrolleeId, bool alwaysManual)
        {
            var enrollee = await _context.Enrollees
                .SingleAsync(e => e.Id == enrolleeId);

            enrollee.AlwaysManual = alwaysManual;
            await _context.SaveChangesAsync();
        }`

prime-angular-frontend/src/app/core/resources/site-resource.service.ts
```diff
@@ -11,6 +11,8 @@ import { LoggerService } from '@core/services/logger.service';
 import { ApiHttpResponse } from '@core/models/api-http-response.model';
 import { ToastService } from '@core/services/toast.service';
 import { NoContent } from '@core/resources/abstract-resource';
+import { BusinessDay } from '@registration/shared/models/business-day.model';
```
AnaisHebert, line 14: Business Day is imported twice

AnaisHebert: That's what is causing the build to fail

mtpultz (commented):
Looks like SiteRegistrationActions spec has an error when running `ng test`

prime-angular-frontend/src/app/modules/adjudication/shared/components/site-registration-actions/site-registration-actions.component.ts
```diff
@@ -49,15 +52,27 @@ export class SiteRegistrationActionsComponent implements OnInit {
...
-    request$.subscribe();
+  public approveSite(): void {
+    if (this.canEdit) {
+      this.approve.emit(this.siteRegistration.siteId);
+    }
+  }
+
+  public declineSite(): void {
+    if (this.canEdit) {
+      this.approve.emit(this.siteRegistration.siteId);
```
mtpultz, line 63: Looks like a copy/paste error.  This should be `this.decline.emit(...)`

mtpultz (approved):
🍦 🥪

PR #814, opened 2020-09-03
PRIME-1000 Display Uploaded Business Licence Documents

AnaisHebert:


prime-angular-frontend/src/app/modules/site-registration/pages/business-licence/business-licence.component.html
```diff
@@ -28,6 +28,19 @@
           No business licence has been uploaded.
         </ng-container>
       </app-alert>
+
+      <ng-container *ngIf="businessLicenceDocuments[0]">
```
mtpultz, line 32: If you're checking for uploaded documents you can using `businessLicenceDocuments.length` would be more explicit, otherwise you're suggesting you're specifically interested in the first index.

mtpultz (approved):
image

james-hollinger (approved):
LGTM

PR #820, opened 2020-09-04
PRIME-982 - Preliminary Programatic Role Asignment Work

james-hollinger:
A new service account for Keycloak + client in the API would enable us to programmatically assign roles to users.
Code currently works but no environement variables have been added nor Keycloak client created.


mtpultz (approved):
image

PR #826, opened 2020-09-09
PRIME-996 - Create Client for eRX

james-hollinger:
Client has been created in keycloak for an eRX integration. This PR is documentation only


mtpultz (approved):
🍦 🥪

PR #827, opened 2020-09-09
PRIME-844 Claim on Site Admin

mtpultz:
Updates include claims/disclaims, and a clean up PR will follow to reduce duplication in the web client around the modals


prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -188,6 +194,72 @@ public async Task<IActionResult> UpdateSiteCompleted(int siteId)
...
+            {
+                return NotFound(ApiResponse.Message($"Site not found with id {siteId}."));
+            }
+
+            if ((adjudicatorId != 0 && !await _adminService.AdminExistsAsync(adjudicatorId)))
+            {
+                return NotFound(ApiResponse.Message($"Admin not found with id {adjudicatorId}."));
+            }
+
+            Admin admin = (adjudicatorId != 0)
```
james-hollinger, line 224: Since adjudicatorId is optional, how about a nullable int with HasValue instead?

prime-dotnet-webapi/Services/SiteService.cs
```diff
@@ -295,6 +295,25 @@ public async Task<int> UpdateCompletedAsync(int siteId)
...
 
+        public async Task<Site> UpdateSiteAdjudicator(int siteId, Admin admin = null)
+        {
+            var site = await GetBaseSiteQuery()
+                .Include(s => s.Adjudicator)
+                .SingleOrDefaultAsync(s => s.Id == siteId);
+
+            site.Adjudicator = admin;
+
+            _context.Update(site);
```
james-hollinger, line 306: You shouldnt need to call update here, EF is tracking the object and knows its updated.
Also, shouldn't need the base site query or need to include the adjudicator.

prime-dotnet-webapi/Controllers/SitesController.cs
mtpultz: Updated

prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -188,6 +194,72 @@ public async Task<IActionResult> UpdateSiteCompleted(int siteId)
...
+        public async Task<ActionResult<Site>> SetSiteAdjudicator(int siteId, [FromQuery] int? adjudicatorId)
+        {
+            var site = await _siteService.GetSiteAsync(siteId);
+
+            if (site == null)
+            {
+                return NotFound(ApiResponse.Message($"Site not found with id {siteId}."));
+            }
+
+            if ((adjudicatorId.HasValue && !await _adminService.AdminExistsAsync((int)adjudicatorId)))
```
james-hollinger, line 219: Instead of casting a nullable, you can call adjudicatorId.Value

mtpultz: Updated

prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -188,6 +194,72 @@ public async Task<IActionResult> UpdateSiteCompleted(int siteId)
...
 
+        // PUT: api/Sites/5/adjudicator
+        /// <summary>
+        /// Add a site's assigned adjudicator.
+        /// </summary>
+        /// <param name="siteId"></param>
+        /// <param name="adjudicatorId"></param>
+        [HttpPut("{siteId}/adjudicator", Name = nameof(SetSiteAdjudicator))]
+        [Authorize(Policy = AuthConstants.ADMIN_POLICY)]
+        [ProducesResponseType(typeof(ApiBadRequestResponse), StatusCodes.Status400BadRequest)]
```
james-hollinger, line 205: These endpoints dont return BadRequest anywhere, so perhaps remove this line?

prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -188,6 +194,72 @@ public async Task<IActionResult> UpdateSiteCompleted(int siteId)
...
+
+            if ((adjudicatorId.HasValue && !await _adminService.AdminExistsAsync(adjudicatorId.Value)))
+            {
+                return NotFound(ApiResponse.Message($"Admin not found with id {adjudicatorId.Value}."));
+            }
+
+            Admin admin = (adjudicatorId.HasValue)
+                ? await _adminService.GetAdminAsync(adjudicatorId.Value)
+                : await _adminService.GetAdminAsync(User.GetPrimeUserId());
+
```
james-hollinger, line 227: At this point, if no adjudicator ID was specified, and the current user does not have an Admin object, GetAdmin will return null. Is the intention to assign null to the site's Adjudicator in that case? If not, we can remove the check on line 219 and check if admin == null here.

prime-dotnet-webapi/Controllers/SitesController.cs
mtpultz: k I'll update the enrollee endpoints as well

mtpultz: Updated

prime-dotnet-webapi/Services/SiteService.cs
mtpultz: Updated

prime-dotnet-webapi/Controllers/SitesController.cs
mtpultz: The assumption was that with the `ADMIN_POLICY` in place that they must have an `Admin` object.  I wouldn't have thought could make it this far without one.

prime-dotnet-webapi/Services/SiteService.cs
```diff
@@ -295,6 +295,19 @@ public async Task<int> UpdateCompletedAsync(int siteId)
...
         }
 
+        public async Task<Site> UpdateSiteAdjudicator(Site site, Admin admin = null)
+        {
+            site.Adjudicator = admin;
+
+            var updated = await _context.SaveChangesAsync();
+            if (updated < 1)
+            {
+                throw new InvalidOperationException($"Could not update the site adjudicator.");
```
james-hollinger, line 305: Pretty sure this will throw if you called this twice in a row (the site's Adjudicator isnt changed in the second call and so no objects are updated).  Is that intended behavior?

mtpultz: Removed markism

james-hollinger (approved):
Looks good 👍

PR #838, opened 2020-09-10
PRIME-1043 - Reassign Claim on Enrolments

mtpultz:


prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -411,21 +411,11 @@ public async Task<int> GetEnrolleeCountAsync()
                 .CountAsync();
         }
 
-        public async Task<Enrollee> UpdateEnrolleeAdjudicator(int enrolleeId, Admin admin = null)
+        public async Task<Enrollee> UpdateEnrolleeAdjudicator(int enrolleeId, Nullable<int> adminId = null)
```
james-hollinger, line 414: ```int?``` is shorthand for Nullable<int>

mtpultz: and updated similar method in SiteService

AnaisHebert (approved):
🎨

PR #847, opened 2020-09-16
PRIME-574 embedded metabase admin view

hogmann:
@ajdeziel we will need add two new environment variables for the metabase url and metabase key. And have them injected into the api. Before this ticket is complete. At that point we can remove the deafult url and key from the PrimeEnvironments file.


hogmann:
@ajdeziel Not sure if its just the Id that needs to be added to the file for the dashboard Id as an injected secret

prime-angular-frontend/src/app/modules/adjudication/pages/metabase-reports/metabase-reports.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { MetabaseReportsComponent } from './metabase-reports.component';
+
+describe('MetabaseReportsComponent', () => {
+  let component: MetabaseReportsComponent;
+  let fixture: ComponentFixture<MetabaseReportsComponent>;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [ MetabaseReportsComponent ]
```
mtpultz, line 11: Does this pass `ng test`?

prime-angular-frontend/src/app/modules/adjudication/pages/metabase-reports/metabase-reports.component.ts
```diff
@@ -0,0 +1,25 @@
...
+export class MetabaseReportsComponent implements OnInit {
+  public metabaseUrl: string;
+
+  constructor(
+    private adjudicationResource: AdjudicationResource,
+  ) { }
+
+  public ngOnInit(): void {
+    this.getMetabaseToken();
+    console.log(this.metabaseUrl);
```
mtpultz, line 18: Oops! missed console.log

prime-dotnet-webapi/PrimeEnvironment.cs
```diff
@@ -70,5 +70,12 @@ public static class AddressAutocompleteApi
             public readonly static string Url = Environment.GetEnvironmentVariable("ADDRESS_AUTOCOMPLETE_API_URL") ?? "https://ws1.postescanada-canadapost.ca/AddressComplete/Interactive/";
             public readonly static string Key = Environment.GetEnvironmentVariable("ADDRESS_AUTOCOMPLETE_API_KEY") ?? "XC95-NM62-BR64-TG37";
         }
+
+        public static class MetabaseApi
+        {
+            public readonly static string Url = Environment.GetEnvironmentVariable("METABASE_SITE_URL") ?? "https://metabase-test.pharmanetenrolment.gov.bc.ca";
+            public readonly static string Key = Environment.GetEnvironmentVariable("METABASE_SECRET_KEY") ?? "f7a2ffbc8ebd7e273603896c0f63ae04a596b94ff76c276398ce5aa8ca216cee";
+            public readonly static string DashboardId = Environment.GetEnvironmentVariable("METABASE_DASHBOARD_ID") ?? "4";
```
james-hollinger, line 78: A bit cleaner to parse this here rather than in the service. Take a look at MailServer.Port for an example.

prime-dotnet-webapi/Services/MetabaseService.cs
```diff
@@ -0,0 +1,49 @@
...
+        public string BuildMetabaseEmbeddedString()
+        {
+            var securityKey = new SymmetricSecurityKey(Encoding.UTF8.GetBytes(PrimeEnvironment.MetabaseApi.Key));
+
+            var credentials = new SigningCredentials(securityKey, SecurityAlgorithms.HmacSha256Signature);
+
+            var header = new JwtHeader(credentials);
+
+            // 10 minutes in the future represented by milliseconds from epoch
+            var expired = DateTime.Now.AddMinutes(10).Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
```
james-hollinger, line 24: I think OAuth wants seconds since Unix Epoc, not milliseconds. Also, .Net MinValue is not the same as the Unix epoch.
Not 100% sure what should be used here RE: DateTime vs DateTimeOffest (time is hard), but there is a DateTimeOffset.ToUnixTimeSeconds() method that we can use. So just

`DateTimeOffset.Now.AddMinutes(10).ToUnixTimeSeconds()`

prime-dotnet-webapi/Controllers/AdminsController.cs
```diff
@@ -98,5 +100,21 @@ public async Task<ActionResult<Admin>> GetAdminById(int adminId)
...
+        [HttpGet("embedded-metabase-url", Name = nameof(GetMetabaseEmbeddedString))]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(typeof(ApiResultResponse<Admin>), StatusCodes.Status200OK)]
+        public async Task<ActionResult<string>> GetMetabaseEmbeddedString()
+        {
+            var token = await Task.Run(() => _metabaseService.BuildMetabaseEmbeddedString());
+
+            return Ok(ApiResponse.Result(token));
```
james-hollinger, line 117: All the processing here is entirely synchronous, so we can make this controller method synchronus as well. Its not required to make controller methods async.

prime-dotnet-webapi/Controllers/AdminsController.cs
```diff
@@ -98,5 +100,21 @@ public async Task<ActionResult<Admin>> GetAdminById(int adminId)
...
         }
+
+        // GET: api/Admins/embedded-metabase-url
+        /// <summary>
+        /// Returns a time sensitive metabase embedded url string
+        /// </summary>
+        [HttpGet("embedded-metabase-url", Name = nameof(GetMetabaseEmbeddedString))]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
```
james-hollinger, line 111: This method does not return 404

prime-dotnet-webapi/Services/MetabaseService.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            var payload = new JwtPayload
+           {
+               { "resource", new DashboardResource(PrimeEnvironment.MetabaseApi.DashboardId)},
+               { "exp",  Math.Round(expired)},
```
james-hollinger, line 29: round was used in the .js example because it involved division, its not needed here

prime-dotnet-webapi/Services/MetabaseService.cs
```diff
@@ -0,0 +1,49 @@
...
+            var credentials = new SigningCredentials(securityKey, SecurityAlgorithms.HmacSha256Signature);
+
+            var header = new JwtHeader(credentials);
+
+            // 10 minutes in the future represented by milliseconds from epoch
+            var expired = DateTime.Now.AddMinutes(10).Subtract(DateTime.MinValue.AddYears(1969)).TotalMilliseconds;
+
+            var payload = new JwtPayload
+           {
+               { "resource", new DashboardResource(PrimeEnvironment.MetabaseApi.DashboardId)},
```
james-hollinger, line 28: Named classes are helpful when de-serializing JSON, but we should be able to use an anonymous class here. Try
` { "resource", new {dashboard = PrimeEnvironment.MetabaseApi.DashboardId} }`

mtpultz (commented):
image

mtpultz (approved):
🍦 🥪

james-hollinger (approved):
Looks snazzy

PR #855, opened 2020-09-22
PRIME-961 sorting on admin enrollee view (client side)

hogmann:


prime-angular-frontend/src/app/core/services/utils.service.ts
```diff
@@ -85,6 +85,39 @@ export class UtilsService {
...
+   */
+  public sortBy<T>(a: T, b: T, isAsc: boolean): number {
+    return this.sort(a, b) * (isAsc ? 1 : -1);
+  }
+
+  /**
+   * @description
+   * Generic sorting of a JSON object by key with nulls at end
+   */
+  public sortWithNullsAtEnd<T>(a: T, b: T): SortWeight {
```
mtpultz, line 100: This might be better as `sortWithTrailingNulls(...)` if we want to have a specific method for sorting `null`s, but feels like we could just by default always sort `null` the same way in `sortByOrder(...)` and might not need this method

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.ts
```diff
@@ -1,11 +1,13 @@
 import { Component, OnInit, Input, Output, EventEmitter } from '@angular/core';
 import { MatTableDataSource } from '@angular/material/table';
+import { Sort } from '@angular/material/sort';
 
 import { EnrolleeListViewModel } from '@shared/models/enrolment.model';
 import { EnrolmentStatus } from '@shared/enums/enrolment-status.enum';
 
 import { AuthService } from '@auth/shared/services/auth.service';
 import { AdjudicationRoutes } from '@adjudication/adjudication.routes';
+import { UtilsService } from '@core/services/utils.service';
```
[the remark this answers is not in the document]
hogmann: Oh, i had no idea.

AnaisHebert (commented):
Should empty dates always be on the bottom no matter if ascending or descending?

hogmann:
@AnaisHebert 
> Should empty dates always be on the bottom no matter if ascending or descending?

So it should be the biggest milliseconds -> smallest milliseconds -> null 
and
null -> smallest milliseconds -> biggest milliseconds

as the two orders

AnaisHebert (approved):
LGTM 🧀

prime-dotnet-webapi/ViewModels/Enrollee/EnrolleeUpdateModel.cs
```diff
@@ -36,6 +36,8 @@ public class EnrolleeUpdateModel
 
         public ICollection<SelfDeclaration> SelfDeclarations { get; set; }
 
+        public ICollection<SelfDeclarationDocument> SelfDeclarationDocuments { get; set; }
```
james-hollinger, line 39: Why was this added  back?

mtpultz (approved):
🍭 🍭 🍭 🍭 🍭

prime-dotnet-webapi/ViewModels/Enrollee/EnrolleeUpdateModel.cs
hogmann: I have zero clues

PR #865, opened 2020-09-24
PRIME-1039 and 1023 - Document Upload Workflow Changes & Self Declaration Document Changes for Minor Update Rule

james-hollinger:
File uploads now go to a temporary folder first. The frontend can get file metadata and can delete these files allowing users to manage file uploads before submission. Once submitted, the files are moved to a permanent location and cannot be queried/deleted.

As well, users could previously upload files to an existing self declaration without triggering the Minor Update Rule, due to the fact that documents were associated immediately rather than at submission time. Since we can now defer the association, the Minor Update Rule can be fixed.


mtpultz (approved):
🍺🍺🍺🍺🍻

AnaisHebert (approved):
🌯🌯🌯🌯

PR #867, opened 2020-09-25
PRIME-1077 Site registration wordsmithing 

AnaisHebert:


prime-angular-frontend/src/app/shared/components/prime-phone/prime-phone.component.ts
```diff
@@ -27,7 +29,12 @@ export class PrimePhoneComponent {
   }
 
   private get primePhone() {
-    return this.config.prime.phone;
+    if(this.phoneNumber == 'prime') {
```
mtpultz, line 32: You'll never want to use `==`.  You'll always want to use `===`.  By using `==` you open yourself up to coercion when you're trying to compare it won't cause an issue here, but you will always want to use `===` in JavaScript.

AnaisHebert: oops!

mtpultz (approved):
image

PR #870, opened 2020-09-30
Fix Backend Tests

james-hollinger:
Fix various issues in mocks/tests to get them working again. Added one new test for the Minor Update Rule RE: document GUIDs on Self Declarations.


AnaisHebert (approved):
🍚🍚🧁

PR #871, opened 2020-09-30
PRIME-1096 Update database, and ensure enrollee TOA still works

AnaisHebert:


prime-dotnet-webapi/ApiDbContext.cs
```diff
@@ -204,11 +204,27 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+                .WithMany(e => e.Agreements)
+                .HasForeignKey(toa => toa.OrganizationId);
+
+            modelBuilder.Entity<Agreement>()
+                .HasOne(toa => toa.Party)
+                .WithMany(e => e.Agreements)
+                .HasForeignKey(toa => toa.PartyId);
+
+            modelBuilder.Entity<Agreement>()
+                .HasCheckConstraint("CHK_OnlyOneForeignKey",
```
james-hollinger, line 223: Do Check Constraints have a similar naming convention to PK and FK constraints where the name of the table is in the name? Something like "CHK_Agreement_OnlyOneForeignKey"

prime-dotnet-webapi/Controllers/EnrolleesAccessTermsController.cs
```diff
@@ -49,8 +49,8 @@ public class EnrolleesAccessTermsController : ControllerBase
         [ProducesResponseType(StatusCodes.Status401Unauthorized)]
         [ProducesResponseType(StatusCodes.Status403Forbidden)]
         [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
-        [ProducesResponseType(typeof(ApiResultResponse<IEnumerable<AccessTerm>>), StatusCodes.Status200OK)]
-        public async Task<ActionResult<IEnumerable<AccessTerm>>> GetAccessTerms(int enrolleeId, [FromQuery] AccessTermFilters filters)
+        [ProducesResponseType(typeof(ApiResultResponse<IEnumerable<Agreement>>), StatusCodes.Status200OK)]
+        public async Task<ActionResult<IEnumerable<Agreement>>> GetAccessTerms(int enrolleeId, [FromQuery] AccessTermFilters filters)
```
james-hollinger, line 53: Shall we rename AccessTermFilters to AgreementFilters while we're at it?

prime-dotnet-webapi/Controllers/EnrolleesAccessTermsController.cs
```diff
@@ -96,7 +96,7 @@ public async Task<ActionResult<AccessTerm>> GetAccessTerm(int enrolleeId, int ac
                 return Forbid();
             }
 
-            AccessTerm accessTerm = await _accessTermService.GetEnrolleeAccessTermAsync(enrolleeId, accessTermId, true);
+            Agreement accessTerm = await _accessTermService.GetEnrolleeAccessTermAsync(enrolleeId, accessTermId, true);
```
james-hollinger, line 99: Could the variable be renamed as well?

AnaisHebert: I was just doing the minimum database changes for org agreements required to allow others to start working on this as well. Still lots to change in the code itself

prime-dotnet-webapi/Migrations/20200929200820_AgreementRenameandAddOrgAndParty.cs
```diff
@@ -0,0 +1,245 @@
...
+            name: "IX_AccessTerm_EnrolleeId",
+            table: "AccessTerm");
+
+            migrationBuilder.DropIndex(
+                name: "IX_AccessTerm_LimitsConditionsClauseId",
+                table: "AccessTerm");
+
+            migrationBuilder.RenameTable(
+                name: "Agreement",
+                schema: null,
```
james-hollinger, line 37: Don't need to specify schema:null, its got a default value of null

prime-dotnet-webapi/Migrations/20200929200820_AgreementRenameandAddOrgAndParty.cs
```diff
@@ -0,0 +1,245 @@
...
+                newName: "AgreementVersion");
+
+            migrationBuilder.RenameTable(
+                name: "AccessTerm",
+                schema: null,
+                newName: "Agreement");
+
+            migrationBuilder.RenameColumn(
+                name: "AgreementId",
+                table: "Agreement",
```
james-hollinger, line 47: I think this is incorrect, if the migrations work the way I think they do. This is modifying the new Agreement table (the one previously called AccessTerms), rather than the old Agreement table (the one now called AgreementVersion). this line should be `table: "AgreementVersion`

AnaisHebert: This was previously called AgreementId on the AccessTerm table, now the AccessTerm table is called Agreement and AgreementId is actually AgreementVersionId, this links Agreement to AgreementVersion. This should be updating the new Agreement table. Renaming a table to one that already existed before is confusing haha

prime-dotnet-webapi/Controllers/EnrolleesAccessTermsController.cs
james-hollinger: fair

prime-dotnet-webapi/ApiDbContext.cs
```diff
@@ -204,11 +204,27 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+                    "( CASE WHEN \"EnrolleeId\" IS NULL THEN 0 ELSE 1 END"
```
james-hollinger, line 224: You could use a string literal here to make this look slightly nicer. 
`@"(CASE WHEN ""EnrolleeId"" IS NULL THEN 0 ELSE 1 END
   + CASE WHEN ""OrganizationId"" IS NULL THEN 0 ELSE 1 END
   + CASE WHEN ""PartyId"" IS NULL THEN 0 ELSE 1 END) = 1"`

prime-dotnet-webapi/Models/Agreements/AgreementVersionType.cs
```diff
@@ -0,0 +1,11 @@
+namespace Prime.Models
+{
+    public enum AgreementVersionType
+    {
+        CommunityPharmacistTOA = 1,
+        RegulatedUserTOA = 2,
+        OboTOA = 3,
```
james-hollinger, line 7: image
🎵Well, OboToa hasn't always been this glam, I was a weekday MOA once🎵

prime-dotnet-webapi/Migrations/20200929200820_AgreementRenameandAddOrgAndParty.cs
```diff
@@ -0,0 +1,243 @@
...
+
+            migrationBuilder.AddForeignKey(
+                name: "FK_Agreement_Party_PartyId",
+                table: "Agreement",
+                column: "PartyId",
+                principalTable: "Party",
+                principalColumn: "Id",
+                onDelete: ReferentialAction.Cascade);
+
+            migrationBuilder.CreateCheckConstraint(
```
james-hollinger, line 125: This is in both migrations, which is odd

prime-dotnet-webapi/Migrations/20200930204650_AgreementVersionType.cs
```diff
@@ -0,0 +1,258 @@
...
+
+            migrationBuilder.DropForeignKey(
+                name: "FK_Agreement_Party_PartyId",
+                table: "Agreement");
+
+            migrationBuilder.DropCheckConstraint(
+                name: "CHK_Agreement_OnlyOneForeignKey",
+                table: "Agreement");
+
+            migrationBuilder.CreateCheckConstraint(
```
james-hollinger, line 25: This is in both migrations, which is odd

AnaisHebert: Yeah the first one was mostly had written cause ef doesnt know how to rename things, and im guessing it redid the foreign keys and check constraint because it wanted to do whatever this is:
`migrationBuilder.AlterColumn<int>(
                name: "PartyId",
                table: "Agreement",
                nullable: true,
                oldClrType: typeof(int),
                oldType: "integer");`

prime-dotnet-webapi/Controllers/EnrolleesAccessTermsController.cs
mtpultz: If we're going to do the minimum to unblock we should keep a list of what needs to found and renamed like local variables (accessTerm) and services (accessTermService) so we don't forget (or overlook) them later in the sprint.

AnaisHebert: Added to the backend subtask

james-hollinger (approved):
Fantastic job, thanks for doing the migrations

mtpultz (approved):
image

PR #880, opened 2020-10-08
PRIME-1038 | Document Manager Healthcheck

ajdeziel:
## What's New
* `healthcheck.py`: Contains utility functions for performing healthchecks to PostgreSQL and Redis from Flask. Logical reason of not performing full scale checks is that if Flask app can initiate successful connection requests to both PostgreSQL and Redis, then inherently Document Manager must be healthy as it was able to initiate both requests while also remaining up itself.

## What's Changed
* `requirements.txt`: Added `py-healthcheck` library to support healthcheck functionality
* `routes.py`: Refactored `/health` endpoint to `/healthcheck`, and made it leverage newly built healthcheck functionality.


mtpultz (approved):
LGTM

PR #881, opened 2020-10-08
PRIME-1088 | Restore hot reloading for Document Manager

ajdeziel:


ajdeziel:
**Note**: There is no way presently to keep the `document-manager-migrate` container up following execution, despite following Docker's best practices and recommendations.

After the `document-manager-migrate` container is done executing and shuts down, migrations that need to be tested will need to be done via the `document-manager-backend` container. This is not an impediment, as the migration container is simply a copy of the backend container.

mtpultz:
@ajdeziel I played around with the dockerfile for a bit, but skipped trying to change the permissions during the Docker build, and updated the permissions using `git update-index --chmod=+x ./init-backend.sh` and with those set on the host it will be used directly in Docker.  Also, directly behind this error was a similar error for `init-migrate` so both of those files have had their permissions changed.  I think this puts a cap on this task and we can merge it in and close it out.

mtpultz (approved):
The changes I added have been tested and verified on Mac.  image

@AnaisHebert  to get the updates you'll need to completely blow away your images, and reset to ensure you get the permission changes.

PR #887, opened 2020-10-09
PRIME-1121 Login Module Redirects

mtpultz:


prime-angular-frontend/src/app/modules/auth/shared/guards/authorization-redirect.guard.ts
```diff
@@ -32,10 +33,15 @@ export class AuthorizationRedirectGuard extends BaseGuard {
         return resolve(true);
       }
 
+      console.log(routePath);
```
hogmann, line 36: Should this still be here?

hogmann (commented):
LGTM other than console.log

prime-angular-frontend/src/app/modules/auth/shared/guards/authorization-redirect.guard.ts
mtpultz: Definitely not :P

AnaisHebert (approved):
🥨

PR #888, opened 2020-10-10
PRIME-1069 Remove the Remote Access Screen from the Site Reg workflow when Community Pharmacy is selected as the care setting

mtpultz:


AnaisHebert (approved):
Looks good!

PR #893, opened 2020-10-16
PRIME-1135 Updates/Fixes Enrolments and Site Registration

mtpultz:
- Ported changes from PROD
- Fix enrolment routing
- Show remote user access on refresh/reload if chosen
- Added missing input bindings for logout redirect
- Added redirect to demographic for BCeID
- Fix for BCeID not allowing province being set


AnaisHebert (approved):
🍷🍷🍷

PR #907, opened 2020-10-28
PRIME-1140 and PRIME-1018 - College and Licence Updates

james-hollinger:
1. Renamed the College of Nurses
2. New colleges (with no college prefix)
3. New Nursing licenses
4. Two existing licenses are now in the CollegeLicenseLookup table (bugfix)
5. FE changes for new colleges (since they have no licenses to select and no college prefixes)
6. Renamed RegulatedUser => NamedInImReg
6. Added column LicensedToProvideCare


mtpultz (approved):
space-man-spiff-phasor

PR #915, opened 2020-10-30
PRIME-1178 - Manual Rule for No Assigned Agreement 

james-hollinger:
If we cannot determine your agreement type, you go to manual.


prime-dotnet-webapi/Models/Statuses/StatusReasonType.cs
```diff
@@ -17,5 +17,6 @@ public enum StatusReasonType
         AssuranceLevel = 13,
         IdentityProvider = 14,
         RequestingRemoteAccess = 15,
+        NoAssignedAgreement = 16,
```
mtpultz, line 20: We'll want to update the enum in the web client

AnaisHebert (approved):
🍿🍿🍿

PR #920, opened 2020-11-02
PRIME-1133 Add Remote Access Indicator in Admin

AnaisHebert:


prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-enrolments/enrollee-enrolments.component.ts
```diff
@@ -12,6 +12,7 @@ import { EnrolmentStatus } from '@shared/enums/enrolment-status.enum';
 
 import { AdjudicationRoutes } from '@adjudication/adjudication.routes';
 import { AdjudicationResource } from '@adjudication/shared/services/adjudication-resource.service';
+import { EnrolmentCard } from '@shared/models/enrolment-card.model';
```
mtpultz, line 15: Probably want to group shared imports for ease of scanning and future lazy loading updates

prime-angular-frontend/src/app/shared/components/enrollee/enrollee-review/enrollee-review.component.ts
```diff
@@ -108,7 +112,8 @@ export class EnrolleeReviewComponent {
       : [];
   }
 
-  public onRoute(routePath: string): void {
+  public onRoute(routePath: string | (string | number)[], event: Event = null): void {
+    event?.preventDefault();
```
mtpultz, line 116: Did you find the event wasn't available?  I would have thought the event would always be present since it bubbles up from the element.  So I would suspected it doesn't need a default value or a null check.

mtpultz (approved):
image

prime-dotnet-webapi/ViewModels/Enrollee/EnrolleeListViewModel.cs
```diff
@@ -57,5 +58,16 @@ public string CurrentTOAStatus
...
         public bool AlwaysManual { get; set; }
+
+        [JsonIgnore]
+        public ICollection<EnrolleeRemoteUser> EnrolleeRemoteUsers { get; set; }
+
+        public bool RemoteAccess
+        {
+            get
+            {
+                return (EnrolleeRemoteUsers.Count > 0) ? true : false;
```
james-hollinger, line 69: You don't need to do a ternary here, you can just do
`return (EnrolleeRemoteUsers.Count > 0);`

PR #932, opened 2020-11-05
PRIME-1007 Site notification template

mtpultz:
* Added notification template
* Implemented on site card only
* Implementation on org card and remaining tasks for PRIME-1007, but there are questions on how we would detect this situation without a new flag being added in the database
* Fixed overview page to show organization information on initial registration


hogmann (approved):
🍰 🚁

PR #934, opened 2020-11-06
PRIME-1018 - Full Agreement Logic

james-hollinger:
A more robust system for assigning Agreement Type.

Tests will be added once tests are un-broken.


mtpultz (approved):
calvin-backflip

PR #938, opened 2020-11-09
Store requestedRemoteAccess as boolean on Submission

AnaisHebert:


prime-dotnet-webapi/ViewModels/Enrollee/EnrolmentCardViewModel.cs
```diff
@@ -19,16 +17,11 @@ public string RemoteAccess
...
                 {
-                    if (enrolleeRemoteUsers.Count > 0)
-                    {
-                        return (AgreementAcceptedDate == null)
-                            ? "User Requested Remote Access"
-                            : "User Approved Remote Access";
-                    }
+                    return (AgreementAcceptedDate == null)
+                        ? "User Requested Remote Access"
+                        : "User Approved Remote Access";
```
james-hollinger, line 24: In the discussions we had with Jolie, there is no difference when the enrollment is accepted or not, they just want to see if the submission had remote access at the time or not. This view model could just return the bool

mtpultz (approved):
image

james-hollinger (approved):
LGMT

PR #939, opened 2020-11-09
PRIME-1103 fire business event with admin document

hogmann:


prime-dotnet-webapi/Services/BusinessEventService.cs
```diff
@@ -100,6 +100,20 @@ public async Task<BusinessEvent> CreateEnrolleeEventAsync(int enrolleeId, string
             return businessEvent;
         }
 
+        public async Task<BusinessEvent> CreateSiteEventAsync(int siteId, string description)
```
mtpultz, line 103: With these having almost the same method body we could have a private method that is passed a `BusinessEvent` from each of these methods to reduce code duplication.

PR #946, opened 2020-11-12
PRIME-# Added fix for race condition between guards and the config resolver

mtpultz:


hogmann (approved):
LGTM, throttled down to pretty slow and nothing went wrong.

PR #951, opened 2020-11-13
Config Update

mtpultz:
* Updated config models with missing keys
* Updated mock config service
* Added new export of lookups


AnaisHebert (approved):
🚵🏼

PR #952, opened 2020-11-13
PRIME-1183 - Agreement Engine Bug

james-hollinger:
An .Include was missing from the load for the enrollee, causing a bug in both the profile snapshot and agreement engine. Trying new DTO paradigm for engines.


mtpultz (approved):
image

PR #962, opened 2020-11-18
Prune Environment Variables

james-hollinger:
Drop some unused environment variables from the API


ajdeziel (approved):
Alt Text

mtpultz (approved):
space-man-spiff-phasor

PR #963, opened 2020-11-18
PRIME-1184 Email Statuses

AnaisHebert:


prime-dotnet-webapi/Controllers/EmailsController.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+
+using Prime.Auth;
+using Prime.Services;
+
+namespace Prime.Controllers
+{
+    [Produces("application/json")]
+    [Route("api/emails")]
```
james-hollinger, line 12: this can just be `[Route("api/[controller]")]`

prime-dotnet-webapi/Controllers/EmailsController.cs
```diff
@@ -0,0 +1,39 @@
...
+        public EmailsController(IEmailService emailService)
+        {
+            _emailService = emailService;
+        }
+
+        // GET: api/emails/management/statuses
+        /// <summary>
+        /// Update all logged email statuses sent using the CHES email service
+        /// </summary>
+        [HttpGet("management/statuses", Name = nameof(UpdateEmailLogStatuses))]
```
james-hollinger, line 27: I think a POST here would be better, we may actually need/want a GET on these in the future

prime-dotnet-webapi/Controllers/EmailsController.cs
```diff
@@ -0,0 +1,39 @@
...
+        [Authorize(Roles = AuthConstants.PRIME_API_SERVICE_ACCOUNT_ROLE + "," + AuthConstants.ADMIN_POLICY)]
```
james-hollinger, line 28: As written, this wont allow admins to access this endpoint. I assume this is a typo, you need the Admin role not the admin policy (PRIME_ADMIN_ROLE). 

I have a slight preference for interpolated string for this kind of thing, but its not that much of a difference.

`[Authorize(Roles = $"{AuthConstants.PRIME_API_SERVICE_ACCOUNT_ROLE},{AuthConstants.PRIME_ADMIN_ROLE}")]`

prime-dotnet-webapi/HttpClients/Email/CHESClient.cs
```diff
@@ -3,30 +3,35 @@
...
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 
 namespace Prime.HttpClients
 {
     public class ChesClient : IChesClient
     {
         private static HttpClient _client;
```
james-hollinger, line 13: I am pretty sure this shouldn't be static, I've just not noticed before.

prime-dotnet-webapi/HttpClients/Email/CHESClient.cs
```diff
@@ -35,30 +40,61 @@ public async Task SendAsync(string from, IEnumerable<string> to, IEnumerable<str
...
-                response = await _client.PostAsync("email", requestContent);
+                HttpResponseMessage response = await _client.GetAsync($"status?msgId={msgId}");
+                var responseString = await response.Content.ReadAsStringAsync();
 
                 if (response.IsSuccessStatusCode)
                 {
-                    var responseJsonString = await response.Content.ReadAsStringAsync();
-                    JsonConvert.DeserializeObject<EmailSuccessResponse>(responseJsonString);
+                    var statusResponse = JsonConvert.DeserializeObject<IList<StatusResponse>>(responseString);
+                    return statusResponse[0].Status;
```
james-hollinger, line 76: If we expect this to only ever return a single status response, it would be more correct to use `statusResponse.Single().Status` instead. If there is maybe more than one but we only care about the first, then we can use `statusResponse.First().Status`

mtpultz (approved):
image

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -304,14 +297,29 @@ public async Task<IEnumerable<string>> GetPharmaNetProvisionerNamesAsync()
                 .ToListAsync();
         }
 
-        private async Task Send(string from, string to, string subject, string body)
+        public async Task<bool> UpdateEmailLogStatuses()
         {
-            await Send(from, new[] { to }, new string[0], subject, body, Enumerable.Empty<(string Filename, byte[] Content)>());
+            var emailLogs = await _context.EmailLogs
+                .Where(e => e.SendType == "CHES"
+                    && e.MsgId != Guid.Empty
```
james-hollinger, line 304: MsgId is nullable, so I assume this should be `e.MsgId != null`

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -304,14 +297,29 @@ public async Task<IEnumerable<string>> GetPharmaNetProvisionerNamesAsync()
...
+                    && e.LatestStatus != "completed")
+                .ToListAsync();
+
+            foreach (var email in emailLogs)
+            {
+                var status = await _chesClient.GetStatusAsync((Guid)email.MsgId);
```
james-hollinger, line 310: rather than casing, you should be able to just do `email.MsgId` or `email.MsgId.Value`

james-hollinger (approved):
L ♟ T M

PR #964, opened 2020-11-19
PRIME-1208 Create PHSA Labtech Module

mtpultz:
Adds the initial PHSA Labtech lazy loaded module with example page component, resource, and routing


AnaisHebert (approved):
Wooohoo🏓

PR #965, opened 2020-11-19
Registration ID is optional

mtpultz:


AnaisHebert (approved):
☄️🔥💥

mtpultz:
Ultimately doesn't matter if the validator exists since the field is disabled so no validations are run on it.  So dropping the changes so if in the future it is possible to toggle enable/disable that it will be required.

PR #967, opened 2020-11-20
Implemented PRIME-1189 (Advanced practice certification options need …

neophyte57:
…to be greyed out except for RN licenses)

### DIFF
--- a/prime-angular-frontend/src/app/shared/components/college-certification-form/college-certification-form.component.ts
+++ b/prime-angular-frontend/src/app/shared/components/college-certification-form/college-certification-form.component.ts
@@ -10,6 +10,7 @@ import { ViewportService } from '@core/services/viewport.service';
 import { FormUtilsService } from '@core/services/form-utils.service';
 import { CollegeLicenceClass } from '@shared/enums/college-licence-class.enum';
 import { EnrolmentService } from '@enrolment/shared/services/enrolment.service';
+import { NursingLicenseCode } from '@shared/enums/nursing-license-code.enum';
 
 @Component({
   selector: 'app-college-certification-form',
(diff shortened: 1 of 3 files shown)


prime-angular-frontend/src/app/shared/components/college-certification-form/college-certification-form.component.ts
```diff
@@ -97,6 +98,19 @@ export class CollegeCertificationFormComponent implements OnInit {
     this.remove.emit(this.index);
   }
 
+  public shouldShowPractices(): boolean {
+    if ((this.collegeCode.value == CollegeLicenceClass.BCCNM) &&
```
mtpultz, line 102: Feels like this could be simplified down to a return statement since the result of the if-statement is a boolean

prime-angular-frontend/src/app/shared/components/college-certification-form/college-certification-form.component.ts
```diff
@@ -97,6 +98,19 @@ export class CollegeCertificationFormComponent implements OnInit {
...
+        (this.licenseCode.value == NursingLicenseCode.NON_PRACTICING_REGISTERED_NURSE ||
```
mtpultz, line 103: Feels like the set of OR'd values could be reduced down to an array of enum values and use the include method to check if the value exists in the array - `[...].includes(value)`

prime-angular-frontend/src/app/shared/components/college-certification-form/college-certification-form.component.ts
```diff
@@ -10,6 +10,7 @@ import { ViewportService } from '@core/services/viewport.service';
 import { FormUtilsService } from '@core/services/form-utils.service';
 import { CollegeLicenceClass } from '@shared/enums/college-licence-class.enum';
 import { EnrolmentService } from '@enrolment/shared/services/enrolment.service';
+import { NursingLicenseCode } from '@shared/enums/nursing-license-code.enum';
```
mtpultz, line 13: We try and group the imports based on module as much as possible.  It makes it easier to scan the dependencies when they are ordered by module.

mtpultz (approved):
image

PR #969, opened 2020-11-20
Schedule B OCP Workplace

itchison:


documentation/Architecture.md
```diff
@@ -14,4 +14,6 @@ PRIME is a traditional N-Tier application architecture. It's composed of a front
 
 ## Architecture Diagram
 The following is a high level application component diagram of the PRIME application
-![alt text](images/architecture.png "PRIME Architecture")
\ No newline at end of file
+![alt text](images/architecture.png "PRIME Architecture")
+
+A change to create a PR
```
mtpultz, line 19: This seems overly ambitious and it might be better to split it up into multiple tickets

mtpultz (changes requested):

itchison:
Was only created to spin up an env to view the data.

PR #971, opened 2020-11-23
Feature/prime 1223 add phsa landing page

neophyte57:


prime-angular-frontend/src/app/modules/auth/auth-routing.module.ts
```diff
@@ -41,6 +43,11 @@ const routes: Routes = [
         component: SiteComponent,
         data: { title: 'Welcome to PRIME' }
       },
+      {
+        path: AuthRoutes.PHSA,
+        component: PhsaComponent,
+        data: { title: 'Enrol for access to PHSA eForms', subtitle: 'Request approval to access PHSA eForms', button: 'Logon with BCSC' }
```
mtpultz, line 49: You won't want to pass in the button label through the route configuration that feels more like a direct template entry.  Likely the title is enough which will match the browser tab and page title.

prime-angular-frontend/src/app/modules/auth/pages/phsa/phsa.component.ts
```diff
@@ -0,0 +1,57 @@
...
+})
+export class PhsaComponent implements OnInit {
+
+  private routeData: Data;
+
+  constructor(
+    @Inject(APP_CONFIG) private config: AppConfig,
+    private authService: AuthService,
+    private viewportService: ViewportService,
+    activatedRoute: ActivatedRoute
```
mtpultz, line 26: You'll want to add private to dependency injected items and provide access to parts of the item data through member variables or methods.  Anything that is public is exposed in the template, and they typically aren't exposed outside the controller.

prime-angular-frontend/src/app/modules/auth/pages/phsa/phsa.component.ts
```diff
@@ -0,0 +1,57 @@
...
+  ) {
+    this.routeData = activatedRoute.snapshot.data;
+  }
+
+  public get loginButtonMarkup(): string {
+    return this.routeData.button;
+  }
+
+  public get subtitle(): string {
```
mtpultz, line 35: You can definitely provide this through a getter, but there's actually a bit of performance to be had if it can be made into a member variable since this method will be invoked on every digest cycle.  If it was me I'd probably have this as a `public title: string;` and then set in the constructor.

mtpultz (approved):

PR #981, opened 2020-11-25
Remove State Machine from SubmissionStateEngine

james-hollinger:
C# 8's new switch expression makes our state machine syntax kinda unnecessary. So I removed it.


neophyte57:
I think it would be useful to add a comment that references the documentation for the newish syntax (e.g. https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#switch-expressions) even if the feature was available in September 2019

AnaisHebert (approved):
Wooo its beautiful🍰

johnnyl6 (approved):
👍

mtpultz (approved):
🐕‍🦺

PR #982, opened 2020-11-25
PRIME-1078 Angular Version Bump from v9.0 to v10.2

mtpultz:
* For your sanity while reviewing skip the `*.spec.ts files` they are all the same change due to a deprecation.  That amounts to about 140 files of changes removed from the review.  Since you can't filter directly on `*.spect.ts` I'd suggest searching the page for `.spec.ts` so they are highlighted and skim over the changes
* You will also see fixes to linting errors around: files not ending in a newline, whitespace on ends of lines, usage of == instead of ===, and a few others that turned up in `ng lint`


prime-angular-frontend/package.json
```diff
@@ -19,68 +19,67 @@
...
     "filepond": "^4.22.1",
     "filepond-plugin-file-validate-size": "^2.2.1",
     "filepond-plugin-file-validate-type": "^1.2.5",
-    "keycloak-angular": "^7.0.1",
-    "keycloak-js": "^6.0.1",
-    "moment": "^2.24.0",
-    "ng-busy": "^9.0.0",
-    "ngx-clipboard": "^12.2.1",
+    "keycloak-angular": "^8.0.1",
+    "keycloak-js": "^10.0.2",
```
mtpultz, line 41: @james-hollinger doesn't seem to affect the application bumping the version, but want to draw your eye to the update since if this isn't possible then Keycloak is the only thing holding back upgrading the front-end framework and dependencies.

johnnyl6 (approved):
Looks like a straight and exiting upgrade to me with some method name refactoring.

AnaisHebert (approved):
🤹‍♂️

PR #995, opened 2020-11-28
PRIME-1207 - Assign Keycloak Attributes and Roles

james-hollinger:
New client to push up user attributes and set roles in Keycloak, primarily for the PHSA eForms Labtechs coming in. New BaseClient to be refactored into existing clients in another PR to clean up serialization. 

Updates to Keycloak:
1. New Keycloak Administration service account/client. Can assign roles and attributes to users.
2. New phsa_eforms_labtech role
3. New phsa-eforms client
4. New mapper under the "phone" scope that includes the optional phoneExtension attribute.

Env variables have been added to dev test and prod


mtpultz (approved):
image

PR #1007, opened 2020-12-03
PRIME-1226 Add route guards for PHSA module

neophyte57:


mtpultz (approved):
The only thing I noticed was in the dashboard for PHSA that the logout redirect URL is incorrect, but that's on me :) 

Looks great!

PR #1015, opened 2020-12-09
PRIME-1232 Verifiable Credential claims list 

AnaisHebert:
MicrosoftTeams-image (2)
can't make a list of items, can have multiple claims with the same key


hogmann (approved):
LGTM

PR #1019, opened 2020-12-14
PRIME-1032 status display for user on site registration

neophyte57:
- Note that there are some comments in the Jira ticket (https://bcgovmoh.atlassian.net/browse/PRIME-1032) that are of relevance 
- I have not merged `develop` into this branch because there is apparently a bug in the `develop` code related to Site Registration


prime-angular-frontend/src/app/modules/site-registration/pages/overview/overview.component.html
```diff
@@ -1,5 +1,6 @@
 <app-page [busy]="busy">
   <app-page-header>PharmaNet Site Registration</app-page-header>
+  <!-- <app-site-progress-indicator [inProgress]="!isCompleted"></app-site-progress-indicator> -->
```
AnaisHebert, line 3: forgot commented code?

neophyte57: Pilar wants the progress bar added to the Review screen but said that she will create a new Jira ticket for this.  I left it commented-out for now

prime-angular-frontend/src/app/modules/site-registration/pages/site-management/site-management.component.ts
```diff
@@ -94,6 +97,8 @@ export class SiteManagementComponent implements OnInit {
   }
 
   public getOrganizationProperties(organization: OrganizationListViewModel): { key: string, value: string }[] {
+    // TODO: Why are these methods called so often despite no user interaction?
+    // this.logger.trace("getOrganizationProperties");
```
AnaisHebert, line 101: should remove these commented out logs

prime-angular-frontend/src/app/modules/site-registration/pages/overview/overview.component.html
AnaisHebert: I think if there will be a separate ticket we should remove it from here so we aren't merging in commented code if avoidable.

neophyte57: OK done

prime-angular-frontend/src/app/modules/site-registration/pages/site-management/site-management.component.ts
neophyte57: OK done

prime-angular-frontend/src/app/modules/site-registration/pages/overview/overview.component.ts
```diff
@@ -97,4 +110,27 @@ export class OverviewComponent implements OnInit {
...
   }
+
+  /**
+   * @description
+   * Infer whether user was required to accept an Organization Agreement.
+   * User saw an Organization Agreement if the current Site has a care setting that is unique
+   * for that Organization.
+   */
+  private determineOrgAgreementRequired$(): Observable<SiteListViewModel[]> {
+    const currentCareSettingCode: number = this.site.careSettingCode;
```
hogmann, line 121: We do have an enum (CareSettingEnum) we can use here to have stronger typing on these variables

prime-angular-frontend/src/app/modules/site-registration/pages/overview/overview.component.ts
```diff
@@ -97,4 +110,27 @@ export class OverviewComponent implements OnInit {
...
+
+    return this.siteResource.getSites(this.organization.id).pipe(
+      tap(sites => {
+        sites.forEach(
+          site => {
+            this.logger.trace("doingBusinessAs", site.doingBusinessAs);
```
hogmann, line 127: This will only log in local, but i'm not sure if we want to leave these in as they don't seem super relevant outside debugging this PR.

prime-angular-frontend/src/app/modules/site-registration/pages/overview/overview.component.ts
```diff
@@ -97,4 +110,27 @@ export class OverviewComponent implements OnInit {
...
     }
   }
+
+  /**
+   * @description
+   * Infer whether user was required to accept an Organization Agreement.
+   * User saw an Organization Agreement if the current Site has a care setting that is unique
+   * for that Organization.
+   */
+  private determineOrgAgreementRequired$(): Observable<SiteListViewModel[]> {
```
hogmann, line 120: This method naming doesn't reflect what it is returning. I would assume this would return a bool, when instead it returns a SiteListViewModel[]. Is this return relevant? Since i see we just ignore the return above.

As below it looks like we just use this method to increment a private class variable that doesn't seem intuitive. 

Maybe just making this method return true/false and containing all the logic instead of splitting it across mutliple methods and hiding some of the meaning within. That way above we just call 

```
if (this.determineOrgAgreementRequired()) {
   do x;
} else {
   do y
}

private determineOrgAgreementRequired(): boolean {}
```

hogmann (changes requested):

prime-angular-frontend/src/app/modules/site-registration/pages/overview/overview.component.ts
neophyte57: Completely agree @hogmann .  However I was having problems getting this to work with RxJs so I ended up with the member variable hack to get functionality to work.  I'll join the Zoom workspace tomorrow to get your assistance  :-)

neophyte57: Addressed in latest commit 68fb52e

prime-angular-frontend/src/app/modules/site-registration/pages/overview/overview.component.ts
neophyte57: Addressed in latest commit 68fb52e

prime-angular-frontend/src/app/modules/site-registration/pages/overview/overview.component.ts
neophyte57: Addressed in latest commit 68fb52e

prime-angular-frontend/src/app/modules/site-registration/pages/site-management/site-management.component.html
```diff
@@ -81,6 +81,14 @@
     </div>
   </ng-template>
 
+  <ng-template #success_notification
+               let-props="props">
+    <div class="d-flex align-self-center mt-2">
+      <mat-icon class="mr-2 green-icon">{{ props.icon }}</mat-icon>
+      <p class="text-green" [innerHTML]="props.text"></p>
```
mtpultz, line 88: Looks like your HTML formatting isn't working either attributes should auto-align vertically:

```html
<p class="text-green" 
   [innerHTML]="props.text"></p>
```

prime-angular-frontend/src/app/modules/site-registration/pages/overview/overview.component.ts
```diff
@@ -12,9 +12,10 @@ import { ConfirmDialogComponent } from '@shared/components/dialogs/confirm-dialo
 
 import { SiteService } from '@registration/shared/services/site.service';
 import { SiteRoutes } from '@registration/site-registration.routes';
-import { Site } from '@registration/shared/models/site.model';
+import { Site, SiteListViewModel } from '@registration/shared/models/site.model';
 import { Organization } from '@registration/shared/models/organization.model';
 import { OrganizationService } from '@registration/shared/services/organization.service';
+import { LoggerService } from '@core/services/logger.service';
```
mtpultz: Should be up a bit higher with the other @core dependencies

prime-angular-frontend/src/app/modules/site-registration/pages/overview/overview.component.ts
mtpultz: TypeScript should infer number so shouldn't have to explicitly state the type

prime-angular-frontend/src/app/modules/site-registration/pages/overview/overview.component.ts
```diff
@@ -82,13 +87,21 @@ export class OverviewComponent implements OnInit {
   }
 
   public nextRoute(): void {
-    this.routeUtils.routeTo([SiteRoutes.MODULE_PATH, SiteRoutes.SITE_MANAGEMENT], {
-      queryParams: { submitted: true }
-    });
+    this.busy = this.determineOrgAgreementRequired$().subscribe(
+      ignored => {
```
mtpultz: If you want to ignore a function parameter its convention in JavaScript or TypeScript that you use an underscore - `(_) => { }`

prime-angular-frontend/src/app/modules/site-registration/pages/overview/overview.component.ts
```diff
@@ -97,4 +110,27 @@ export class OverviewComponent implements OnInit {
...
+   * @description
+   * Infer whether user was required to accept an Organization Agreement.
+   * User saw an Organization Agreement if the current Site has a care setting that is unique
+   * for that Organization.
+   */
+  private determineOrgAgreementRequired$(): Observable<SiteListViewModel[]> {
+    const currentCareSettingCode: number = this.site.careSettingCode;
+
+    return this.siteResource.getSites(this.organization.id).pipe(
+      tap(sites => {
```
mtpultz: I don't think you will want to use `tap`.  Tap can be used for logging or performing side effects like you are currently, but it looks like you just want to get a value passed through the observable to the subscribe method.  If I wanted to do this I would use `map` in the pipe and within the map I would `site.reduce` to a number that you return from the map, and it will be passed to the subscribe.

prime-angular-frontend/src/app/modules/site-registration/pages/site-management/site-management.component.ts
```diff
@@ -119,6 +123,39 @@ export class SiteManagementComponent implements OnInit {
     };
   }
 
+  public isUnderReview(site: SiteListViewModel): boolean {
+    return (site.submittedDate && (site.status === 0 || site.status === SiteStatusType.UNDER_REVIEW));
```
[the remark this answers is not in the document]
mtpultz: Actually, `DRAFT` might be the best name for a `site.status === 0` that indicates it has been created, but not submitted

prime-angular-frontend/src/app/modules/site-registration/pages/site-management/site-management.component.ts
```diff
@@ -94,6 +97,7 @@ export class SiteManagementComponent implements OnInit {
   }
 
   public getOrganizationProperties(organization: OrganizationListViewModel): { key: string, value: string }[] {
+    // TODO: Why are these methods called so often despite no user interaction?
```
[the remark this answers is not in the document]
neophyte57: Thanks for the insight.

mtpultz (approved):
image

PR #1022, opened 2020-12-14
VC - comma separate care settings on claim

AnaisHebert:
- Multiple claims with the same name did not work with the verifier app, could only choose one care setting at a time to send. 
- Also removed time from renewal date claim


hogmann (commented):
Look Good To Me

hogmann (approved):

PR #1024, opened 2020-12-15
PRIME-934 local angular docker container

neophyte57:
I know this isn't high-priority but it was something I looked into last week out of curiosity.
 
Mainly was just wondering why `nginx.conf` is the way it is?  My change doesn't completely fix everything but instead of 
ERR_SOCKET_NOT_CONNECTED in Chrome at http://localhost:4200/, we at least get the pulsating PRIME logo and a 404 page at http://localhost:4200/info.  That is, now nginx is actually listening

If there is more information already known about this issue, could we add it to https://bcgovmoh.atlassian.net/browse/PRIME-934?


mtpultz:
@neophyte57 looks like there might be an issue with Keycloak with the error on http://localhost:4200 being `clientId is missing` so might be an issue during the build where it replaces the `environment.prod` file with `environment.prod.template` during the Angular build.  At least that's what I see locally:

image

neophyte57:
Yes @mtpultz  I was just puzzled with the state of `nginx.conf` and wondering if anyone had looked at this issue in any significant way.

mtpultz:
> Yes @mtpultz I was just puzzled with the state of `nginx.conf` and wondering if anyone had looked at this issue in any significant way.

Yah @neophyte57 it been in our backlog, but not addressed so that's great that you moved it a little closer to working locally!!!

mtpultz (approved):
We're so close to having this working again locally!!! image

PR #1025, opened 2020-12-15
PRIME-1278 Care setting column admin

hogmann:


prime-dotnet-webapi/ViewModels/Enrollee/EnrolleeListViewModel.cs
```diff
@@ -60,5 +60,7 @@ public string CurrentTOAStatus
         public bool AlwaysManual { get; set; }
 
         public bool RemoteAccess { get; set; }
+
+        public ICollection<EnrolleeCareSetting> EnrolleeCareSettings { get; set; }
```
james-hollinger, line 64: This should likely be a flattened collection of CareSettings or care setting IDs rather than EnrolleeCareSettings

AnaisHebert (approved):
🦕🦕🦕🦕

prime-angular-frontend/src/app/shared/pipes/care-setting-abbr.pipe.ts
```diff
@@ -0,0 +1,32 @@
...
+        return 'CP';
+      case CareSettingEnum.DEVICE_PROVIDER:
+        return 'DP';
+      case CareSettingEnum.HEALTH_AUTHORITY:
+        return 'HA';
+      case CareSettingEnum.PRIVATE_COMMUNITY_HEALTH_PRACTICE:
+        return 'PCHP';
+      default:
+        return '-';
+    }
```
johnnyl6, line 29: From the OOP perspective, seems these abbr strings can be part of the CareSettingEnum rather than defined in a helper transform class. Any benefit of using the PipeTransform?

james-hollinger (approved):
LGTM

prime-angular-frontend/src/app/shared/pipes/care-setting-abbr.pipe.ts
hogmann: In the enum we already map the code to the name. As far as I'm aware in TS we can't map a third property. The other thing is we can't define a helper method to map these abbr, without merging the enum with a namespace. I went with this route as i didn't really want to clutter the enum, but im open to some suggestions as maybe containing the infor to that file might be more beneficial.

johnnyl6: > In the enum we already map the code to the name. As far as I'm aware in TS we can't map a third property. The other thing is we can't define a helper method to map these abbr, without merging the enum with a namespace. I went with this route as i didn't really want to clutter the enum, but im open to some suggestions as maybe containing the infor to that file might be more beneficial.

Fair enough. LGTM. 

Another possible way is to have static instances of a regular class CareSetting, so that the third property can be added.  


```
class CareSetting {
    id: number;
    name: string;
    abbr: string;
    
    public static COMMUNITY_PHARMACIST: CareSetting = new CareSetting(0, 'COMMUNITY_PHARMACIST', 'CP');
    .......
}
```

hogmann: Just pushed some changes that move the abbr logic into the enum. With not a crazy amount of magic, seems like the way to merge a enum and namespace

johnnyl6: > Just pushed some changes that move the abbr logic into the enum. With not a crazy amount of magic, seems like the way to merge a enum and namespace

Agreed that they belong to be in the same file, makes more sense and easier reading as well..

johnnyl6 (approved):
✅

PR #1028, opened 2020-12-17
PRIME-1231 Overview for PRIME adjudication 

johnnyl6:


prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html
```diff
@@ -0,0 +1,62 @@
+<app-page [busy]="busy"
+          mode="full">
+
+  <app-page-header>Overview</app-page-header>
+
+  <app-enrollee-table *ngIf="busy.closed || dataSource.data.length"
```
AnaisHebert, line 6: Here you should use the component `app-adjudication-container` instead, it contains the enrollee table and the adjudicator actions. You can look at the `enrollee-events.component.html` to see an example.

AnaisHebert (changes requested):

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html
johnnyl6: > Here you should use the component `app-adjudication-container` instead, it contains the enrollee table and the adjudicator actions. You can look at the `enrollee-events.component.html` to see an example.

I thought about using app-adjudication-container template but had difficulties to adapt to the double scrollable pane design due to the position of the adjudicator action card. The overview component actually does extends from the adjudicator container, just with a different template.

AnaisHebert: oh okay, fair enough!

AnaisHebert (approved):
🐧🐧🐧

PR #1030, opened 2020-12-22
PRIME-1309 - PHSA eForms Immunizers

james-hollinger:
The PHSA eForms enrollment is being refactored to remove the "secret code" page and to include immunizers. The registration is now based on matching the person's name and email address to a known list that will be added to the database directly. Immunizers have their own role in keycloak called phsa_eforms_immunizer_covid19.


prime-dotnet-webapi/Services/PartyService.cs
```diff
@@ -99,6 +100,18 @@ public async Task DeletePartyAsync(int partyId)
...
 
+        public async Task<IEnumerable<PartyType>> GetPreApprovedRegistrationsAsync(string firstName, string lastName, string email)
+        {
+            return await _context.PreApprovedRegistrations
+                .AsNoTracking()
+                .Where(pre => pre.FirstName == firstName
+                    && pre.LastName == lastName
+                    && pre.Email == email)
+                .Select(pre => pre.PartyType)
+                .Distinct()
```
itchison, line 111: 👍

itchison (approved):
lgtm

prime-dotnet-webapi/HttpClients/Keycloak Administration/KeycloakAdministrationClient.cs
```diff
@@ -100,7 +103,18 @@ public async Task<bool> UpdateUserInfo(Guid userId, string email, string phoneNu
...
+            {
+                success &= await AssignRealmRole(userId, Roles.PhsaLabtech);
+            }
+
+            if (party.PartyTypes.Contains(PartyType.Immunizer))
+            {
+                success &= await AssignRealmRole(userId, Roles.PhsaImmunizer);
+            }
+
+            return success;
```
johnnyl6, line 117: It seems public method UpdatePhsaUserInfo will only assign labtech and immunizer roles which is very business specific. Maybe a method name with specific operation suffix is better since the generic name might cause confusion that it will update using the all info of input param "party". This is totally optional since this method is only used by phsa module.

johnnyl6 (approved):
🎉

prime-dotnet-webapi/HttpClients/Keycloak Administration/KeycloakAdministrationClient.cs
itchison: If this was UpdateUserInfo I'd agree. It's very specific to the eforms registration and may be refactored in the future if we do other provisioning for other systems. imo no changes required.

PR #1035, opened 2020-12-24
Update Keycloak Documentation

james-hollinger:
Updates to test and prod Keycloak to allow redirects for the new PHSA frontend.
Added a default redirect to BCSC in prod


mtpultz (approved):
image

PR #1038, opened 2021-01-04
PRIME-1330 Fix for PHSA redirects

mtpultz:
* Added new roles
* Updated to redirect appropriately when PHSA
* Updated route to distinguish the PHSA module from the Auth module to prevent infinite route loop when attempting to access the root route ('') of the PHSA Labtech module

### DIFF
--- a/prime-angular-frontend/src/app/modules/auth/shared/enum/role.enum.ts
+++ b/prime-angular-frontend/src/app/modules/auth/shared/enum/role.enum.ts
@@ -4,7 +4,10 @@ export enum Role {
   SUPER_ADMIN = 'prime_super_admin',
   READONLY_ADMIN = 'prime_readonly_admin',
 
+  PHSA_LABTECH = 'phsa_eforms_labtech',
+  PHSA_IMMUNIZER = 'phsa_eforms_immunizer_covid19',
+
   // Feature Flags
   FEATURE_VC_ISSUANCE = 'feature_vc_issuance',
-  FEATURE_SITE_PHARMACIST = 'feature_site_pharmacist'
+  FEATURE_SITE_PHARMACIST = 'feature_site_pharmacist',
 }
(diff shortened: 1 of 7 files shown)


prime-angular-frontend/src/app/modules/auth/shared/enum/role.enum.ts
```diff
@@ -4,7 +4,10 @@ export enum Role {
   SUPER_ADMIN = 'prime_super_admin',
   READONLY_ADMIN = 'prime_readonly_admin',
 
+  PHSA_LABTECH = 'phsa_eforms_labtech',
+  PHSA_IMMUNIZER = 'phsa_eforms_immunizer_covid19',
```
mtpultz, line 8: Distinguishing role indicating they are not just a `prime_user`, but specifically part of PHSA

prime-angular-frontend/src/app/modules/phsa-labtech/phsa-labtech.routes.ts
```diff
@@ -1,5 +1,5 @@
 export class PhsaLabtechRoutes {
-  public static PHSA_LABTECH = 'phsa';
+  public static PHSA_LABTECH = 'phsa-labtech';
```
mtpultz, line 2: Infinite loop created when routing to the default route of PHSA and not being able to distinguish between it and the AuthModule route which were both `/phsa`

hogmann (approved):
LGTM, Just once built i want to do a quick run through on openshifty

prime-angular-frontend/src/app/modules/phsa-labtech/phsa-labtech.routes.ts
neophyte57: Is this a problem or not?  User bookmarks `http://.../phsa-labtech/demographic`, logs out, and later returns to bookmark.  Then they are directed to `http://.../info` ... probably not what they expected?  Simply User Error?

mtpultz: Suspect that it is a user that submits, remains logged in, receives an email with a link to login, but since they were already logged in they hit the redirect guard and go to sites.  The infinite loop was behind that issue where Angular was having issues due to the `http://localhost:4200/phsa` route being the same for PHSA login and PhsaLabtechModule.

PR #1042, opened 2021-01-06
Refactor Base Clients

james-hollinger:
Using the BaseClient for the other clients plus some other minor refactoring.


mtpultz (approved):
image

PR #1044, opened 2021-01-08
Updates to PHSA

mtpultz:
* Changed validation message to AlertComponent to match radio buttons since mat-error doesn't work with checkboxes and radio buttons
* Updated PHSA URL to `/phsa-eforms`
* Clean up bindings in templates
* Use existing form validator
* Refactor to pipe replacing namespace
* Refactor PHSA to use FormState builders to decrease complexity of form state services


prime-angular-frontend/src/app/modules/phsa-labtech/phsa-labtech.routes.ts
```diff
@@ -1,5 +1,5 @@
 export class PhsaLabtechRoutes {
-  public static PHSA_LABTECH = 'phsa-labtech';
+  public static PHSA_LABTECH = 'phsa-eforms';
```
AnaisHebert, line 2: should we change the variable name to `PHSA_EFORMS`?

prime-angular-frontend/src/app/modules/phsa-labtech/shared/services/phsa-labtech-form-state.service.ts
```diff
@@ -28,21 +29,23 @@ export class PhsaFormStateService extends AbstractFormStateService<PhsaEnrollee>
   }
 
   public get json(): PhsaEnrollee {
-    const { phone, phoneExtension, email } = this.demographicsForm.getRawValue();
-    // TODO: Needs to get value (code) of selected checkboxes
-    const { partyTypes } = this.availableAccessForm.getRawValue();
+    // const accessCode = this.accessCodeFormState.json;
```
AnaisHebert, line 32: should this be here?

mtpultz: I don't know.  Was removed when I was on vacation

prime-angular-frontend/src/app/modules/phsa-labtech/phsa-labtech.routes.ts
mtpultz: Sure, but then I'd want to rename the entire module.  So we could do that.

AnaisHebert (approved):
LGTM! 🎡

PR #1045, opened 2021-01-08
Updated orgbook displayed and business licence validations

mtpultz:
* Update org name displayed orgbook that was displaying the form group clear (X) icon
* Update bus licence validation to not show both validation messages when only a single error exists


AnaisHebert (approved):
🍕🍕

johnnyl6 (approved):
🏂

PR #1046, opened 2021-01-08
PRIME-1267 Small changes on admin

AnaisHebert:


mtpultz (approved):
Screen Shot 2021-01-08 at 11 45 24 AM

PR #1048, opened 2021-01-08
Resolve Remaining Analyzer Fixes

james-hollinger:
Getting the code suggestions down to zero. No functionality should be impacted by these changes.


prime-dotnet-webapi/Models/Feedback.cs
```diff
@@ -13,7 +13,8 @@ public class Feedback : BaseAuditable
         public int EnrolleeId { get; set; }
 
         [Required]
-        public string route { get; set; }
+        [Column("route")]
+        public string Route { get; set; }
```
james-hollinger, line 17: Decided to just keep the DB column lower case to avoid making another migration. We can update this when/if Feedback is looked at again.

AnaisHebert (approved):
💥💥

mtpultz (approved):
🧐

PR #1049, opened 2021-01-08
PRIME-1335 Wording/spelling/punctuation fixes - user enrolment

mtpultz:


AnaisHebert (approved):
🚳💯

prime-angular-frontend/src/app/modules/enrolment/pages/bcsc-demographic/bcsc-demographic.component.html
```diff
@@ -15,8 +15,8 @@
           This information comes from your BC Services Card. If you use a different name professionally, please
           select "I have a preferred name."
           <small class="font-italic d-block">
-            Entering a preferred name in PRIME will not change your BC Services Card name.<br>
-            NOTE: The name you provide must match your College registration information, if applicable.
+            Entering a preferred name in PRIME won't change your services card.<br>
```
itchison, line 18: I'm not positive this is correct. It's weird that we don't call out the BC Services Card in proper caps and format but do everywhere else.

PR #1053, opened 2021-01-11
Added unsubscribe method to new base component class

mtpultz:
Added a new base component class that can be used with `takeUntil` with subscriptions that need to be unsubscribed when components are destroyed to ensure we don't get memory leaks.


AnaisHebert (approved):
🥞🥞🥞

PR #1058, opened 2021-01-12
PRIME-1230 Triage initial commit

hogmann:


prime-angular-frontend/src/app/lib/modules/ngx-contextual-help/contextual-help/contextual-help.component.html
```diff
@@ -1,18 +1,24 @@
 <a mat-icon-button
-   color="primary"
+   [color]="color"
    aria-label="Help"
+   class="{{ small ? 'small-container' : ''}}"
    [matMenuTriggerFor]="help"
    (click)="$event.stopPropagation()">
-  <mat-icon>info</mat-icon>
+  <mat-icon class="{{ small ? 'small-icon' : ''}}">{{icon}}</mat-icon>
```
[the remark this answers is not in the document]
hogmann: I was trying to remember how to do this, but couldn't remember what to search

mtpultz (commented):
Feels like this might be a good representation of the menu:

image

* Feels like we should extend the menu width out to about `380px`.  We don't seem to be considering responsiveness on the admin and that width even works on a Pixel 2.  It should help the menu feel a bit less squished.
* Feels like the icon in the menu can be bumped up in size maybe even (or close to) dialog size in the image I set it to 2.2rem.  I would also move the `mat-icon` outside the `h1` and instead use flexbox so you can vertically align the icon and title.  It will also allow you to pull the underline back so it only appears under the `h1` like in the dialog.
* Feels like the title should be bolded, and let `<h2>` have a normal font-weight
1. `<h2>` with `mb-3`
2. `<div>` content container `mb-4 pl-3` 
3. `<div>` inner content `mb-2`
4. or replace 2 and 3 with a unstyled `<ul>` and decrease the indent since this is technically a list of reasons 
5. `<button>` with `mb-2`

prime-angular-frontend/src/app/lib/modules/ngx-contextual-help/contextual-help/contextual-help.component.html
```diff
@@ -1,18 +1,24 @@
 <a mat-icon-button
-   color="primary"
+   [color]="color"
```
mtpultz, line 2: Feels like `color` should be renamed since using `color` indicates that the entire menu might be that colour, but it only affects the menu icon.

prime-angular-frontend/src/app/lib/modules/ngx-contextual-help/contextual-help/contextual-help.component.html
```diff
@@ -1,20 +1,27 @@
...
           [xPosition]="xPosition"
           [yPosition]="yPosition">
   <ng-template matMenuContent>
     <!-- Content project contextual help components when available -->
-    <h1 *ngIf="contextualHelpTitleChildren.length"
-        class="title lead">
-      <ng-content select="[appContextualTitle]"></ng-content>
-    </h1>
+    <div *ngIf="contextualHelpTitleChildren.length"
+         class="d-flex align-items-stretch mb-2">
```
mtpultz, line 16: Looking at the UI feels like this should be `mb-3` to create a bit more space between the title and content.

prime-angular-frontend/src/app/shared/components/dialogs/content/triage/triage.component.html
```diff
@@ -0,0 +1,59 @@
...
+
+  <ng-container appContextualTitle>
+    Triage
+  </ng-container>
+
+  <ng-container appContextualContent>
+    <h2 class="h5 mb-3">
+      Reason for manual adjudication:
+    </h2>
+    <div class="pl-2 mb-4">
```
mtpultz, line 15: Feels like this should be something like:

```
<ul class="list-unstyled mb-4 ml-3">
  <li *ngFor="let reason of (status$ | async)?.enrolmentStatusReasons; let last = last" 
      class="blue"
      [class.pb-2]="!last">
    {{ reason.statusReason.name }}
  </li>
</ul>
```

mtpultz (approved):
image

AnaisHebert (approved):
🥞lgtm

PR #1061, opened 2021-01-13
PRIME-1343 Preferred Name is only appearing on the Review Page after you hover over the first pencil icon 

mtpultz:


hogmann (approved):
LGTM

PR #1062, opened 2021-01-13
PRIME-1047 Admin Visual Indicator of Community Pharmacies without a Business Licence

johnnyl6:


prime-angular-frontend/src/app/modules/adjudication/shared/components/site-registration-table/site-registration-table.component.ts
```diff
@@ -82,5 +83,12 @@ export class SiteRegistrationTableComponent implements OnInit {
       : 'N/A';
   }
 
+  public missingBusinessLicence(siteRegistration: SiteRegistrationListViewModel) {
```
mtpultz, line 86: You might be able to use the `YesNoPipe` with a `DefaultPipe` for `N/A`.  Been a bit since I looked at that pipe, but feels like something like this might work so we can try and keep our controller methods down:

```
{ yourValue | yesNo: yourArgument | default: 'N/A' }
```

johnnyl6: Love the pipes😎

mtpultz (approved):
image

PR #1064, opened 2021-01-15
PRIME-1211 - College Licence Prefix

james-hollinger:
College Prefix is now based on Licence rather than College.


AnaisHebert (approved):
LGTM 🎈🎈🎈

PR #1074, opened 2021-01-20
PRIME 1289 - CRON job for Renewal Emails

neophyte57:
This cronjob expects certain secrets to be available.  They are currently only in the `dev` OpenShift environment

prime-service-account > prime_service_account_client_id
prime-service-account > prime_service_account_client_secret

image


james-hollinger (approved):
LGTM

PR #1079, opened 2021-01-21
PRIME-1354 PHSA eForms Email Language update 

johnnyl6:
Refactored contact information component so that
- shows different language for email section for PHSA eForm
- hide/show the sms phone number field


prime-angular-frontend/src/app/shared/components/contact-information-form/contact-information-form.component.ts
```diff
@@ -10,11 +10,11 @@ import { FormUtilsService } from '@core/services/form-utils.service';
 })
 export class ContactInformationFormComponent implements OnInit {
   @Input() public form: FormGroup;
-  @Input() public showSmsPhone: boolean;
+  @Input() public forPhsa: boolean;
```
neophyte57, line 13: Works fine but I just wonder for a shared component, should we have a parameter named `forPhsa` that is very PHSA-specific?  Would it be overkill to provide a parameter that allows the alternative text, that is, "Please provide the email address you have provided to your Health Authority to be associated with your eForms enrolment."?  Thoughts from others?

mtpultz: Yah @neophyte57  I definitely agree.  We'd want to keep shared components generic, and if there is specific text that is required to be inserted based on the current module use projection.

prime-angular-frontend/src/app/shared/components/contact-information-form/contact-information-form.component.html
```diff
@@ -44,12 +44,15 @@
     <app-page-subheader>
       <ng-container appPageSubheaderTitle>Notification Information</ng-container>
       <ng-container appPageSubheaderSummary>
-        Provide an email address 
-        <ng-container *ngIf="showSmsPhone">
-          and phone number (optional) 
-        </ng-container>
-        that may be used for sending notifications from the PRIME
-        system.
+        <div *ngIf="!forPhsa else PhsaSummary">
```
mtpultz, line 47: I'm surprised this works unless there is an update to `*ngIf` structural directive that I'm not aware of but typically there's a semicolon between the statements `*ngIf="!forPhsa; else PhsaSummary"`

prime-angular-frontend/src/app/shared/components/contact-information-form/contact-information-form.component.html
```diff
@@ -44,12 +44,15 @@
...
+          Provide an email address and phone number (optional)
```
mtpultz, line 48: Specific text can be projected into the component using `<ng-content></ng-content>` to keep this generic

prime-angular-frontend/src/app/shared/components/contact-information-form/contact-information-form.component.html
```diff
@@ -69,16 +72,16 @@
 
           </div>
 
-          <ng-container *ngIf="showSmsPhone">
+          <ng-container *ngIf="!forPhsa">
```
mtpultz, line 75: If this is only used for PHSA you'll want to project the markup into the component and use a selector - `<ng-content select="..."></ng-content>`

prime-angular-frontend/src/app/modules/phsa-eforms/pages/bcsc-demographic/bcsc-demographic.component.html
```diff
@@ -32,7 +32,12 @@
     </section>
 
     <app-contact-information-form [form]="form"
-                                  [showSmsPhone]="false"></app-contact-information-form>
+                                  [showSmsPhone]="false">
+      <div notificationInfoSummary>
```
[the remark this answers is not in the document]
johnnyl6: > When you use something like this it makes it look like an Angular directive when it isn't. You can create a directive like we do for components like `PageSubheader2Component` where you might add `ContactInfoSummaryDirective` and be able to drop the `<div>` since it isn't a structural or styled element:
> 
> ```js
> <ng-container appContactInfoSummary>Lorem ipsum dolor sit...</ng-container>
> ```
> 
> In the `ContactInformationForm` you can do a similar `select="[appContactInfoSummary]"`, but the directive allows you in the controller to use `@ContentChildren(ContactInfoSummaryDirective) public summary QueryList<ContactInfoSummaryDirective>;` to check the length of the projected content and reduce a bit more of the markup using `NgIfElse` structural directive on a getter like `<ng-container *ngIf="hasCustomSummary; else default">...</ng-container>` and have the default in a `<ng-template #default>...</ng-template>`.

Yep PageSubheader2Component is a very good example.👍

mtpultz (approved):
image

PR #1083, opened 2021-01-26
PRIME 1268 Add assignee to claimed (aborted)

neophyte57:
This was a start on the ticket based on pre-existing functionality.  Feel free to ignore or pick and choose any code you like @hogmann

### DIFF
--- a/prime-angular-frontend/src/app/modules/adjudication/shared/components/adjudication-container/adjudication-container.component.ts
+++ b/prime-angular-frontend/src/app/modules/adjudication/shared/components/adjudication-container/adjudication-container.component.ts
@@ -109,6 +109,7 @@ export class AdjudicationContainerComponent implements OnInit {
 
     this.busy = this.dialog.open(ClaimEnrolleeComponent, { data }).afterClosed().subscribe();
 
+    // TODO: Execute only if user performed Claim/Assign action
     this.adjudicationResource
       .setEnrolleeAdjudicator(enrolleeId)
       .subscribe((updatedEnrollee: HttpEnrollee) => this.updateEnrollee(updatedEnrollee));
--- a/prime-angular-frontend/src/app/modules/adjudication/shared/components/adjudication-container/adjudication-container.component.ts
+++ b/prime-angular-frontend/src/app/modules/adjudication/shared/components/adjudication-container/adjudication-container.component.ts
@@ -103,6 +103,12 @@ export class AdjudicationContainerComponent implements OnInit {
   }
 
   public onClaim(enrolleeId: number) {
+    const data: DialogOptions = {
+      data: { possibleAction: ClaimActionEnum.Claim },
+    };
+
+    this.busy = this.dialog.open(ClaimEnrolleeComponent, { data }).afterClosed().subscribe();
+
     this.adjudicationResource
       .setEnrolleeAdjudicator(enrolleeId)
       .subscribe((updatedEnrollee: HttpEnrollee) => this.updateEnrollee(updatedEnrollee));
@@ -110,7 +116,7 @@ export class AdjudicationContainerComponent implements OnInit {
 
   public onDisclaim(enrolleeId: number) {
     const data: DialogOptions = {
-      title: 'Disclaim Enrolment',
+      data: { possibleAction: ClaimActionEnum.Disclaim },
       component: ManualFlagNoteComponent
     };
 
--- a/prime-angular-frontend/src/app/shared/components/dialogs/content/claim-enrollee/claim-enrollee.component.html
+++ b/prime-angular-frontend/src/app/shared/components/dialogs/content/claim-enrollee/claim-enrollee.component.html
@@ -1,17 +1,25 @@
 <h2 mat-dialog-title
     class="d-flex align-items-stretch">
   <span class="flex-grow-1">
-    Disclaim Enrollee
+    {{ this.getTitle() }}
   </span>
 </h2>
 <mat-dialog-content>
-  <span>You can unclaim the enrollee or assign them to a colleague.</span>
+  <span>{{ this.getPromptText() }}</span>
 
   <div class="mt-4">
-    <button mat-flat-button
-            color="primary"
-            (click)="onDisclaim()"
-            class="mr-4">Disclaim Enrollee</button>
+    <span *ngIf="displayAsClaimPrompt()">
+      <button mat-flat-button
+              color="primary"
+              (click)="onClaim()"
+              class="mr-4">Claim</button>
+    </span>
+    <span *ngIf="!displayAsClaimPrompt()">
+      <button mat-flat-button
+              color="primary"
+              (click)="onDisclaim()"
+              class="mr-4">Disclaim Enrollee</button>
+    </span>
     <button mat-flat-button
             color="primary"
             [matMenuTriggerFor]="menu"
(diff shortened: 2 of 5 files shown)


prime-angular-frontend/src/app/shared/components/dialogs/content/claim-enrollee/claim-enrollee.component.ts
```diff
@@ -24,21 +25,41 @@ export class ClaimEnrolleeAction {
...
+  public displayAsClaimPrompt(): boolean {
+    return this.possibleAction == ClaimActionEnum.Claim;
+  }
+
+  public getTitle(): string {
+    return (this.possibleAction == ClaimActionEnum.Claim ? "Claim" : "Disclaim") + " Enrollee";
+  }
+
+  public getPromptText(): string {
+    return (this.possibleAction == ClaimActionEnum.Claim ?
```
mtpultz, line 51: You can use this ternary directly in the template to keep the number of methods in the controller down, and there's also a bit of performance not having methods invoked on each digest cycle.

prime-angular-frontend/src/app/shared/components/dialogs/content/claim-enrollee/claim-enrollee.component.ts
```diff
@@ -24,21 +25,41 @@ export class ClaimEnrolleeAction {
...
 
+  /**
+   * @returns Whether to display as a "claim" form (`true`) or as a "disclaim" form (`false`)
+   */
+  public displayAsClaimPrompt(): boolean {
+    return this.possibleAction == ClaimActionEnum.Claim;
+  }
+
+  public getTitle(): string {
+    return (this.possibleAction == ClaimActionEnum.Claim ? "Claim" : "Disclaim") + " Enrollee";
```
mtpultz, line 47: You can use this ternary directly in the template `{{ ternary }} Enrollee` you just need to add `public ClaimAction = ClaimAction;` at the top of the component to allow enums in the template.

prime-angular-frontend/src/app/shared/components/dialogs/content/claim-enrollee/claim-enrollee.component.ts
```diff
@@ -24,21 +25,41 @@ export class ClaimEnrolleeAction {
...
   ) {
+    this.possibleAction = data.data?.possibleAction;
     this.adjudicators$ = new BehaviorSubject<Admin[]>([]);
   }
 
+  /**
+   * @returns Whether to display as a "claim" form (`true`) or as a "disclaim" form (`false`)
+   */
+  public displayAsClaimPrompt(): boolean {
+    return this.possibleAction == ClaimActionEnum.Claim;
```
mtpultz, line 43: You can do this directly in the template, and you always want to use `===` in JS to avoid coercion when using `==`.

prime-angular-frontend/src/app/shared/components/dialogs/content/claim-enrollee/claim-enrollee.component.ts
mtpultz, line 47: You'll always want to use `===` in JS to avoid coercion when using `==`.

prime-angular-frontend/src/app/shared/components/dialogs/content/claim-enrollee/claim-enrollee.component.ts
mtpultz, line 51: You'll always want to use `===` in JS to avoid coercion when using `==`.

prime-angular-frontend/src/app/shared/components/dialogs/content/claim-enrollee/claim-enrollee.component.ts
```diff
@@ -24,21 +25,41 @@ export class ClaimEnrolleeAction {
...
+  }
+
   public onDisclaim(): void {
     const output = new ClaimEnrolleeAction();
     output.action = ClaimActionEnum.Disclaim;
     this.dialogRef.close({ output });
   }
 
-  public onClaim(adminId: number): void {
+  public onClaim(adminId?: number): void {
```
mtpultz, line 62: Would this not be required?  Feels odd that this would be optional.

prime-angular-frontend/src/app/shared/components/dialogs/content/claim-enrollee/claim-enrollee.component.html
```diff
@@ -1,17 +1,25 @@
...
-            color="primary"
-            (click)="onDisclaim()"
-            class="mr-4">Disclaim Enrollee</button>
+    <span *ngIf="displayAsClaimPrompt()">
+      <button mat-flat-button
+              color="primary"
+              (click)="onClaim()"
+              class="mr-4">Claim</button>
+    </span>
+    <span *ngIf="!displayAsClaimPrompt()">
```
mtpultz, line 17: You'll want to use a `templateRef` that's referenced in the above `*ngIf` using an `else`-statement, and change it to use a `ng-container` instead of a `<span>`.  If we can avoid adding markup to the template we should to decrease complexity.  `ng-container` is a placement marker and when rendered it will render it's content or the else content underneath the tag so it doesn't disrupt the hierarchy of the styles.

```
<ng-container *ngIf="possibleAction === ClaimActionEnum.Claim; else disclaim">

// Removed for brevity

<button #disclaim 
        mat-flat-button ...>
```

prime-angular-frontend/src/app/shared/components/dialogs/content/claim-enrollee/claim-enrollee.component.ts
```diff
@@ -24,21 +25,41 @@ export class ClaimEnrolleeAction {
 })
 export class ClaimEnrolleeComponent implements OnInit {
   public adjudicators$: BehaviorSubject<Admin[]>;
+  private possibleAction: ClaimActionEnum;
```
mtpultz, line 28: Feels like this should be `action` since once we have this dialog open it isn't a possibleAction it's the actual action.

prime-angular-frontend/src/app/modules/adjudication/shared/components/adjudication-container/adjudication-container.component.ts
```diff
@@ -103,14 +103,21 @@ export class AdjudicationContainerComponent implements OnInit {
...
+    // TODO: Execute only if user performed Claim/Assign action
     this.adjudicationResource
       .setEnrolleeAdjudicator(enrolleeId)
       .subscribe((updatedEnrollee: HttpEnrollee) => this.updateEnrollee(updatedEnrollee));
   }
 
   public onDisclaim(enrolleeId: number) {
     const data: DialogOptions = {
-      title: 'Disclaim Enrolment',
+      data: { possibleAction: ClaimActionEnum.Disclaim },
```
mtpultz, line 120: Feels like this should be `action` since this is the action that will be represented in the dialog

prime-angular-frontend/src/app/modules/adjudication/shared/components/adjudication-container/adjudication-container.component.ts
```diff
@@ -103,14 +103,21 @@ export class AdjudicationContainerComponent implements OnInit {
   }
 
   public onClaim(enrolleeId: number) {
+    const data: DialogOptions = {
+      data: { possibleAction: ClaimActionEnum.Claim },
```
mtpultz, line 107: Feels like this should be `action` since this is the action that will be represented in the dialog

PR #1085, opened 2021-01-27
PRIME-1349 Viewing preferred name in admin

neophyte57:


prime-angular-frontend/src/app/shared/components/enrollee/enrollee-review/enrollee-review.component.html
```diff
@@ -1,6 +1,6 @@
 <app-enrollee-profile [enrollee]="enrollee"></app-enrollee-profile>
 
-<section *ngIf="hasPreferredName && identityProvider === IdentityProviderEnum.BCSC"
+<section *ngIf="hasPreferredName"
```
mtpultz, line 3: This change would display the information to a BCeID enrollee.  Feels like the update might be the addition of OR  to this statement of

```
... && (identityProvider === IdentityProvider.BCSC || identityProvider === IdentityProvider.IDIR)
```

neophyte57: I can make that change @mtpultz but why would displaying their preferred name back to a BCeID enrollee be a problem?

neophyte57: OK, I am looking at `bceid-demographic.component.ts` and its template but don't think I understand the form without actually running it and I don't have a BCeID to test with.  

So are you saying that `hasPreferredName` will evaluate to `true` for BCeID enrollees and that their `EnrolleeProfile` instance `preferredFirstName` will always be identical to their `firstName` and `preferredLastName` will also equal `lastName`?  If so, yes doesn't make sense to show preferred name fields for BCeID enrollees.   

@mtpultz we can chat on phone if that's easier  😊

mtpultz (approved):
👍

PR #1077, opened 2021-01-21
PRIME-1338 - Email Refactor

james-hollinger:
Made a 'canonical' model for Emails that are consumed by the CHES and SMTP clients. Lots of simplification.

PR #1095, opened 2021-01-29
Fix for defect related to renewal emails

neophyte57:
Easiest way to test is to use temporary `EmailsControllers` changes found in PR #1077 comments to invoke renewal emails algorithm and then manipulate `ExpiryDate` value in `Agreement` table so that emails are triggered


mtpultz (approved):
he 🏒  he 🥅

james-hollinger (approved):
🦶🐛

PR #1105, opened 2021-02-03
PRIME-1382 Return all-no self-declaration questions to enrolment cards

mtpultz:


itchison (approved):
LGTM - Good idea on the boolean.

PR #1106, opened 2021-02-03
PRIME-630 - Added Tests to Pipes

mtpultz:
`ng test --include='./app/**/pipes/*.spec.ts'`


neophyte57:
Is there a way to just run the `pipe` tests?  As I am getting these compile errors:

image

mtpultz:
> Is there a way to just run the `pipe` tests? As I am getting these compile errors:
> 
> image

@neophyte57 no as far as I know you can't run the tests singularly like in backend frameworks like XUnit, PHPUnit, etc.  We won't be able to run the entire suite until Anais' branch is in.

AnaisHebert (approved):
❄❄❄❄

johnnyl6 (approved):

PR #1113, opened 2021-02-06
Fix OboSite Validations

mtpultz:


AnaisHebert (approved):
Submit on overview page error is fixed 🎈🎈🎈

PR #1115, opened 2021-02-08
PRIME-1374 Select health authorities

neophyte57:
Other than checking that it works and whether following proper conventions, please review the added code comments and verify they make sense and are accurate.


prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
```diff
@@ -79,6 +95,23 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
       }
     });
 
+    const enrolleeId = this.enrolment.id;
+
+    this.enrolleeHealthAuthorities.clear();
+    // Any checked HA is converted into an enrollee health authority FormGroup,
+    // which is used to create the payload to back-end
+    this.availableHAs?.controls.forEach((checkbox, i) => {
```
mtpultz, line 103: Is it possible that `availableHealthAuthorities` would be `null` or `undefined` instead of an empty `FormArray`?

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
```diff
@@ -130,6 +163,13 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
     return this.careSettingTypes;
   }
 
+
+  public hasSelectedHACareSetting(): boolean {
+    // When code is in template, get: `Parser Error: Bindings cannot contain assignments`
```
mtpultz, line 168: Don't know if this comment provides something we want to document since this is just something that Angular does not allow.

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
mtpultz: Also, I'm not sure you need to do this on submission.  Transforming the data should occur in the `FormStateService` in the `json` method.  That's where you can get the `form.value` and loop over it and transform it for submission or any other use outside the form.  That way all form models and logic are only manipulated through the form state service.  An example, would be `convertSelfDeclarationsToJson` or `convertRemoteAccessSitesToJson` in the `EnrolmentFormStateService`.

prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts
```diff
@@ -515,6 +524,14 @@ export class EnrolmentFormStateService extends AbstractFormStateService<Enrolmen
     });
   }
 
+  public buildEnrolleeHealthAuthorityFormGroup(): FormGroup {
+    // Needs to have the fields to carry data to back-end
+    return this.fb.group({
+      enrolleeId: [null, []],
```
mtpultz, line 530: You won't need to include this in the form as it can be added as it's transformed in the `json` method of the `EnrolmentFormStateService`. An example, would be `convertSelfDeclarationsToJson` or `convertRemoteAccessSitesToJson` in the `EnrolmentFormStateService`.

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
neophyte57: I don't understand this error message as I'm not doing any assignment?  This doesn't fully explain it:  https://stackoverflow.com/questions/43117917/angular-2-bindings-cannot-contain-assignments

prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts
neophyte57: It seems I still need `buildEnrolleeHealthAuthorityFormGroup` in `patchForm`

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
```diff
@@ -169,6 +200,17 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
     if (!this.careSettings?.length) {
       this.addCareSetting();
     }
+
+    // Initialize Health Authority form even if it might not be used:
+    // Create unchecked checkboxes for each known Health Authority.
+    this.form.controls.selectableHealthAuthorities = this.fb.array(this.configService.healthAuthorities.map(() => this.fb.control(false)));
```
mtpultz, line 206: We shouldn't be adding controls directly to the controls array of a form.  There is an API for adding controls if needed, but it feels like this might be possible with just the single form group - `enrolleeHealthAuthorities`.  Currently, you're dynamically adding a new `FormArray` and populating it with controls, but could it be accomplished using `enrolleeHealthAuthorities` to avoid transferring between the `FormArray`s in the JSON method?

neophyte57: @mtpultz About this feedback and feedback re: comments, let's talk later.  Thanks.

prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts
```diff
@@ -398,6 +409,29 @@ export class EnrolmentFormStateService extends AbstractFormStateService<Enrolmen
     });
   }
 
+  private convertCareSettingFormToJson(enrolleeId: number): any {
+    // Variable names must match keys for FormArrays in the form to get values
+    const { careSettings, selectableHealthAuthorities } = this.careSettingsForm.getRawValue();
+
+    // Any checked HA is converted into an enrollee health authority,
+    // which is used to create the payload to back-end
+    let enrolleeHealthAuthorities = selectableHealthAuthorities.map((checkState, i) => {
```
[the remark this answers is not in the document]
mtpultz: @neophyte57  alternatively if you want to avoid the filter when you see a map followed by a filter you can almost always replace it with a reduce:

```js
const {  careSettings, enrolleeHealthAuthorities: updatedSelections } = this.careSettingsForm.getRawValue();

const enrolleeHealthAuthorities = updatedSelections
  .reduce((selectedHealthAuthorities: ProperType, checked: boolean, i) => {
    if (checked) {
      selectedHealthAuthorities.push({ enrolleeId, healthAuthorityCode: this.configService.healthAuthorities[i] })
    }
    return selectedHealthAuthorities;
  }, []);

return {  careSettings,  enrolleeHealthAuthorities };
```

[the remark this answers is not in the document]
mtpultz: @neophyte57 don't know if you saw the comment above, but instead of filtering out only a single negative result instead of all negative results we can avoid the filter and use a reduce.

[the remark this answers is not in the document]
neophyte57: @mtpultz Sorry missed the `reduce` earlier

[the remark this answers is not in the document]
mtpultz: Looks awesome @neophyte57

mtpultz (approved):
image

AnaisHebert (commented):
Should we remove the "Facility" text below the health authority on the overview page
image

AnaisHebert (commented):
also the pencil icon leads to the old health-authority page instead of the one on the care setting page

AnaisHebert (changes requested):
also the pencil icon leads to the old health-authority page instead of the one on the care setting page

james-hollinger:
Just for future reference, any change to the EnrolleeUpdateModel has the potential to break the AllowableChanges part of the minor update rule, as comparing collections its done on a property by property basis. 

Looks like the minor update rule wasn't updated when this feature was originally added to the app, so now that the feature is being re-enabled you should know that changing your health authority will not be detected as a major change.

neophyte57:
> and changing your care setting from health authority to something else should remove the health authorities that you have chosen

@AnaisHebert  Thanks for testing further.  I've implemented the changes in your comments.  Please re-review.

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
```diff
@@ -169,6 +186,8 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
     if (!this.careSettings.length) {
       this.addCareSetting();
     }
+
+    this.healthAuthorities = this.configService.healthAuthorities;
```
mtpultz, line 190: @neophyte57 your PR has a few errors.  This is one of the issues where the setting of configuration values for health authorities is delayed by not setting them immediately in the constructor.  When you come from the overview to the care setting view the `ngOnInit` lifecycle hook has already been invoked, and your template is being bound with public data, but it's missing the configuration when looping over the health authorities.  You'll want to use the constructor like we do for `careSetting` config and the other components to set up initial component values like configuration.

mtpultz (approved):
image

PR #1116, opened 2021-02-08
PRIME-1384 Practitioner Id improvements

AnaisHebert:


prime-angular-frontend/src/app/shared/components/forms/college-certification-form/college-certification-form.component.ts
```diff
@@ -33,13 +34,15 @@ export class CollegeCertificationFormComponent implements OnInit {
    * @description
    * Indicates the licenceCode is validated by PharmaNet.
    */
-  public licenceValidatedByPharmaNet: boolean;
   public practices: PracticeConfig[];
   public filteredLicenses: Config<number>[];
   public filteredPractices: Config<number>[];
   public hasPractices: boolean;
+  public prescriberIdMandatory: PrescriberIdType;
```
AnaisHebert, line 41: this one should be removed

prime-angular-frontend/src/app/shared/components/forms/college-certification-form/college-certification-form.component.ts
```diff
@@ -221,9 +238,9 @@ export class CollegeCertificationFormComponent implements OnInit {
     return this.practices.filter(p => p.collegePractices.map(cl => cl.collegeCode).includes(collegeCode));
   }
 
-  private checkLicenceCodeValidatedByPharmaNet(licenceCode: number): boolean {
+  private getPrescriberIdType(licenceCode: number): PrescriberIdType | undefined {
```
mtpultz, line 238: It's a bit weird to have a return type that can be explicitly `undefined`.  I would add an appropriate default of `null` using null coalesce - `...pop() ?? null`

prime-angular-frontend/src/app/shared/components/forms/college-certification-form/college-certification-form.component.html
```diff
@@ -55,32 +57,45 @@
...
                   <input matInput
-                         placeholder="Practitioner ID"
+                         placeholder="Prescriber ID"
                          formControlName="practitionerId">
                   <mat-error *ngIf="form.get('practitionerId').hasError('required')">Required</mat-error>
                   <mat-error *ngIf="form.get('practitionerId').hasError('numeric')">
-                    Must contain only numeric characters
+                    Must contain only 5 numeric characters
+                  </mat-error>
+                  <mat-error *ngIf="form.get('practitionerId').hasError('length')">
```
mtpultz, line 97: Has the same error message so we can just `&&` the check for length and numeric

prime-angular-frontend/src/app/shared/components/forms/college-certification-form/college-certification-form.component.html
```diff
@@ -55,32 +57,45 @@
...
-                       placeholder="Licence Number from College"
-                       formControlName="licenseNumber"
-                       (blur)="onLicenceNumberBlur()">
-                <mat-error *ngIf="form.get('licenseNumber').hasError('required')">Required</mat-error>
-                <mat-error *ngIf="form.get('licenseNumber').hasError('alphanumeric')">
-                  Must contain only alphanumeric characters
-                </mat-error>
-              </mat-form-field>
+        <div class="col-sm-12 pl-5"
+             *ngIf="!condensed && prescriberIdMandatory > 1">
```
johnnyl6: Feels like there is a better way than using magic number "1", either defining the enum as a public property and refer in template, or define a method to return result for the ngIf

mtpultz:
Need to consider the responsiveness of the form:

Screen Shot 2021-02-09 at 10 38 42 AM

Screen Shot 2021-02-09 at 10 39 18 AM

Screen Shot 2021-02-09 at 10 39 38 AM

mtpultz (commented):
Also, should we be showing the checkbox if we know that the user can't interact with it?  I would prefer not to have it displayed when prescriber ID is mandatory unless that's a hard and fast business decision.  It's just confusing to have a field in the UI that you can't interact with as a user.

prime-angular-frontend/src/app/shared/components/forms/college-certification-form/college-certification-form.component.html
mtpultz: I definitely agree

prime-dotnet-webapi/Services/Rules/AutomaticAdjudicationRules.cs
```diff
@@ -67,9 +67,19 @@ public override async Task<bool> ProcessRule(Enrollee enrollee)
...
             {
                 PharmanetCollegeRecord record = null;
+
+                var licenceNumber = cert.License.PrescriberIdType.Equals(PrescriberIdType.NA)
+                                ? cert.LicenseNumber
+                                : cert.PractitionerId;
+
+                if (cert.License.PrescriberIdType.Equals(PrescriberIdType.Optional) && cert.PractitionerId == null)
+                {
+                    return passed;
```
james-hollinger, line 77: I think this should be `continue`, rather than returning.

james-hollinger: At the same time might as well move this check up to line 69 since its an early exit clause you can check before deciding which number to send to pharmanet

james-hollinger (approved):
LGTM 👍

PR #1119, opened 2021-02-10
PRIME-1275 Three different roles on admin

johnnyl6:
- Angular permissions
- Dotnet permissions


prime-dotnet-webapi/Controllers/LookupsController.cs
```diff
@@ -14,8 +14,7 @@ namespace Prime.Controllers
     [Produces("application/json")]
     [Route("api/[controller]")]
     [ApiController]
-    // User needs at least the READONLY ADMIN or ENROLLEE role to use this controller
-    [Authorize(Policy = Policies.User)]
+    [Authorize(Roles = Roles.PrimeEnrollee + "," + Roles.ViewEnrollee + "," + Roles.ViewSite)]
```
james-hollinger, line 17: I think in this case, using `Roles.PrimeAdministrant` is appropriate, as the lookups are available to every user of PRIME

james-hollinger: We could also consider just using `[Authorize]` here. No permissions are required, only a valid token. This would mean that our service accounts could access these endpoints, which isn't a bad thing I don't think.

neophyte57: I've adjusted to just use `[Authorize]` for `LookupsController`

prime-angular-frontend/src/app/modules/auth/shared/services/permission.service.ts
```diff
@@ -0,0 +1,30 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class PermissionService implements IPermissionService {
+
+  constructor(
+    private accessTokenService: AccessTokenService
+  ) { }
+
+  public hasRoles(...roles: Role[]): boolean {
```
mtpultz, line 21: When you use the rest operator (`...roles`) you're saying that if I pass in `hasRoles(Role.1, Role.2, Role.3, ..., Role.N)` that you want it as an array of roles, but if I pass in an array of Roles `hasRoles([Role.1, Role.2, Role.3, ..., Role.N])` you would get `[[Role.1, Role.2, Role.3, ..., Role.N]]` so a developer consuming your method would need to know they have to invoke it like `hasRoles(...[Role.1, Role.2, Role.3, ..., Role.N]);`.  The rest operator is great in methods if you need a variadic method with infinity arity, but in this case it would be easier for developers to consume if you manage the difference between singular and multiples of a type within the method:

```js
public hasRoles(roles: Role | Role[]): boolean {
  roles = (Array.isArray(roles)) ? roles : [roles];
  const assignedRoles = this.accessTokenService.roles();
  return roles.every(r => assignedRoles.includes(r));
}
```

AnaisHebert (commented):
image
Changing the TOA type opens up the assign enrolment screen instead.

neophyte57:
> image
> Changing the TOA type opens up the assign enrolment screen instead.

Hmmm, seems to be behaving properly in the PR environment currently:

image

AnaisHebert (commented):
TOA selection looks good now, noticed that Limits and conditions isnt working
image
image

neophyte57:
> TOA selection looks good now, noticed that Limits and conditions isnt working
> image
> image

This error should be fixed now.

prime-angular-frontend/src/app/shared/components/dialogs/content/triage/triage.component.html
```diff
@@ -31,25 +31,25 @@ <h2 class="h5 mb-3">
...
     <button mat-menu-item
-            [disabled]="!canEdit || assigned"
+            [disabled]="!((Role.TRIAGE_ENROLLEE | inRole) || assigned)"
             (click)="onEscalate()">
       <mat-icon>error_outline</mat-icon>
       <span>Enrolment Escalation</span>
     </button>
     <button mat-menu-item
-            [disabled]="!canEdit"
+            [disabled]="!(Role.TRIAGE_ENROLLEE | inRole)"
```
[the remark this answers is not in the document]
johnnyl6: Added 'in' and 'notIn' support for the full version RolePipe. feels like a notInRole pipe could lead to logical ambiguity so not implementing it for now.

AnaisHebert (approved):
LGTM 🌞

mtpultz (approved):

james-hollinger (approved):
LGTM

PR #1122, opened 2021-02-10
Fix Auditable Properties

james-hollinger:
Created Id/Timestamp were supposed to be immutable, but were not. Also, were implemented in a inefficient way.


mtpultz (approved):
spaceman-spiff

PR #1124, opened 2021-02-11
Fix toggle on back

mtpultz:


mtpultz:
Placed into another PR

PR #1131, opened 2021-02-13
Fix for final step of site registration failing in SitesController

neophyte57:
`await _emailService.SendRemoteUserNotificationsAsync(site, site.RemoteUsers);` fails when there are no remote users.  Reproducible in `develop`. 

A unit test and fix


mtpultz (approved):
image

AnaisHebert (approved):
⛄⛄⛄❄

PR #1138, opened 2021-02-18
PRIME-1418 Send a message to wallet when credential is revoked

AnaisHebert:


prime-dotnet-webapi/Services/VerifiableCredentialService.cs
```diff
@@ -19,6 +19,7 @@ public class WebhookTopic
         public const string Connections = "connections";
         public const string IssueCredential = "issue_credential";
         public const string RevocationRegistry = "revocation_registry";
+        public const string BasicMessage = "basicmessages";
```
mtpultz, line 22: For consistency with the other this could be snake_cased

AnaisHebert: Yeah I expected it to be snake cased, but that is what they send us

mtpultz (approved):
image

hogmann (approved):
🚗 ⛵ 🏠

PR #1140, opened 2021-02-18
Missed Audit Columns on Addresses

james-hollinger:


AnaisHebert (approved):
A M A Z I N G

johnnyl6 (approved):
Good last minute catch. 
This makes me thinking why it was not caught by the compiler like so:

`
public class PartyAddress: AuditableEntity
`
and then some sort of database integrity unit test would catch the issue. I might have to get familiar with more with the database code.

PR #1141, opened 2021-02-20
Separate Pharmanet College API Definitions

james-hollinger:
Moving things around to make the College Licence client more like the other clients


mtpultz (approved):
image

PR #1142, opened 2021-02-22
PRIME-1402 Pharmanet API call Business Event

neophyte57:
To test how it handles error situations, I hacked up a copy of `prime-dotnet-webapi-tests` since the test project did not compile at the time.  Eventually I created a PR so that I could switch the `PHARMANET_API_URL` environment variable to something incorrect to simulate an API connection problem: 

image


prime-dotnet-webapi/Services/Rules/AutomaticAdjudicationRules.cs
```diff
@@ -98,10 +102,12 @@ public override async Task<bool> ProcessRule(Enrollee enrollee)
                 try
                 {
                     record = await _collegeLicenceClient.GetCollegeRecordAsync(cert.License.Prefix, licenceNumber);
+                    await _businessEventService.CreatePharmanetApiCallEventAsync(enrollee.Id, cert.License.Prefix, licenceNumber, $"API call to {Prime.PrimeEnvironment.PharmanetApi.Url} completed without basic errors.");
```
AnaisHebert, line 105: I wonder if having the word error in the successful log could be confusing?

AnaisHebert (commented):
I got this for my business event log, even though my record was not found and i was sent to manual because of it. I think that should be reflected in the log?
image
image

prime-dotnet-webapi/Configuration/BusinessEventTypeConfiguration.cs
```diff
@@ -17,7 +17,8 @@ public override IEnumerable<BusinessEventType> SeedData
                     new BusinessEventType { Code = 5, Name = "Enrollee", CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
                     new BusinessEventType { Code = 6, Name = "Site", CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
                     new BusinessEventType { Code = 7, Name = "Admin View", CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
-                    new BusinessEventType { Code = 8, Name = "Organization", CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE }
+                    new BusinessEventType { Code = 8, Name = "Organization", CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE },
+                    new BusinessEventType { Code = 9, Name = "Pharmanet API Call", CreatedTimeStamp = SEEDING_DATE, UpdatedTimeStamp = SEEDING_DATE }
```
hogmann, line 21: In the ticket it states "Add new businessEvent Type for PharmaNet query", not sure if "Pharmanet API Call" might be a little too much developer jargon since this is a business facing event. Personally i'm not sure which way to go, just wanted to raise it, in the case the admins can get the most value from these events.

neophyte57: Good point @hogmann , so I asked the question in the Jira ticket

neophyte57:
> I got this for my business event log, even though my record was not found and i was sent to manual because of it. I think that should be reflected in the log?
> image
> image

I made some changes to address both your comments @AnaisHebert

AnaisHebert (approved):
lgtm ☕

prime-dotnet-webapi/Services/Rules/AutomaticAdjudicationRules.cs
```diff
@@ -100,34 +104,45 @@ public override async Task<bool> ProcessRule(Enrollee enrollee)
...
+                    await _businessEventService.CreatePharmanetApiCallEventAsync(enrollee.Id, cert.License.Prefix, licenceNumber, "(Name Discrepancy) " + message);
                     passed = false;
                 }
                 if (record.DateofBirth.Date != enrollee.DateOfBirth.Date)
                 {
-                    enrollee.AddReasonToCurrentStatus(StatusReasonType.BirthdateDiscrepancy, $"{licenceText} returned {record.DateofBirth:d MMM yyyy}");
+                    string message = $"{licenceText} returned {record.DateofBirth:d MMM yyyy}";
+                    enrollee.AddReasonToCurrentStatus(StatusReasonType.BirthdateDiscrepancy, message);
+                    await _businessEventService.CreatePharmanetApiCallEventAsync(enrollee.Id, cert.License.Prefix, licenceNumber, "(Birthdate Discrepancy) " + message);
                     passed = false;
```
james-hollinger, line 134: If someone enters the wrong License Number, but manage to enter a number for someone else, it is likely that both the name and birthdate will not match. This will create two business events, which I don't think we want because it will be misleading to the adjudicators. Also, this is now rather redundant with the status reason.

james-hollinger: I feel the intention of this business event was primarily to inform the adjudicators whether or not a call was made to the API. If we want to log the response from the API, it should be done in one location.

prime-dotnet-webapi/Configuration/BusinessEventTypeConfiguration.cs
neophyte57: Nick thinks the wording is OK, see his comment: https://bcgovmoh.atlassian.net/secure/RapidBoard.jspa?rapidView=1&modal=detail&selectedIssue=PRIME-1402

prime-dotnet-webapi/Services/Rules/AutomaticAdjudicationRules.cs
neophyte57: My original implementation relied on `enrollee.AddReasonToCurrentStatus` and didn't have the `_businessEventService.CreatePharmanetApiCallEventAsync` within each of the conditional blocks.  But that didn't provide enough context in the business event message?  From Jira ticket:  "We should log what we sent to pharmanet, but not log the resulting dataset.  
Only record any mismatch in dataset and date/time of request."

For when both name and birthdate do not match, we can consolidate to a single business event re: something like enrollee likely entered license number of another practitioner?

neophyte57:
As discussed with Nick, Jason, and James, latest commit now records a business event to indicate (1) API error (in a generic manner), (2) no record, or (3) a record (in a generic manner).  Mismatches and other rules are captured as status change reasons, not in a business event.

james-hollinger (approved):
LGTM

PR #1146, opened 2021-02-23
PRIME-1414 remote user info should be deleted if license type is updated

johnnyl6:
- Added warning msg when license type changes to no long eligible for remote access
-Removing remote user data onsubmit when applicable


prime-angular-frontend/src/app/modules/enrolment/pages/regulatory/regulatory.component.ts
```diff
@@ -84,6 +85,11 @@ export class RegulatoryComponent extends BaseEnrolmentProfilePage implements OnI
     this.routeTo(EnrolmentRoutes.CARE_SETTING);
   }
 
+  public shouldRemoveRemoteAccess(): boolean {
+    return this.hasRemoteAccess && !this.isInitialEnrolment &&
```
mtpultz, line 89: Feels like the best solution might be a `valueChanges` listener in `initForm`, which would mean you could just use the `*ngIf` directive with `!this.hasRemoteAccess && !this.isInitialEnrolment` and not have to invoke a method in the controller when it can be avoided:

```
this.form.valueChanges
  .subscribe(() => this.hasRemoteAccess = this.canRequestRemoteAccess());
```

The advantage of doing it this way means: 
1. It only checks when an update occurs to the form
2. Preserves the use of the member variable `hasRemoteAccess`, which was only being used on initialization until the first form update, and 
3. Removes a method that would be invoked by the template 100s of times per minute depending on how the user interacts with the application as it would keep checking for changes that can only occur when the form changes.  You can test this out by logging the form updates in the subscribe vs the `shouldRemoveRemoteAccess` method

johnnyl6: Good thinking. However, the *ngIf checks and display the warning message when enrollee currently has remote access but changes to a licence type which is no longer eligible. If this.hasRemoteAccess is subscribed to the form.valuechanges,  and init to be false, which leads to the warning messages to be displayed. This is not expected for an enrollee currently without remote access.
This idea is to initialize hasRemoteAccess and it indicates the current status of remote access before any changes. Then we determine if the enrollee has changes in license type in the form by calling canRequestRemoteAccess. The shouldRemoveRemoteAccess is just the final logical check. I feel a bit cleaner having a descriptive function than writing the logic directly like *ngIf="this.hasRemoteAccess && !this.isInitialEnrolment && !this.canRequestRemoteAcccess()" If there is a better solution, let me know 😊

johnnyl6: Agreed with the extra firing of the method. Changing it subscribe to valuchanges.

> 3. Removes a method that would be invoked by the template 100s of times per minute depending on how the user interacts with the application as it would keep checking for changes that can only occur when the form changes.  You can test this out by logging the form updates in the subscribe vs the `shouldRemoveRemoteAccess` method

prime-angular-frontend/src/app/modules/enrolment/pages/regulatory/regulatory.component.ts
```diff
@@ -87,6 +89,11 @@ export class RegulatoryComponent extends BaseEnrolmentProfilePage implements OnI
   public ngOnInit() {
     this.createFormInstance();
     this.patchForm().subscribe(() => this.initForm());
+
+    this.form.valueChanges.subscribe(() => {
+      this.shouldRemoveRemoteAccess = this.hasRemoteAccess && !this.isInitialEnrolment &&
```
mtpultz, line 94: I'm probably missing something, or maybe my comment wasn't well worded, but you shouldn't need a new member variable to track the changes.  My comment was indicating that `hasRemoteAccess` is being orphaned after it is initially set, and that it can be more thoroughly used:

1. Having `this.hasRemoteAccess = this.canRequestRemoteAccess();` set in `initForm` and `subscribe`  so it is being reused
2. Make `hasRemoteAccess` as a public member variable for use in the template, and
3. In the template use *ngIf="!hasRemoteAccess && !initialEnrolment" so changes in the value are reflected in the UI

Let me know if this doesn't work for some reason, and maybe you can catch me up on what I'm not seeing.

johnnyl6: `hasRemoteAccess ` is not orphaned but used to store the initial status of enrolle's "have remote access". If you subscribe it, the value will be overwritten and the ngIf logic is no longer correct. 

For example, 
1. enrollee has CPBC pharmacist licence type hence no remote access. 
2. Hitting the page to modify licence type,  `hasRemoteAccess ` is initialized to false.  and ngIf evaluated to be true. The warning message will be displayed. This is not expected. I don't think we need to remind enrollee in this case.

prime-angular-frontend/src/app/modules/enrolment/pages/regulatory/regulatory.component.ts
mtpultz: @johnnyl6 thanks, so instead of using a member variable that only has a singular purpose of indicating an initial value I would use a closure.  To keep with self documenting code we could name it `initialRemoteAccess`, and also rename `shouldRemoveRemoteAccess` to `cannotRequestRemoteAccess` which aligns a better with its compliment `canRequestRemoteAccess`:

```
public ngOnInit() {
  this.createFormInstance();
  this.patchForm().subscribe(() => this.initForm());

  const initialRemoteAccess = this.canRequestRemoteAccess();

  this.form.valueChanges
    .pipe(map((_) => initialRemoteAccess))
    .subscribe((initialRemoteAccess: boolean) =>
      this.cannotRequestRemoteAccess = initialRemoteAccess && !this.canRequestRemoteAccess() && !this.isInitialEnrolment
    );
}
```

The only other change I would make is initializing `this.cannotRequestRemoteAccess` in the `constructor` instead of `initForm`.

mtpultz (approved):

prime-angular-frontend/src/app/modules/enrolment/pages/regulatory/regulatory.component.ts
```diff
@@ -104,11 +109,23 @@ export class RegulatoryComponent extends BaseEnrolmentProfilePage implements OnI
     if (!this.certifications.length) {
       this.addEmptyCollegeCertification();
     }
+
+    const initialRemoteAccess = this.canRequestRemoteAccess();
+
+    this.form.valueChanges
```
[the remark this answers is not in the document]
mtpultz: @johnnyl6 let me know if you have any questions, otherwise this PR is ready image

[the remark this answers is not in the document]
johnnyl6: > @johnnyl6 let me know if you have any questions, otherwise this PR is ready image

Good finding. LGTM✌

PR #1148, opened 2021-02-24
PRIME-1425 Hide "upload documents" from clerks

johnnyl6:
- using  the Role.MANAGE_ENROLLEE (enrollee_elevated_management) which currently only Manager has
- this will also hide the button from analyst.


james-hollinger (changes requested):
image

The corresponding role on the back end is enrollee_approve, which analysts have.
This should either be enrollee_approve in the front end, or manage in the back end,  depending on who should be allowed to do this.

james-hollinger (approved):
LGTM

PR #1149, opened 2021-02-24
PRIME-1428 - Update Keycloak with new roles/groups

james-hollinger:
- Removed old feature flags from PROD
 - Removed old roles from DEV
 - Added new groups and roles to TEST and PROD


documentation/Keycloak settings/realm-export-prod.json
```diff
@@ -66,17 +57,17 @@
...
         "clientRole": false,
         "containerId": "v4mbqqas",
         "attributes": {}
       },
       {
-        "id": "839e1a7e-31a5-43bd-b303-8e8f6f00bbf6",
-        "name": "feature_site_registration",
-        "description": "Site Registration feature flag",
+        "id": "5f4c16dd-507f-4e8f-b44a-b2a0e188f248",
+        "name": "prime_readonly_admin",
```
johnnyl6, line 70: This role is not referenced in either frontend or backend. Any reason to still keep it in prod?

james-hollinger: The code currently deployed in prod still references these old roles. We cannot remove them until the new code is deployed all the way into prod.

johnnyl6 (approved):
LGTM🎉

PR #1154, opened 2021-02-26
Updated config type and generic

mtpultz:


prime-angular-frontend/src/app/config/config.service.ts
```diff
@@ -160,8 +160,8 @@ export class ConfigService implements IConfigService {
    * @description
    * Sort the configuration by weight.
    */
-  private sortConfigByWeight(): (a: LicenseConfig, b: LicenseConfig) => SortWeight {
-    return (a: LicenseConfig, b: LicenseConfig) =>
-      this.utilsService.sortByKey<LicenseConfig>(a, b, 'weight');
+  private sortConfigByWeight<T extends IWeightedConfig>(): (a: T, b: T) => SortWeight {
+    return (a: T, b: T) =>
+      this.utilsService.sortByKey<T>(a, b, 'weight');
```
johnnyl6, line 165: Feels like we keep adding private methods like `sortConfigByXXX()`  in `ConfigService `class., and it's merely just a shorthand of calling the `sortConfigByKey `method. 

Since these sorting methods are just private methods and not consumed outside of `ConfigService`, it's unlikely to cause problems elsewhere. Why not directly use `sortConfigByKey<T>('name')`, `sortConfigByKey<T>('code')`, and `sortConfigByKey<T>('weight') `? Also, it's strange to have sorting methods in a `ConfigService `class, maybe move somewhere else to be extension methods if really needed.

mtpultz: Yah, we can definitely remove those methods out of the config.

mtpultz: I just wanted to fix the typings since they weren't generic

johnnyl6: > I just wanted to fix the typings since I noticed they weren't generic, but can definitely remove them. Do you mind taking that on @johnnyl6?

@mtpultz  I did some refactoring. Let me know if they are okay.

mtpultz: Look great.  Thanks @johnnyl6

johnnyl6 (approved):

prime-angular-frontend/src/app/core/services/utils.service.ts
```diff
@@ -77,8 +77,8 @@ export class UtilsService {
    * @description
    * Generic sorting of a JSON object by key.
    */
-  public sortByKey<T>(a: { [key: string]: any }, b: { [key: string]: any }, key: string): SortWeight {
-    return this.sort<T>(a[key], b[key]);
+  public sortByKey<T extends { [key: string]: any }>(key: string): (a: T, b: T) => SortWeight {
```
mtpultz, line 80: Looks great @johnnyl6.  I made an update to improve the typing for `sortByKey` since this method is  extending from an non-specific object literal instead a specific one so if I sort by a key that doesn't exist on the object there's no protection:

**Was**
```
public sortByKey<T extends { [key: string]: any }>(key: string): (a: T, b: T) => SortWeight { ... }
```

**Now**
```
public sortByKey<T extends { [key: string]: any }>(key: keyof T): (a: T, b: T) => SortWeight { ... }
````

PR #1156, opened 2021-02-27
Missing Site Address

mtpultz:
A missing site address error can occur when creating a site, but exiting before adding a site address to the site management


AnaisHebert (approved):
Seems to work 🤸‍♀️

PR #1157, opened 2021-02-28
Remote access fields not being patched

mtpultz:
Remote access fields (physicalAddress and siteVendors) were not being patched outside of the remote access page since they weren't in the form state and were showing up as dashes in overview since the fields were always null.


AnaisHebert (approved):
🎈🎈

PR #1158, opened 2021-02-28
Remote access request button hidden when not editable

mtpultz:


neophyte57:
I was trying to test this but ended up with this situation.  I think it started by clicking on "PRIME History", getting the toast notification in the screenshot, and then clicking the other left menu items.

image

After F5 refresh, it's behaving OK.

neophyte57:
Another thing unrelated to PR change but I noticed while testing.  I lock the enrollee but the Status History doesn't refresh ... should it?

image

AnaisHebert (approved):
🎈🎈🎈🎈

neophyte57 (approved):
I notice that in `develop`, when enrollee is Under Review or approved (but before enrollee accepts ToA), the "Request Remote Access" button is visible but not in `pr-1158` so fix looks good

mtpultz:
@neophyte57 I made a note of those issues you noticed and will PR some fixes.  Nice catch!

PR #1160, opened 2021-03-03
Only invoke current-status when opened

mtpultz:
Enrollee list view was invoking current-status for each triage component, and not only invokes current-status when the triage overlay is opened.


johnnyl6 (approved):
LGTM🎉

PR #1161, opened 2021-03-03
Updated caddy download url in schemaspy

johnnyl6:
- old URL no longer exists


neophyte57 (approved):
Looks good, thanks for fixing this annoying issue  :-)

PR #1162, opened 2021-03-04
PRIME-1408 Secure PLR HL7 API using Client Cert

neophyte57:
This requires a bit of setup to test locally.  I'll probably have to work with @ajdeziel to make changes to the other `nginx.conf` files that are used in the OpenShift environments, after local testing is verified.

# Setup dotnet-webapi #
1. Trust the ASP.NET Core HTTPS development certificate:  https://docs.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-2.2&tabs=visual-studio#trust-the-aspnet-core-https-development-certificate-on-windows-and-macos
2. Verify no issues hitting https://localhost:5001/api/PLRHL7 with Chrome.  Should see the WSDL file 

# Setup Nginx locally #
1. Install Nginx locally [problems as nginx Docker container, due to webapi not running as Docker container]
2. Copy development certificate and private key (`server.crt` and `server.key`) to Nginx conf directory from `prime-dotnet-webapi-tests\Utils\Auth` 
3. Export "ASP.NET Core HTTPS development certificate" from Personal Certificates of Current User (on Windows, Windows key + R > certmgr.msc), convert to .pem file with `openssl x509 -inform der -in backend-api.cer -out backend-api.pem`, and copy to Nginx conf directory
4. Replace default nginx configuration with `nginx-plrhl7.conf` (for example, `C:\Users\177092\Source\Repos\moh-prime\prime-dotnet-webapi-tests\Integration> cp .\nginx-plrhl7.conf C:\Programs\nginx-1.18.0\conf\nginx.conf`)
5. Start nginx (for example, `C:\Programs\nginx-1.18.0> .\nginx.exe`) 
6. Start dotnet-webapi
7. Send requests to dotnet-webapi via nginx proxy over HTTPS (for example, `~/Source/Repos/moh-prime/prime-dotnet-webapi-tests/Utils/Auth $  curl --insecure --cert client.crt:changeme --key client.key --data-binary "@../../Integration/PRPM_IN301030CA.xml" https://127.0.0.1:443/api/PLRHL7`)
8. Setup expected client certificate thumbprint like this:  `setx PLR_INTEGRATION_CLIENT_CERT_THUMBPRINT 6FA357305008C8EA814EA98768C21783DCCA374E`.  A certificate's thumbprint can be determined like this `openssl x509 -in client.crt -fingerprint -noout` although the output will have colons inserted for human-readability, e.g. "SHA1 Fingerprint=6F:A3:57:30:50:08:C8:EA:81:4E:A9:87:68:C2:17:83:DC:CA:37:4E"

Most of the file additions are just resources for testing.  

At this point, no need to wait for PR OpenShift environment to test :-)


neophyte57:
At this point, no need to wait for PR OpenShift environment to test  :-)

prime-dotnet-webapi/Program.cs
```diff
@@ -39,7 +40,15 @@ public static int Main(string[] args)
...
         public static IHostBuilder CreateHostBuilder(string[] args) =>
             Host.CreateDefaultBuilder(args)
-                .ConfigureWebHostDefaults(webBuilder => webBuilder.UseStartup<Startup>())
+                .ConfigureWebHostDefaults(webBuilder =>
+                {
+                    // webBuilder.ConfigureKestrel(o =>
+                    //     {
+                    //         o.ConfigureHttpsDefaults(o =>
+                    //             o.ClientCertificateMode = ClientCertificateMode.RequireCertificate);
+                    //     });
```
james-hollinger, line 49: Is this commented code intentional?

neophyte57: Yes, the webapi should **not** be expecting the client certificate when establishing the HTTPS connection since there will be proxies in between; see https://docs.microsoft.com/en-us/aspnet/core/security/authentication/certauth?view=aspnetcore-5.0#proxy-and-load-balancer-scenarios.  The client certificate will come over a HTTPS header.

But no, we don't need the commented-out code to stay in this file.

mtpultz (dismissed):
image

prime-dotnet-webapi/Services/SoapService.cs
```diff
@@ -22,9 +21,40 @@ public class SoapServiceOperationTuner : IServiceOperationTuner
 
         public void Tune(HttpContext httpContext, object serviceInstance, SoapCore.ServiceModel.OperationDescription operation)
         {
+            // This must be the header used by proxies that first receive the client certificate
+            // (and pass on via HTTP Header)
+            const string CLIENT_CERT_HEADER = "X-SSL-CERT";
```
james-hollinger, line 26: There are already some constants for this service defined on lines 19/20, do you think this should go there?
Also, I tend to avoid using screaming snake case, preferring to have consistent PascalCase instead.

prime-dotnet-webapi/Services/SoapService.cs
```diff
@@ -22,9 +21,40 @@ public class SoapServiceOperationTuner : IServiceOperationTuner
...
+
             if (serviceInstance is SoapService service)
             {
-                service.DocumentRoot = GetRequestBody(httpContext, Prefix, Uri, operation.Name);
+                string urlEncoded;
+                StringValues paramValue;
+                if (httpContext.Request.Headers.TryGetValue(CLIENT_CERT_HEADER, out paramValue))
+                {
+                    urlEncoded = paramValue[0];
+                }
```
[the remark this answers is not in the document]
james-hollinger: This entire section of code seems like a useful extension method, something like:

```C#
public static class HttpRequestExtensions
{
    public static string GetHeader(this HttpRequest request, string name)
    {
        if (httpContext.Request.Headers.TryGetValue(name, out var value)
            {
                return value.SingleOrDefault();
            }
            else
            {
                return null;
            }
    }
}
```

prime-dotnet-webapi/Services/SoapService.cs
```diff
@@ -22,9 +21,40 @@ public class SoapServiceOperationTuner : IServiceOperationTuner
...
+                if (PrimeEnvironment.PlrIntegration.ClientCertThumbprint == null)
+                {
+                    throw new SystemException("Receiving system is not configured properly; please advise system administrator.");
+                }
+
+                string decoded = HttpUtility.UrlDecode(urlEncoded);
+                // https://stackoverflow.com/questions/65349878/c-sharp-convert-certificate-string-into-x509-certificate
+                byte[] certAsBytes = Encoding.ASCII.GetBytes(decoded);
+                var clientCert = new X509Certificate2(certAsBytes);
+                if (clientCert.Thumbprint.Equals(PrimeEnvironment.PlrIntegration.ClientCertThumbprint))
```
james-hollinger, line 50: This line reads slightly easier if you just use == rather than .Equals

prime-dotnet-webapi/Services/SoapService.cs
```diff
@@ -22,9 +21,40 @@ public class SoapServiceOperationTuner : IServiceOperationTuner
...
+                    throw new ArgumentException("Client certificate expected to be provided.");
+                }
+
+                if (PrimeEnvironment.PlrIntegration.ClientCertThumbprint == null)
+                {
+                    throw new SystemException("Receiving system is not configured properly; please advise system administrator.");
+                }
+
+                string decoded = HttpUtility.UrlDecode(urlEncoded);
+                // https://stackoverflow.com/questions/65349878/c-sharp-convert-certificate-string-into-x509-certificate
```
james-hollinger, line 47: I feel the following lines are rather self explanatory and don't need this link. 
X509Certificate2 basically only has two types of data sources: certs stored on disk; and byte arrays.

prime-dotnet-webapi/Services/SoapService.cs
```diff
@@ -22,9 +21,40 @@ public class SoapServiceOperationTuner : IServiceOperationTuner
...
+                byte[] certAsBytes = Encoding.ASCII.GetBytes(decoded);
+                var clientCert = new X509Certificate2(certAsBytes);
+                if (clientCert.Thumbprint.Equals(PrimeEnvironment.PlrIntegration.ClientCertThumbprint))
+                {
+                    service.DocumentRoot = GetRequestBody(httpContext, Prefix, Uri, operation.Name);
+                }
+                else
+                {
+                    throw new ArgumentException($"The certificate with thumbprint {clientCert.Thumbprint} is invalid to the receiving system.");
```
james-hollinger, line 56: I would lean away from echoing information when handling certs, even client provided information.

neophyte57:
The PR can be tested in OpenShift now, for example, `curl  --insecure --cert client.crt:changeme --key client.key --data-binary "@../../Integration/PRPM_IN301030CA.xml" https://plrhl7.pr-1162.pharmanetenrolment.gov.bc.ca/api/v1/PLRHL7`

neophyte57:
@ajdeziel In order to use a recent version of Nginx (rather than 1.10.3, from Jan 2017), I converted `openshift.dockerfile` to a multi-stage build.  Front-end appears to continue to work fine but please review the changes:  

I mainly moved the existing Docker commands around, without changing the commands, despite some concerns about them (`chmod -R 777`?!?)

ajdeziel (dismissed):
> @ajdeziel In order to use a recent version of Nginx (rather than 1.10.3, from Jan 2017), I converted `openshift.dockerfile` to a multi-stage build. Front-end appears to continue to work fine but please review the changes:
> 
> I mainly moved the existing Docker commands around, without changing the commands, despite some concerns about them (`chmod -R 777`?!?)

The changes you've made to the frontend's `Dockerfile` are completely sensible as this is work I've already done as part of the OCP 4 migration (changes yet to be merged) to refactor it to a standard Docker design pattern.

As for the concerns around `chmod 777`, I've also addressed these in my changes so it's ok to leave them here for now as they won't be around for much longer.

I'd also ask that any work you've set up in the OCP 3 console (Config Maps, Secrets) to also be added to the OCP 4 console as well. I can walk you through that if you'd like.

neophyte57:
> > @ajdeziel In order to use a recent version of Nginx (rather than 1.10.3, from Jan 2017), I converted `openshift.dockerfile` to a multi-stage build. Front-end appears to continue to work fine but please review the changes:
> > I mainly moved the existing Docker commands around, without changing the commands, despite some concerns about them (`chmod -R 777`?!?)
> 
> The changes you've made to the frontend's `Dockerfile` are completely sensible as this is work I've already done as part of the OCP 4 migration (changes yet to be merged) to refactor it to a standard Docker design pattern.
> 
> As for the concerns around `chmod 777`, I've also addressed these in my changes so it's ok to leave them here for now as they won't be around for much longer.
> 
> I'd also ask that any work you've set up in the OCP 3 console (Config Maps, Secrets) to also be added to the OCP 4 console as well. I can walk you through that if you'd like.

Thanks for the reminder @ajdeziel.  I manually added this to the Environment:

image

Should I add this to `openshift\dotnet-webapi.dc.yaml`, although the thumbprint value would be different in other environments like production ...?  We can talk about this further on MS Teams, etc.

ajdeziel:
> > > @ajdeziel In order to use a recent version of Nginx (rather than 1.10.3, from Jan 2017), I converted `openshift.dockerfile` to a multi-stage build. Front-end appears to continue to work fine but please review the changes:
> > > I mainly moved the existing Docker commands around, without changing the commands, despite some concerns about them (`chmod -R 777`?!?)
> > 
> > 
> > The changes you've made to the frontend's `Dockerfile` are completely sensible as this is work I've already done as part of the OCP 4 migration (changes yet to be merged) to refactor it to a standard Docker design pattern.
> > As for the concerns around `chmod 777`, I've also addressed these in my changes so it's ok to leave them here for now as they won't be around for much longer.
> > I'd also ask that any work you've set up in the OCP 3 console (Config Maps, Secrets) to also be added to the OCP 4 console as well. I can walk you through that if you'd like.
> 
> Thanks for the reminder @ajdeziel. I manually added this to the Environment:
> 
> image
> 
> Should I add this to `openshift\dotnet-webapi.dc.yaml`, although the thumbprint value would be different in other environments like production ...? We can talk about this further on MS Teams, etc.

Yes, it should be added to OpenShift. It'll obviously be different for each environment, but that's par for the course because we'll be specifying it under each environment.

itchison:
Please don't merge this until you've talked to me. Thanks! @neophyte57 @ajdeziel @james-hollinger

neophyte57:
I removed unnecessary lines from the `server` block in `nginx.template.conf` so that there is less duplication and renamed the subdomain so that it is more generic ("mutualauth") in case there are future end-points that we also want to enforce authentication using a client certificate.  

So now you can hit the PLR HL7 end-point like this `curl  --insecure --cert client.crt:changeme --key client.key --data-binary "@PRPM_IN303030CA.xml" https://mutualauth.pr-1162.pharmanetenrolment.gov.bc.ca/api/v1/PLRHL7`

I smoke-tested that Enrollment and the Admin areas of the application continue to work.

mtpultz (approved):
image

neophyte57:
@itchison @james-hollinger @ajdeziel Are we good to merge this into `develop`?  Or even ask Vinder of PLR team to test sending distribution messages to PR environment at `https://mutualauth.pr-1162.pharmanetenrolment.gov.bc.ca/api/v1/PLRHL7`?

mtpultz:
@neophyte57 I didn't want to merge this last week since we hadn't pushed to test in awhile and didn't wanted PLR going further than develop.  I've pushed out the latest increment into test so I don't see any reason not to merge so we can continue pushing PLR forward this sprint.

PR #1163, opened 2021-03-09
PRIME-1438 Need to support existing vs new site id for all site registration

johnnyl6:
- site id required for existing site
- site id not required for new site
- included PEC field in both frontend form & backend SiteUpdateModel


mtpultz (approved):
image

PR #1165, opened 2021-03-09
Added busy to block until request is finished

mtpultz:
Fix for bug when approving an enrolment by blocking the screen until prerequisite request is completed

Bug could not be replicated locally, but appears to be the most likely issue if her network connection is sooo super uber slow that she can open the dropdown, select approve, have the modal open, and either enter/skip adding a note before submitting.


johnnyl6 (approved):
🚀

PR #1169, opened 2021-03-11
Frontend unit test fix

johnnyl6:
- fixed failed angular unit tests
- fixed broken angular unit tests which did not run properly
- added null checks (mostly caught by unit test)
- refactored onlyBc directive to inBc (caught by unit test)
- increased timeout in karma.conf.js due to slower machine could never run the test


prime-angular-frontend/src/app/lib/utils/string-utils.class.ts
```diff
@@ -1,7 +1,8 @@
 export class StringUtils {
   public static capitalize(value: string): string {
+    const strValue: String = new String(value);
```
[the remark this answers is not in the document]
johnnyl6: Explicitly wrapping the primitive string into a String object, so that the unit test is happy. Was complaining about charAt() and splice() is undefined function.

prime-angular-frontend/src/app/modules/site-registration/pages/organization-agreement/organization-agreement.component.ts
```diff
@@ -141,6 +141,9 @@ export class OrganizationAgreementComponent implements OnInit, IPage {
     const organization = this.organizationService.organization;
     const siteId = this.route.snapshot.params.sid;
     this.isCompleted = organization?.completed;
+    if (!organization) {
```
mtpultz, line 144: @johnnyl6  my only feelings on the null checks like this are if the main modal (site, org, enrolment, etc) doesn't exist the page should fail since you shouldn't be able to make it to the page without that model.  So if we go this route we'll to update the templates to be user friend when this occurs by hiding content and showing toasts, or we don't worry about null checks on the main models and add the mocks to each test, as well as, the schema for shallow testing.

johnnyl6: Agreed. The component is only shared within PRIME and unlikely to be reused as a truly 3rd party lib, which requires much more validations on external input values. Will be going to the route to modify the test with valid mocks.

mtpultz (approved):

mtpultz:
image

PR #1171, opened 2021-03-12
Enrollee History Submission

mtpultz:
* Change endpoint URL to current API
* Update template to prevent failing during load
* Added missing verified address fields


AnaisHebert (commented):
Should we add the verified address to the enrollee review? It is also above in the BCSC information

AnaisHebert (approved):
Otherwise, looks good! 🤸‍♀️

PR #1172, opened 2021-03-13
PRIME-1429 - Remove Old Roles From Test Keycloak

james-hollinger:
Now that we have deployed to test, removed old roles from keycloak.


mtpultz (approved):
🍦 🥪

PR #1176, opened 2021-03-16
PRIME-1409 Intake PLR seed data

neophyte57:
Console app that saves into database.


prime-dotnet-webapi/Models/PlrProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        public DateTime StatusStartDate { get; set; }
+
+        public DateTime StatusExpiryDate { get; set; }
+
+        public string Expertise { get; set; }
+
+        public string Languages { get; set; }
+
+        public string Address1_Line1 { get; set; }
```
james-hollinger, line 48: I would prefer if we followed standard C# naming conventions and avoided underscores in property names

neophyte57:
I think Vinder (with PLR) and I have worked out the necessary columns PRIME will need:  An updated sample Excel file has been committed.  Should be suitable to do "final" review.  @james-hollinger

mtpultz (approved):
image

PR #1180, opened 2021-03-16
PRIME-1281 GIS Enrolment 

AnaisHebert:


prime-dotnet-webapi/Controllers/GisController.cs
```diff
@@ -0,0 +1,210 @@
...
+            }
+            if (!gisEnrolment.Party.PermissionsRecord().AccessableBy(User))
+            {
+                return Forbid();
+            }
+
+            return Ok(ApiResponse.Result(gisEnrolment));
+        }
+
+        // GET: api/Gis/5
```
james-hollinger, line 124: this should be api/gis

prime-dotnet-webapi/Controllers/GisController.cs
```diff
@@ -0,0 +1,210 @@
...
+using Microsoft.AspNetCore.Mvc;
+
+using Prime.Services;
+using Prime.Models.Api;
+using Prime.ViewModels.Parties;
+
+namespace Prime.Controllers
+{
+    [Produces("application/json")]
+    [Route("api/[controller]")]
```
james-hollinger, line 12: I think it would make more sense if this route was api/parties/[controller], like the phsa enrolments.

prime-dotnet-webapi/Controllers/GisController.cs
```diff
@@ -0,0 +1,210 @@
...
+    [ApiController]
+    public class GisController : ControllerBase
```
james-hollinger, line 14: These endpoints are not authenticated; I assume that is on purpose until we have the MOH Keycloak integration working?

prime-dotnet-webapi/Controllers/GisController.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// <param name="gisId"></param>
+        /// <param name="changeModel"></param>
+        [HttpPut("{gisId}", Name = nameof(UpdateGisEnrollee))]
+        // [Authorize(Roles = Roles.PrimeEnrollee)]
+        [ProducesResponseType(typeof(ApiBadRequestResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        public async Task<IActionResult> UpdateGisEnrollee(int gisId, GisChangeModel changeModel)
```
james-hollinger, line 69: These first three controller methods have different return types:
```C#
Task<ActionResult>
Task<IActionResult>
Task<ActionResult<GisViewModel>>
```
I feel we should change them all to Task<ActionResult> for consistency. We are not actually using the feature offered by typing our ActionResult(s) anywhere in our app.

prime-dotnet-webapi/Controllers/GisController.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        // POST: api/gis
+        /// <summary>
+        /// Creates a new Gis Enrolment
+        /// </summary>
+        [HttpPost(Name = nameof(CreateGisParty))]
+        [ProducesResponseType(typeof(ApiBadRequestResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status200OK)]
```
james-hollinger, line 30: This is slightly incorrect, this endpoint returns a 201 with the GIS in the body, so is more accurately `[ProducesResponseType(typeof(ApiResultResponse<GisViewModel>), StatusCodes.Status201Created)]`

prime-dotnet-webapi/Controllers/GisController.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// <summary>
+        /// Gets a specific Gis Enrolment using the logged in User.
+        /// </summary>
+        [HttpGet("", Name = nameof(GetGisEnrolmentByUserId))]
+        [ProducesResponseType(typeof(ApiBadRequestResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(typeof(ApiResultResponse<GisViewModel>), StatusCodes.Status200OK)]
+        public async Task<ActionResult<GisViewModel>> GetGisEnrolmentByUserId()
```
james-hollinger, line 134: This is how we do this kind of thing in our app currently for other types of users (Enrollees, etc), but I would prefer to do away with this pattern (and eventually refactor those other endpoints as well).  I feel this should be an endpoint that gets the enrollment by UserId (api/parties/gis/{userId:guid}) and verifies the UserId, rather that silently pulling the UserId from the token "behind the scenes". It would also allow Admin users to hit this endpoint if we needed to.

mtpultz (approved):
🏖 🦆

prime-dotnet-webapi/HttpClients/LDAP Api/LdapClient.cs
```diff
@@ -0,0 +1,92 @@
...
+            : base(PropertySerialization.CamelCase)
+        {
+            // Auth header and api-key are injected in Startup.cs
+            _client = client;
+            _logger = logger;
+        }
+
+        public async Task<JObject> GetUserAsync(string username, string password)
+        {
+            JObject messageObject = new JObject
```
james-hollinger, line 27: Does this have to be a JObject? I think we should avoid using JObject when its just as convienient to do:
```C#
var messageObject = new 
{
    UserName = username,
    Password = password
}
```

prime-dotnet-webapi/HttpClients/LDAP Api/LdapClient.cs
```diff
@@ -0,0 +1,92 @@
...
+            if (!response.IsSuccessStatusCode)
+            {
+                await LogError(response);
+            }
+
+            _logger.LogInformation("CONTENT RESPONSE: {body}", await response.Content.ReadAsStringAsync());
+
+            try
+            {
+                JObject body = JObject.Parse(await response.Content.ReadAsStringAsync());
```
james-hollinger, line 54: Can we de-serialize this to a known class rather than JObject?
```C#
private class GisUser
{
    public string GisUerRole {get; set;}
}

```

prime-dotnet-webapi/Services/GisService.cs
```diff
@@ -0,0 +1,119 @@
...
+                await _context.SaveChangesAsync();
+            }
+
+            return success;
+        }
+
+        public async Task<int> SubmitApplicationAsync(int gisId)
+        {
+            var gisEnrolment = await _context.GisEnrolments
+                .Include(g => g.Party)
```
james-hollinger, line 111: This include is not necessary.

PR #1184, opened 2021-03-17
PRIME-1470 Site ID Updates

johnnyl6:
- merge site name and site ID into one reusable component
- added site id to site registration overview page
- added site id to site registration email pdf template
- removed site ID from care setting page


mtpultz (approved):
image

PR #1195, opened 2021-03-24
Site Name Dropdown

mtpultz:
Don't display the dropdown for doing business as choices when there are no options

NOTE: built off Clean Up so will become less noisy once that PR is approved and merged into develop


johnnyl6 (approved):
🎉

AnaisHebert (approved):
👻👻

PR #1199, opened 2021-03-24
Business Licence Deferred Reason Toggle Undefined

mtpultz:
* Use of ViewChild on a component that does not always exist in the template

NOTE: built off FormState Memory Leaks so will become less noisy once that PR is approved and merged into develop


AnaisHebert (approved):
🧨🧨🧨

PR #1200, opened 2021-03-24
Bug: Reset Deferred Reason when Care Setting Changed

mtpultz:
* Deferred reason blocks updates to the Site Name if the care setting was updated before submission of the site from Comm Pharm to Comm Prac

Testing Case 1:
1. Create a site as Comm Pharm with any vendor
2. Don't make any changes to Business Licence
3. Route back to Care Setting and change to Comm Prac with any vendor
4. Observe network and there should be no PUT to update business licence when no deferred reason exists

Testing Case 2:
1. Create a site as Comm Pharm with any vendor
2. Add a deferred reason on business licence upload
3. Route back to Care Setting and change to Comm Prac with any vendor
4. Observe network for PUT to remove deferred reason if it exists
5. Observe on next page GET for deferred reason of null

Test Case 3:
1. Create a site as Comm Pharm with any vendor
2. Add a deferred reason on business licence upload
3. Upload a business licence document as well
3. Route back to Care Setting and change to Comm Prac with any vendor
4. Observe network and there should be no PUT since business licences can't be updated after a document has been uploaded so the deferred reason will persist

Test Case 4:
When the deferred reason is reset it also resets the site completed status forcing wizard-like routing versus spoking.
1. Create a site as Comm Pharm with any vendor
2. Add a deferred reason on business licence upload
3. Take the site to completion, but don't submit
4. From overview route to care setting and immediately save making no changes; you should be routed to overview
5. From overview route to care setting, change a vendor, and save; you should be routed to overview
5. From overview route to care setting and update to Comm Prac with any vendor
5. Should routing in wizard mode forcing the user to upload a missing business licence

NOTE: built off Business Licence Toggle Undefined so will become less noisy once that PR is approved and merged into develop


AnaisHebert (approved):
Amazing-designstyle-amazing-m

PR #1201, opened 2021-03-25
Bug: Site Registration Note Save Button Disabled

johnnyl6:
- save button was improperly disabled when typing notes


mtpultz (approved):
image

PR #1202, opened 2021-03-26
Bug: Fixed Role for Assign/Reassign Button in Enrolment

johnnyl6:
- made it consistent with backend which uses 'enrollee_triage'


mtpultz (approved):
image

PR #1205, opened 2021-03-27
Bug: Remote User

mtpultz:
Addition of new remote user was not working properly, and found while fixing the FE tests


prime-angular-frontend/src/app/modules/site-registration/pages/remote-user-page/remote-user-page.component.ts
```diff
@@ -71,14 +72,22 @@ export class RemoteUserPageComponent extends AbstractEnrolmentPage implements On
...
+   * Remote user certifications specific to the local form.
+   */
+  public get remoteUserCertifications(): FormArray {
+    return this.form.get('remoteUserCertifications') as FormArray;
+  }
+
   public get selectedCollegeCodes(): number[] {
-    return this.formState.remoteUserCertifications.value
+    return this.remoteUserCertifications.value
       .map((certification: RemoteUserCertification) => +certification.collegeCode);
```
johnnyl6, line 85: Just curious, any particular reason to use unary operator + here to convert to `number` since `collegeCode` is already `number` type? Maybe should guard `collegeCode` field as `readonly` in `RemoteUserCertification` since it's immutable?

johnnyl6 (approved):
LGTM🎉

AnaisHebert (commented):
When I add a remote user if i press continue i get this 
image
This will be addressed in a different PR

AnaisHebert (approved):
☕

PR #1206, opened 2021-03-27
Util Tests

mtpultz:
* Set of tests for utils


prime-angular-frontend/src/app/lib/utils/string-utils.class.ts
```diff
@@ -1,8 +1,7 @@
 export class StringUtils {
   public static capitalize(value: string): string {
-    const strValue: String = new String(value);
-    return (value)
-      ? `${strValue.charAt(0).toUpperCase()}${strValue.slice(1).toLowerCase()}`
+    return (value && typeof value === 'string')
```
mtpultz, line 3: @johnnyl6 this is a good example of where making tests pass doesn't quite work.  The root of the issue was the type provided was a number not a string.  See the `claim-note.component.spec.ts` where the type is set to 1, but the enum used for type is a string enum that should have been `ClaimType.ENROLLEE`.  So I reverted the change that uses String which is a linting error, but also added a type guard to protect against types that are not strings that get passed into the pipe.

johnnyl6: Awesome! I should have traced back all reference usages of `StringUtils.capitalize`. Great findings.👏

johnnyl6 (approved):
🎉

PR #1209, opened 2021-03-30
PRIME-1459 Enter site info for each Health Authority

neophyte57:
As per Jira ticket PRIME-1459 (and see comments), each selected health authority needs address information.  Job title is also moved to the care setting/health authority-level, rather than at the Enrollee level.

Another change as a result of these changes is that the Admin Enrollee Overview screen required adjustments as well.  

There is an outstanding issue re: migrating data in production out of the `Job` table.


prime-angular-frontend/src/app/modules/enrolment/pages/job/job.component.ts
```diff
@@ -242,7 +212,9 @@ export class JobComponent extends BaseEnrolmentProfilePage implements OnInit, On
     this.oboSites.clear();
```
johnnyl6, line 212: Got `Cannot read property 'clear' of null ` error on regulatory page when hitting 'Save and continue" button.

Testing case: new enrolment, selected more than 1 HA in care setting,
image

AnaisHebert (commented):
image
I got this error after starting out with two health authorities, navigating to the overview page, then going back to care setting and adding a 3rd health authority. The job page then had this error

prime-dotnet-webapi/Models/OboSite.cs
```diff
@@ -20,12 +20,20 @@ public class OboSite : BaseAuditable, IEnrolleeNavigationProperty
...
+        public HealthAuthority HealthAuthority { get; set; }
+
         public string SiteName { get; set; }
 
         public string PEC { get; set; }
 
         public string FacilityName { get; set; }
 
+        [Required]
+        public string JobTitle { get; set; }
```
AnaisHebert, line 35: If we are adding jobs to the oboSites, then we should probably remove the jobs Collection on Enrollee and migrate all jobs to exist within an oboSite

neophyte57: I agree re: "remove the jobs Collection on Enrollee" and at some point there should be a Migration to drop the `Job` table.  

Related, the previous arrangement of associating jobs at the enrollee-level makes it difficult to know which Obo Site existing Jobs (in production) should be associated with, *if* an enrollee has more than one Care Setting/Health Authority.

AnaisHebert: I think before this can be merged in we will have to come up with a solution for the migration. Agreed though it will be a tricky one, will need some discussion

neophyte57:
> image
> I got this error after starting out with two health authorities, navigating to the overview page, then going back to care setting and adding a 3rd health authority. The job page then had this error

This should be fixed and generally I did more testing with related pages and addressed failures.

prime-dotnet-webapi/Models/OboSite.cs
neophyte57: The proposed data migration is in the comments of the Jira ticket:  https://bcgovmoh.atlassian.net/browse/PRIME-1459

prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts
```diff
@@ -168,14 +170,39 @@ export class EnrolmentFormStateService extends AbstractFormStateService<Enrolmen
...
+            break;
+          }
+        }
+        // this.logger.trace(`Checkbox at ${i} -> result of ${foundMatchingHAOboSite}`);
+        if (!foundMatchingHAOboSite) {
+          hasOboSiteForEveryHA = false;
+          break;
+        }
+      } else {
+        // this.logger.trace(`Checkbox at ${i} was unchecked`);
```
AnaisHebert, line 200: Can we remove these commented out lines?

neophyte57: Yes, done.

neophyte57:
@james-hollinger, @AnaisHebert data migration added to PR

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
```diff
@@ -248,7 +255,13 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
         return clear(form.get('communityPharmacySites') as FormArray);
       }
       case CareSettingEnum.HEALTH_AUTHORITY: {
-        return clear(form.get('healthAuthoritySites') as FormArray);
+        const healthAuthoritySites = form.get('healthAuthoritySites') as FormGroup;
+        Object.keys(healthAuthoritySites.controls).forEach(healthAuthorityCode => {
+          const sitesOfHealthAuthority = healthAuthoritySites.get(String(healthAuthorityCode)) as FormArray;
```
mtpultz, line 260: In the rest of the app using ES6 syntax we're converting to a string using ticks -  `${healthAuthorityCode}`

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
```diff
@@ -190,12 +193,16 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
...
+    const oboSites = this.enrolmentFormStateService.jobsForm.get('oboSites').value as OboSite[];
+    const certifications = this.enrolmentFormStateService.regulatoryFormState.certifications;
 
     let nextRoutePath: string;
     if (!this.isProfileComplete) {
       nextRoutePath = EnrolmentRoutes.REGULATORY;
-    } else if (jobs?.length) {
+    } else if (oboSites?.length) {
+      nextRoutePath = EnrolmentRoutes.JOB;
+    } else if (certifications.length === 0 && oboSites?.length === 0) {
```
mtpultz, line 204: You're using JS language syntax for falsey above with a null/length check on `oboSites?.length`, but then change it to an explicit check on 0.  I know you don't like JS language syntax, but you're using it then not using it, but it would be more consistent if you had: `!certifications?.length && !oboSites?.length`

prime-angular-frontend/src/app/modules/enrolment/pages/health-authority/health-authority.component.ts
```diff
@@ -105,12 +105,12 @@ export class HealthAuthorityComponent extends BaseEnrolmentProfilePage implement
...
 
   protected nextRouteAfterSubmit() {
-    const jobs = this.enrolmentFormStateService.jobsForm.get('jobs').value as Job[];
+    const oboSites = this.enrolmentFormStateService.jobsForm.get('oboSites').value as OboSite[];
 
     let nextRoutePath: string;
     if (!this.isProfileComplete) {
       nextRoutePath = EnrolmentRoutes.REGULATORY;
-    } else if (jobs.length) {
+    } else if (oboSites.length) {
```
mtpultz, line 113: You likely need a null check on this since you're doing it the care-setting.component - `oboSites?.length`

image

prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts
```diff
@@ -168,14 +170,36 @@ export class EnrolmentFormStateService extends AbstractFormStateService<Enrolmen
...
-    const jobs = this.jobsForm.get('jobs') as FormArray;
+    const oboSites = this.jobsForm.get('oboSites') as FormArray;
     const certifications = this.regulatoryFormState.certifications;
+    const enrolleeHealthAuthorities = this.careSettingsForm.get('enrolleeHealthAuthorities') as FormArray;
+    let hasOboSiteForEveryHA = true;
+    // Using `for` loop rather than `every()` method for ease of debugging
+    for (let i = 0; i < enrolleeHealthAuthorities.controls.length; i++) {
+      const checkbox = enrolleeHealthAuthorities.controls[i];
+      // For every selected Health Authority ...
+      if (checkbox.value === true) {
```
mtpultz, line 184: You don't need to check for `true` in JS if it exists its true so `if(checkbox.value) { ... }`

prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts
```diff
@@ -168,14 +170,36 @@ export class EnrolmentFormStateService extends AbstractFormStateService<Enrolmen
...
+        let foundMatchingHAOboSite = false;
+        // ... there must be at least one Obo Site for that Health Authority
+        for (let j = 0; j < oboSites.controls.length; j++) {
```
mtpultz, line 187: If you're checking for at least one you don't need a `for` loop since JS is functional it has a bunch of methods for doing checks for a single and exit - `oboSites.controls.some(predicateFunction)`.  It will exit with true if your predicate passes at any point and immediately exit, otherwise false.  I've added this to the example above.

prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts
```diff
@@ -168,14 +170,36 @@ export class EnrolmentFormStateService extends AbstractFormStateService<Enrolmen
...
+   * Check for the requirement of at least one certification, or one obo site/job.
    */
   public hasCertificateOrJob(): boolean {
-    const jobs = this.jobsForm.get('jobs') as FormArray;
+    const oboSites = this.jobsForm.get('oboSites') as FormArray;
     const certifications = this.regulatoryFormState.certifications;
+    const enrolleeHealthAuthorities = this.careSettingsForm.get('enrolleeHealthAuthorities') as FormArray;
+    let hasOboSiteForEveryHA = true;
+    // Using `for` loop rather than `every()` method for ease of debugging
+    for (let i = 0; i < enrolleeHealthAuthorities.controls.length; i++) {
```
mtpultz, line 181: With JS being functional there isn't much need to do full `for` loop syntax and if you're not using `every` use `forEach`.  I would suggest `every` if that is the intent since you have a comment explaining why you're not using `every`, but I don't see how not using it eases debugging.

prime-angular-frontend/src/app/modules/enrolment/shared/components/job-form/job-form.component.ts
```diff
@@ -12,6 +12,9 @@ import { Job } from '@enrolment/shared/models/job.model';
   templateUrl: './job-form.component.html',
   styleUrls: ['./job-form.component.scss']
 })
+/**
+ * @deprecated Due to moving Job Title to Obo Site (PRIME-1459)
```
mtpultz, line 16: If this is deprecated and unused we can drop the code and component

prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts
```diff
@@ -589,6 +592,49 @@ export class EnrolmentFormStateService extends AbstractFormStateService<Enrolmen
...
+
+  public removeUnselectedHAOboSites() {
+    // Obo Sites need to be removed from two different collections
+    const oboSites = this.jobsForm.get('oboSites') as FormArray;
+    const healthAuthoritySites = this.jobsForm.get('healthAuthoritySites') as FormGroup;
+    const enrolleeHealthAuthorities = this.careSettingsForm.get('enrolleeHealthAuthorities') as FormArray;
+    // If the checkbox for the health authority is not selected, remove the corresponding Obo Sites
+    this.configService.healthAuthorities.forEach((healthAuthority, index) => {
+      if (!enrolleeHealthAuthorities.at(index).value) {
+        for (let i = oboSites.controls.length - 1; i >= 0; i--) {
```
mtpultz, line 627: With JS being functional there isn't much need to do full `for` loop syntax and instead you can use `forEach` - `oboSites.controls.forEach((os: FormGroup) => { ... });`

prime-angular-frontend/src/app/modules/enrolment/shared/components/obo-site-form/obo-site-form.component.html
```diff
@@ -24,14 +24,33 @@
 
           <mat-form-field class="w-100">
             <input matInput
-                   placeholder="Health Authority Facility Name"
+                   placeholder="'{{healthAuthorityName}}' Health Authority Facility Name"
```
mtpultz, line 27: Is there a reason why we're surrounding this with single quotes?  If it is to emphasize the health authority name it doesn't do that visually.

image

If you want to emphasis the health authority name you'll want to add a `mat-label`:

```
<mat-label> 
  <strong>{{ healthAuthorityName }}</strong> Health Authority Facility Name
</mat-label>
```

prime-angular-frontend/src/app/modules/enrolment/shared/components/obo-site-form/obo-site-form.component.html
```diff
@@ -24,14 +24,33 @@
...
                    formControlName="facilityName">
             <mat-error>Required</mat-error>
           </mat-form-field>
 
         </div>
       </div>
 
+      <mat-form-field class="w-100">
```
mtpultz, line 35: Needs a bit of vertical spacing otherwise it's sitting right on top of the next field:

image

The spacing should match the kind of spacing seen when the error message appears so the page doesn't move when the validation message appears:

image

prime-dotnet-webapi/Migrations/20210329212828_HALikeCareSettingWithJobTitle.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            migrationBuilder.AddForeignKey(
+                name: "FK_OboSite_HealthAuthorityLookup_HealthAuthorityCode",
+                table: "OboSite",
+                column: "HealthAuthorityCode",
+                principalTable: "HealthAuthorityLookup",
+                principalColumn: "Code",
+                onDelete: ReferentialAction.Restrict);
+
+            // Update the Job Titles of all Obo Sites of an Enrollee
```
AnaisHebert, line 33: Migrations look good to me, tested.

AnaisHebert: Actually, I dont think you should drop the Job table here, or if you do then you also have to remove Job from everywhere in the backend code. It fails to load enrollees after running this migration.

neophyte57: Thanks @AnaisHebert for checking.  I'll remove the Job table deletion.  Maybe removing Job "artifacts" can be a clean-up ticket next sprint?

mtpultz: Clean up of artifacts can definitely occur next sprint.

prime-angular-frontend/src/app/modules/enrolment/shared/components/job-form/job-form.component.ts
mtpultz: @neophyte57 I've made a note of this so we can deal with it next sprint with other job related clean up tasks.

prime-angular-frontend/src/app/modules/enrolment/shared/components/obo-site-form/obo-site-form.component.html
neophyte57: Addressed I believe

prime-dotnet-webapi/Migrations/20210329212828_HALikeCareSettingWithJobTitle.cs
neophyte57: Addressed

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
neophyte57: Addressed

prime-angular-frontend/src/app/modules/enrolment/pages/health-authority/health-authority.component.ts
neophyte57: Addressed

james-hollinger:
If we are no longer updating the Job collection on Enrollee directly, We should remove it from the EnrolleeUpdateModel

neophyte57:
> If we are no longer updating the Job collection on Enrollee directly, We should remove it from the EnrolleeUpdateModel

I believe next sprint, we'll have a Jira ticket to clean up Job table from database and other Job-related references ...

AnaisHebert (approved):
lgtm 🌞

prime-angular-frontend/src/app/modules/enrolment/shared/components/obo-site-form/obo-site-form.component.html
mtpultz: Updated

prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts
mtpultz: I'll make these updates quickly in a separate PR

prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts
mtpultz: I'll make these updates quickly in a separate PR

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
```diff
@@ -190,12 +193,16 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
...
   protected nextRouteAfterSubmit() {
-    const jobs = this.enrolmentFormStateService.jobsForm.get('jobs').value as Job[];
+    const oboSites = this.enrolmentFormStateService.jobsForm.get('oboSites').value as OboSite[];
+    const certifications = this.enrolmentFormStateService.regulatoryFormState.certifications;
 
     let nextRoutePath: string;
     if (!this.isProfileComplete) {
       nextRoutePath = EnrolmentRoutes.REGULATORY;
-    } else if (jobs?.length) {
+    } else if (oboSites?.length) {
```
mtpultz, line 202: I can make this update quickly in a separate PR

mtpultz (approved):

PR #1211, opened 2021-04-01
PRIME-1430 - Remove old roles from Keycloak

james-hollinger:
Old roles are now gone from Keycloak in all environments


mtpultz (approved):
Grayson_Wow

PR #1212, opened 2021-04-01
PRIME-1452 - Remove DOB and Address from eForms JWT

james-hollinger:
Created new client scope for eForms that does not include date of birth, and removed address scope.


mtpultz (approved):
Grayson_Wow

PR #1215, opened 2021-04-07
Two Realms

james-hollinger:
Sets up the backend for integration into both our own Keycloak realm and also the MOH's Keycloak.

Environment variables have been added to Dev only:
KEYCLOAK_REALM_URL
MOH_KEYCLOAK_REALM_URL


mtpultz (approved):
Looks good to me image

PR #1219, opened 2021-04-19
Remove MOH Keycloak from default Policy

james-hollinger:
The MOH Keycloak integration was preventing login due to it unexpectedly still running even after the PRIME Keycloak had already validated the token. This was compounded by the fact that the MOH Keycloak is behind a firewall, so would return 403 FORBIDDEN when the app tried to validate the token.

MOH Keycloak has been removed from the default policy. To use it, you can specify it in the Authorize attribute on a method/controller:

```C#
[Authorize(AuthenticationSchemes = Schemes.MohJwt)]
```

Note that, due to the PRIME scheme being part of the default policy, this example will _also_ run the PRIME JWT Scheme. If this becomes a problem, we may need to set up some explicit policies.


mtpultz (approved):
94213729-7e88ea00-fe8c-11ea-8194-de8895548df1

hogmann (approved):
🐶 🎈

PR #1230, opened 2021-04-23
PRIME-1439 HA Site Registration changes

johnnyl6:
- Authorized User changes:
   remove fax field
   add health authority drop down
- Site Information changes:
   Site ID toggle to behavior change
- PharmaNet Administrator
  remove fax field
  add button: add PharmaNet Administrator
  reverse toggle to function like this:

  change how this toggle behaves for all personas(privacy officer and technical support contact) as well as in all other site types (community practice and community pharmacy)

- Privacy Officer
  remove fax field
- technical support contact
  remove fax field


prime-angular-frontend/src/app/shared/components/forms/address-form/address-form.component.html
```diff
@@ -12,7 +12,7 @@
 </app-address-form-autocomplete>
 
 <div [formGroup]="form"
-     [hidden]="!showAddressFields">
+     [hidden]="!showAddressFields && form.untouched">
```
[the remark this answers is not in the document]
johnnyl6: > Feels like if it's possible to avoid adding this we should. I would have thought that if you wanted the form field to be displayed you would set `showAddressFields = true` so whether it is displayed or not is independent of form state.

When the form is hidden and user click "save & continue", the validations on the form wont show. Adding form.untouched is quick hack.

prime-angular-frontend/src/app/modules/site-registration/pages/technical-support-page/technical-support-page.component.ts
```diff
@@ -55,6 +55,8 @@ export class TechnicalSupportPageComponent extends AbstractEnrolmentPage impleme
       contact.physicalAddress = new Address();
     }
     this.formState.form.patchValue(contact);
+    // mark the form touched so address form can expand
+    this.formState.form.markAllAsTouched();
```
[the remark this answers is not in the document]
johnnyl6: > Feels like we could pass in a binding to `ContactProfileForm` that sets `showAddress` for the physical address on the profile form for each of these contact pages.

This is related to the above form.untouched fix. After patching and markAsTouched, the hidden address form will show otherwise it remains hidden.

prime-angular-frontend/src/app/shared/components/forms/address-form/address-form.component.html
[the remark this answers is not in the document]
mtpultz: I realize what it does, but the reason this looks odd to me is due to AddressForm being a base component.  By doing this each component throughout the application is affected with hopefully no side-effects, and in order to consume the component you need to know that the state can control visibility not your choice of bindings.

prime-angular-frontend/src/app/modules/site-registration/pages/technical-support-page/technical-support-page.component.ts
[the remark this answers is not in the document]
mtpultz: I realize the relationship it was the reason I pointed this out since it emphasizes my initial point where you need to know that the you have to mark the state as touched in order to affect a component 2 levels below the current component instead of using bindings.  I'm trying to illustrate that even though it works it isn't necessarily intuitive or a good idea for a base component, which could also have application wide side-effects.

prime-angular-frontend/src/app/shared/components/forms/site-information-form/site-information-form.component.html
```diff
@@ -32,27 +32,35 @@
...
+    <div class="mb-3">
+      <mat-slide-toggle [formControl]="isNewSite"
+                        color="primary">
+        {{ this.isNewSite.value ? 'This is a new Site without a Site ID' :
+        'This is
+        not a new Site and it has a PEC Code
+        assigned' }}
+      </mat-slide-toggle>
+    </div>
+    <ng-container *ngIf="this.isNewSite.value; else notNewSite">
```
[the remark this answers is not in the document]
mtpultz: I made this update

prime-angular-frontend/src/app/modules/site-registration/pages/administrator-page/administrator-page.component.ts
```diff
@@ -45,6 +48,7 @@ export class AdministratorPageComponent extends AbstractEnrolmentPage implements
 
     this.title = route.snapshot.data.title;
     this.routeUtils = new RouteUtils(route, router, SiteRoutes.MODULE_PATH);
+    this.formSubmittingEvent = new Subject<void>();
```
mtpultz, line 51: Is there a reason to have this as a `Subject`?  With a form submission you can't unring the bell once you submit.  Feels like this could be `formSubmitted: boolean` since `Input` bindings are reactive which would be defaulted to false and once changed can be picked up in the `ngOnChange` lifecycle hook if you need to do perform tasks on the change.

prime-angular-frontend/src/app/shared/components/site/contact-profile-form/contact-profile-form.component.html
```diff
@@ -15,9 +15,43 @@
       </ng-container>
     </app-page-subheader2>
 
-    <ng-content select="app-same-as"></ng-content>
-
-    <div class="row">
+    <ul class="list-unstyled mb-4 ml-3">
```
mtpultz, line 18: Was this contact card supposed to be implemented in this ticket?  They are still making decisions around contacts, and site registration and health authority site registrations won't be the same.

johnnyl6: yes, confirmed with Pilar. She did not add mockup screen to the ticket though.. Will add the mockup to ticket too.

image

mtpultz: @johnnyl6 please don't make changes to the tickets acceptance criteria mid sprint even with Pilar.  This was a 1 point ticket, and if I had known I would have let you know that we weren't planning on maintaining this as a shared component between registrations

prime-angular-frontend/src/app/modules/site-registration/pages/administrator-page/administrator-page.component.ts
mtpultz: `ngOnChange` gets invoked, but it has a `SimpleChanges` param that shows whether a value has changed.  You don't have to worry about it being a performance issue since it gets invoked whether you use it or not with it being a lifecycle hook.

mtpultz: By using an observable you're saying that subsequent events might change the value, but once you've submitted once it should remain true forever so the observable isn't being used as you would expect where data changes overtime.

johnnyl6: It's a Subject holding void value and the subscriber contact-form component only cares about the form submitting event triggering. This is an alternative of onSubmit hook for child component, not meant to pass true/false value downstream. Will try out the ngOnChange hook approach.

prime-angular-frontend/src/app/shared/components/site/contact-profile-form/contact-profile-form.component.html
mtpultz: We'll break this out into just HA in the next ticket

mtpultz (approved):

PR #1244, opened 2021-04-30
Fix Site registration selenium test

johnnyl6:
- more robust wait for page to load complete
- increased timeout
- must call driver.quit() on teardown to avoid memory leak
- improved and reliable logic on workflow check
- removed  calls to obsolete code (added reference to SeleniumExtras.WaitHelpers


mtpultz (approved):
image

PR #1246, opened 2021-04-30
Added null check to site reg overview container

mtpultz:
Fixes and issue where manually inserted sites won't have business licences, which throws an error - enter solution = null check


johnnyl6 (approved):
👍🎉✨

PR #1248, opened 2021-05-04
PRIME-1533 Enable Editing on Site Registration

johnnyl6:


james-hollinger:
The JIRA ticket mentioned "Avoid adding state". We had intended that enabling editing would simply clear the submitted date on the site.

johnnyl6:
> The JIRA ticket mentioned "Avoid adding state". We had intended that enabling editing would simply clear the submitted date on the site.

Thanks James. Missed the assumption at the bottom😅 Now removed the added state.

prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -803,6 +803,29 @@ public async Task<ActionResult<Site>> DeclineSite(int siteId)
...
+        /// <summary>
+        /// Enable editing a site
+        /// </summary>
+        /// <param name="siteId"></param>
+        [HttpPut("{siteId}/enable-editing", Name = nameof(EnableEditingSite))]
+        [Authorize(Roles = Roles.EditSite)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
```
james-hollinger, line 816: This method returns 200 OK, but this says 204 No Content. Do you need to return the site data? Seems reasonable to change line 826 to `NoContent()`.

prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -803,6 +803,29 @@ public async Task<ActionResult<Site>> DeclineSite(int siteId)
...
+        public async Task<ActionResult<Site>> EnableEditingSite(int siteId)
```
james-hollinger, line 817: I've been pushing to slowly stop using the typed versions of `ActionResult` unless we actually use it. I'd prefer if this was `public async Task<ActionResult> EnableEditingSite(int siteId)`

prime-dotnet-webapi/Services/SiteService.cs
```diff
@@ -350,6 +350,17 @@ public async Task<Site> DeclineSite(int siteId)
             return site;
         }
 
+        public async Task<Site> EnableEditingSite(int siteId)
+        {
+            var site = await _context.Sites.SingleOrDefaultAsync(s => s.Id == siteId);
+            site.SubmittedDate = null; ;
```
james-hollinger, line 356: two semicolons

prime-dotnet-webapi/Services/SiteService.cs
```diff
@@ -350,6 +350,17 @@ public async Task<Site> DeclineSite(int siteId)
...
+            await _context.SaveChangesAsync();
+
+            await _businessEventService.CreateSiteEventAsync(site.Id, site.Organization.SigningAuthorityId, "Site Enabled Editing");
```
james-hollinger, line 359: I don't feel this is a site creation event, it just seems to be the only business event we have for sites.
I would prefer seeing a new business event for sites; either a generic one or one for status changes. The Approve endpoint should also trigger this new business event.

prime-dotnet-webapi/Controllers/SitesController.cs
johnnyl6: The returned updatedSite can be used to reflect the submission date changed to empty in the site registration mat-table.  I think it's more clear to use the returned site from webapi rather than client side code to modify the submission date, hence business logic only resides on server side. Will change 204 to 200.

prime-dotnet-webapi/Services/SiteService.cs
johnnyl6: I thought the `CreateSiteEventAsync` does not mean site creation but to 'create' a site event itself. The event description is passed in as parameter "Site Enabled Editing", isn't it?

prime-angular-frontend/src/app/core/resources/site-resource.service.ts
```diff
@@ -385,9 +385,22 @@ export class SiteResource {
       );
   }
 
+  public enableEditingSite(siteId: number): Observable<Site> {
+    return this.apiResource.put<Site>(`sites/${siteId}/enable-editing`)
+      .pipe(
+        map((response: ApiHttpResponse<Site>) => response.result),
+        tap(() => this.toastService.openSuccessToast('Site registration has been enabled editing')),
```
itchison, line 392: I think all of these messages should have the format of "Site registration editing has been enabled" or "Site registration editing could not be enabled".

prime-dotnet-webapi/Services/SiteService.cs
james-hollinger: I think you are right, my mistake

prime-angular-frontend/src/app/core/resources/site-resource.service.ts
johnnyl6: Fixed in 5973e27

james-hollinger:
It looks like the state breaks when the user re-submits the site after editing is enabled: the status stays as "editable" despite the submitted date being filled out once again.

johnnyl6:
> It looks like the state breaks when the user re-submits the site after editing is enabled: the status stays as "editable" despite the submitted date being filled out once again.

'editable' has been removed in 411076f so only the original approved, declined and under_review status exist now. Are you running localhost or the PR branch on server (which might not reflect the changes)?

james-hollinger:
> 
> 
> > It looks like the state breaks when the user re-submits the site after editing is enabled: the status stays as "editable" despite the submitted date being filled out once again.
> 
> 'editable' has been removed in [411076f](https://github.com/bcgov/moh-prime/commit/411076fdddd990881b7e38739a1282aa7012e743) so only the original approved, declined and under_review status exist now. Are you running localhost or the PR branch on server (which might not reflect the changes)?

On the PR branch:
image

Editable has not been removed from the frontend.

johnnyl6:
> > > It looks like the state breaks when the user re-submits the site after editing is enabled: the status stays as "editable" despite the submitted date being filled out once again.
> > 
> > 
> > 'editable' has been removed in [411076f](https://github.com/bcgov/moh-prime/commit/411076fdddd990881b7e38739a1282aa7012e743) so only the original approved, declined and under_review status exist now. Are you running localhost or the PR branch on server (which might not reflect the changes)?
> 
> On the PR branch:
> image
> 
> Editable has not been removed from the frontend, looks like your last changes didn't get deployed?

Right. It wasn't deployed until  11:00am today.

prime-dotnet-webapi/Controllers/SitesController.cs
mtpultz: @johnnyl6 PR looks good if we can get this addressed we can get it merged in

johnnyl6: done in da81eee

prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -803,6 +803,29 @@ public async Task<ActionResult<Site>> DeclineSite(int siteId)
...
+        /// <summary>
+        /// Enable editing a site
+        /// </summary>
+        /// <param name="siteId"></param>
+        [HttpPut("{siteId}/enable-editing", Name = nameof(EnableEditingSite))]
+        [Authorize(Roles = Roles.EditSite)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status200OK)]
```
james-hollinger, line 816: Now that you have updated this to 200OK from 204NoContent, this should have the return type of the endpoint:
```C#
[ProducesResponseType(typeof(ApiResultResponse<Site>), StatusCodes.Status200OK)]
```

johnnyl6: fixed in 53ba7fd

PR #1249, opened 2021-05-04
PRIME-1545 Need to hide Health Authority site registration in production.

AnaisHebert:


prime-angular-frontend/src/app/app-config.module.ts
```diff
@@ -24,6 +24,7 @@ export class AppConfig {
   phoneNumbers: {
     director: string;
   };
+  env: string;
```
mtpultz, line 27: I would probably make this a union type of `'prod' | 'test' | 'dev'` instead of a string so no one can mistakenly type `production`, `develop`, or just have a typo.

mtpultz (approved):
image

PR #1252, opened 2021-05-06
PRIME-1512 - Cancel TOA Assignment

james-hollinger:
Ability for adjudicators to undo a miss-assigned TOA type, usually (but not exclusively) after manual adjudication.

Slightly waiting on if additional enrollee notification is required.


mtpultz:
👀

mtpultz (approved):
image

PR #1264, opened 2021-05-11
PRIME-1472 Health Authority admin tabs/filtering + HA Grid

johnnyl6:
added tab filter by careSetting for site registration admin


hogmann:
So this ticket wasn't caught early enough but this is in conflict with the work Marty and I worked on with the Authorized user work. Since the tabs code was in the work I did that wasn't used a few sprints ago so we just used that code. Marty is writing a note about it on the ticket.

PR #1266, opened 2021-05-11
PRIME-1540 Authorized User Application/Approval

mtpultz:
This includes PRIME-1540 and PRIME-1541.

Login for health authority site registration is at `/health-authority`


johnnyl6:
Got an error in FE test.

```
ERROR in src/app/modules/health-auth-site-reg/pages/authorized-user-page/authorized-user-page.component.spec.ts:79:63 - error TS2339: Property 'authorizedUserFormState' does not exist on type 'HealthAuthSiteRegFormStateService'.

79       component.formState = healthAuthSiteRegFormStateService.authorizedUserFormState;
```

johnnyl6:
I think one of the requirements in PRIME-1472 Health Authority admin tabs/filtering + HA Grid is to have different columns for PCHP and CP.
As per requirements  in PRIME-1472
1. Since the tab already indicates careSetting, "Care Setting' column is removed 
2. Column "Missing Business Licence' is also redundant and removed for PCHP
3. Column 'Remote Users" is redundant and removed for CP.

mtpultz:
> Got an error in FE test.
> 
> ```
> ERROR in src/app/modules/health-auth-site-reg/pages/authorized-user-page/authorized-user-page.component.spec.ts:79:63 - error TS2339: Property 'authorizedUserFormState' does not exist on type 'HealthAuthSiteRegFormStateService'.
> 
> 79       component.formState = healthAuthSiteRegFormStateService.authorizedUserFormState;
> ```

Nice catch.  Had to remove authorized user from the FormStateService.  Will push an update in a bit

hogmann:
> Got an error in FE test.
> 
> ```
> ERROR in src/app/modules/health-auth-site-reg/pages/authorized-user-page/authorized-user-page.component.spec.ts:79:63 - error TS2339: Property 'authorizedUserFormState' does not exist on type 'HealthAuthSiteRegFormStateService'.
> 
> 79       component.formState = healthAuthSiteRegFormStateService.authorizedUserFormState;
> ```



> I think one of the requirements in PRIME-1472 Health Authority admin tabs/filtering + HA Grid is to have different columns for PCHP and CP.
> As per requirements in PRIME-1472
> 
> 1. Since the tab already indicates careSetting, "Care Setting' column is removed
> 2. Column "Missing Business Licence' is also redundant and removed for PCHP
> 3. Column 'Remote Users" is redundant and removed for CP.

prime-dotnet-webapi/Controllers/AuthorizedUsersController.cs
```diff
@@ -0,0 +1,185 @@
...
+            _organizationService = organizationService;
+        }
+
+        // GET: api/parties/authorized-users/5fdd17a6-1797-47a4-97b7-5b27949dd614
+        /// <summary>
+        /// Gets a AuthorizedUser by user ID.
+        /// </summary>
+        /// <param name="userId"></param>
+        [HttpGet("{userId:guid}", Name = nameof(GetAuthorizedUserByUserId))]
+        [ProducesResponseType(typeof(ApiBadRequestResponse), StatusCodes.Status400BadRequest)]
```
james-hollinger, line 41: This does not return a BadRequest() anywhere, so we can remove this

prime-dotnet-webapi/Controllers/AuthorizedUsersController.cs
```diff
@@ -0,0 +1,185 @@
...
+            return Ok(ApiResponse.Result(authorizedUser));
+        }
+
+        // GET: api/parties/authorized-users/5
+        /// <summary>
+        /// Gets a specific AuthorizedUser.
+        /// </summary>
+        /// <param name="authorizedUserId"></param>
+        [HttpGet("{authorizedUserId:int}", Name = nameof(GetAuthorizedUserById))]
+        [ProducesResponseType(typeof(ApiBadRequestResponse), StatusCodes.Status400BadRequest)]
```
james-hollinger, line 63: See above

prime-dotnet-webapi/Controllers/AuthorizedUsersController.cs
```diff
@@ -0,0 +1,181 @@
...
+        // GET: api/parties/authorized-users/5fdd17a6-1797-47a4-97b7-5b27949dd614
+        /// <summary>
+        /// Gets a AuthorizedUser by user ID.
+        /// </summary>
+        /// <param name="userId"></param>
+        [HttpGet("{userId:guid}", Name = nameof(GetAuthorizedUserByUserId))]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(typeof(ApiResultResponse<AuthorizedUserChangeModel>), StatusCodes.Status200OK)]
```
james-hollinger, line 44: I assume this is the wrong ViewModel?

prime-dotnet-webapi/Controllers/AuthorizedUsersController.cs
```diff
@@ -0,0 +1,181 @@
...
+        // GET: api/parties/authorized-users/5
+        /// <summary>
+        /// Gets a specific AuthorizedUser.
+        /// </summary>
+        /// <param name="authorizedUserId"></param>
+        [HttpGet("{authorizedUserId:int}", Name = nameof(GetAuthorizedUserById))]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(typeof(ApiResultResponse<AuthorizedUserChangeModel>), StatusCodes.Status200OK)]
```
james-hollinger, line 65: as above

AnaisHebert (approved):
🥓🥓🥓

prime-dotnet-webapi/Controllers/AuthorizedUsersController.cs
james-hollinger: Shouldn't this be
```C#
 [ProducesResponseType(typeof(ApiResultResponse<AuthorizedUserViewModel>), StatusCodes.Status200OK)]
 ```

Comment (james-hollinger) [reply prime-dotnet-webapi/Controllers/AuthorizedUsersController.cs:45]: Shouldn't this be
```C#
 [ProducesResponseType(typeof(ApiResultResponse<AuthorizedUserViewModel>), StatusCodes.Status200OK)]
 ```

File: prime-dotnet-webapi/Controllers/AuthorizedUsersController.cs [hunk 5]
```diff
@@ -0,0 +1,198 @@
...
+using Prime.ViewModels;
+using Prime.ViewModels.Parties;
+
+namespace Prime.Controllers
+{
+    [Produces("application/json")]
+    [Route("api/parties/authorized-users")]
+    [ApiController]
+    [Authorize(Roles = Roles.PrimeEnrollee + "," + Roles.ViewSite)]
+    public class AuthorizedUsersController : ControllerBase
```
james-hollinger, line 23: None of the endpoints are checking that the user has permission to access this Party (by UserId)

prime-dotnet-webapi/Controllers/AuthorizedUsersController.cs
```diff
@@ -0,0 +1,198 @@
...
+            return NoContent();
+        }
+
+        // POST: api/parties/authorized-user/5/approve
+        /// <summary>
+        /// Approves the authorized user.
+        /// </summary>
+        /// <param name="authorizedUserId"></param>
+        [HttpPost("{authorizedUserId}/approve", Name = nameof(ApproveAuthorizedUser))]
+        [Authorize(Roles = Roles.ViewSite)]
```
james-hollinger, line 180: This should probably be EditSite

prime-dotnet-webapi/Services/AuthorizedUserService.cs
```diff
@@ -0,0 +1,137 @@
...
+            var authorizedUser = await GetBaseAuthorizedUserQuery()
+                .SingleOrDefaultAsync(au => au.Party.UserId == user.GetPrimeUserId());
+
+            if (authorizedUser == null)
+            {
+                var party = new Party
+                {
+                    Addresses = new List<PartyAddress>()
+                };
+                _context.Parties.Add(party);
```
james-hollinger, line 71: This will throw if the User already has a Party in the database but isn't already an AuthorizedUser. This will try to add another Party to the DB with a duplicate UserId

james-hollinger: As an aside: if you are creating multiple entities at the same time, you shouldn't need to add them to all the DB sets individually; EF tracks the whole object graph. This block of code could be simplified to 
```C#
authorizedUser = new AuthorizedUser
{
    Party = new Party
    {
        Addresses = new List<PartyAddress>()
    },
    Status = AccessStatusType.UnderReview
};
_context.AuthorizedUsers.Add(authorizedUser);
```

prime-dotnet-webapi/Services/HealthAuthorityService.cs
```diff
@@ -0,0 +1,42 @@
...
+                .ProjectTo<AuthorizedUserViewModel>(_mapper.ConfigurationProvider)
+                .ToListAsync();
+        }
+
+        public async Task<IEnumerable<HealthAuthorityCode>> GetHealthAuthorityCodesWithUnderReviewAuthorizedUsersAsync()
+        {
+            var underReviewUsers = await _context.AuthorizedUsers
+                .Where(u => u.Status == AccessStatusType.UnderReview).ToListAsync();
+
+            return underReviewUsers.Select(u => u.HealthAuthorityCode).ToList().Distinct();
```
james-hollinger, line 39: This Select() can be done in the DB:
```C#
var codes = await _context.AuthorizedUsers
    .Where(u => u.Status == AccessStatusType.UnderReview)
    .Select(u => u.HealthAuthorityCode)
    .ToListAsync();

return codes.Distinct();
```

james-hollinger:
image
Would be way funnier this way:
image

prime-dotnet-webapi/Services/HealthAuthorityService.cs
```diff
@@ -0,0 +1,42 @@
...
+            : base(context, httpContext)
+        {
+            _mapper = mapper;
+        }
+
+        public async Task<IEnumerable<AuthorizedUserViewModel>> GetAuthorizedUsersByHealthAuthorityAsync(HealthAuthorityCode code)
+        {
+            return await _context.AuthorizedUsers
+                .Where(u => u.HealthAuthorityCode == code)
+                .ProjectTo<AuthorizedUserViewModel>(_mapper.ConfigurationProvider)
```
james-hollinger, line 30: Going by whats in the FE:
image

This viewmodel only needs Name, Job title, Id and Status. I assume this method is re-using the AuthorizedUserViewModel to keep it really small because we expect that we are removing this controller and service in the near future? If that's the case, could we add a comment to that affect somewhere to avoid confusion? (Also, this wont populate the Addresses because they are added in Includes() rather than the mapping configuration)

prime-dotnet-webapi/ViewModels/AutoMapping.cs
```diff
@@ -74,5 +76,21 @@ public AutoMapping()
...
+            .ForMember(dest => dest.DateOfBirth, opt => opt.MapFrom(src => src.Party.DateOfBirth))
+            .ForMember(dest => dest.PreferredFirstName, opt => opt.MapFrom(src => src.Party.PreferredFirstName))
+            .ForMember(dest => dest.PreferredMiddleName, opt => opt.MapFrom(src => src.Party.PreferredMiddleName))
+            .ForMember(dest => dest.PreferredLastName, opt => opt.MapFrom(src => src.Party.PreferredLastName))
+            .ForMember(dest => dest.VerifiedAddress, opt => opt.MapFrom(src => src.Party.VerifiedAddress))
+            .ForMember(dest => dest.PhysicalAddress, opt => opt.MapFrom(src => src.Party.PhysicalAddress))
+            .ForMember(dest => dest.Email, opt => opt.MapFrom(src => src.Party.Email))
+            .ForMember(dest => dest.Phone, opt => opt.MapFrom(src => src.Party.Phone))
+            .ForMember(dest => dest.SmsPhone, opt => opt.MapFrom(src => src.Party.SMSPhone))
+            .ForMember(dest => dest.JobRoleTitle, opt => opt.MapFrom(src => src.Party.JobRoleTitle));
```
james-hollinger, line 94: Just in case you weren't aware, the default way to do this kind of mapping is to use naming conventions. AutoMapper will map automatically map src.Party.FirstName to dest.PartyFirstName (for example).

_However_, we want to name our viewmodels like it's a flat model. So I did some digging and I found the proper way to do this:

https://docs.automapper.org/en/stable/Flattening.html#includemembers

Short of it is we make a mapping from Party to AuthorizedUserViewModel (so it knows to match by property names), then we invoke .IncludeMembers() to "hoist up" the child objects and also map them to the destination object:

```C#
CreateMap<AuthorizedUser, AuthorizedUserViewModel>().IncludeMembers(src => src.Party);
CreateMap<Party, AuthorizedUserViewModel>();
```

prime-dotnet-webapi/ViewModels/Parties/AuthorizedUserChangeModel.cs
```diff
@@ -0,0 +1,86 @@
...
+using Prime.Models;
+
+namespace Prime.ViewModels.Parties
+{
+    public class AuthorizedUserChangeModel : IPartyChangeModel
+    {
+        public string PreferredFirstName { get; set; }
+        public string PreferredMiddleName { get; set; }
+        public string PreferredLastName { get; set; }
+        public VerifiedAddress VerifiedAddress { get; set; }
```
james-hollinger, line 11: Thank you for accidentally informing me that we are passing up the `VerifiedAddress` in our other PartyChangeModels rather than mapping or validating against the BC Services Card 😅

Given this is exactly the same way we are doing it in other places, this is _fine for now_ ᵇᵘᵗ ʷᵉ ˢʰᵒᵘˡᵈ ʳᵉᵃˡˡʸ ᶜʰᵃⁿᵍᵉ ᵗʰᶦˢ ᵉᵛᵉʳʸʷʰᵉʳᵉ ᶦⁿ ᵗʰᵉ ᵃᵖᵖ

prime-dotnet-webapi/Services/AuthorizedUserService.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            await _partyService.DeletePartyAsync(authorizedUser.Party.Id);
+        }
+
+        private IQueryable<AuthorizedUser> GetBaseAuthorizedUserQuery()
+        {
+            return _context.AuthorizedUsers
+                .Include(au => au.Party)
+                    .ThenInclude(p => p.Addresses)
+                        .ThenInclude(pa => pa.Address)
```
james-hollinger, line 133: The reason you weren't seeing the Addresses after projecting is because `VerifiedAddress`, `PhysicalAddress`, and `MailingAddress` are not marked with [Computed], so wont be translated into SQL. The `PartyAddress` table is never SELECTed into and so is null. you could avoid these includes by updating `Party`:

```C#
[NotMapped]
[Computed]
public VerifiedAddress VerifiedAddress
{
    get => Addresses
        .Select(a => a.Address)
        .OfType<VerifiedAddress>()
        .SingleOrDefault();
}
```

But this change ould probably cause regressions on other Parties, so its fine for now.

prime-dotnet-webapi/Services/AuthorizedUserService.cs
```diff
@@ -0,0 +1,137 @@
...
+                .Include(au => au.HealthAuthority);
```
james-hollinger, line 134: I don't think `HealthAuthority` is ever used (as opposed to `HealthAuthorityCode`)? I don't think you need this `Include()`.

mtpultz: Yah definitely don't.  It would be used if I hadn't stashed that endpoint for the next iteration of health authorities.  I'll add this to my stash.

prime-dotnet-webapi/Controllers/AuthorizedUsersController.cs
mtpultz: @james-hollinger sure is.  Actually all the endpoints except no content have the wrong model.  I'm amazed as a strongly typed language that this generic response doesn't show up as an error, or prevent the build.

prime-dotnet-webapi/Services/AuthorizedUserService.cs
mtpultz: @james-hollinger I thought mine was a small improvement over the `PartyService` version, but love that it is fully tracked so I updated the below from `PartyService`:

image

to:

image

prime-dotnet-webapi/Services/AuthorizedUserService.cs
mtpultz: @james-hollinger when you're back can you provide recap on automapper.  I have a bunch of scattered pieces on how this works, and had a peak at the docs, but would be good to go over.

prime-dotnet-webapi/ViewModels/AutoMapping.cs
mtpultz: @james-hollinger sweetness!

prime-dotnet-webapi/ViewModels/Parties/AuthorizedUserChangeModel.cs
mtpultz: @james-hollinger this is a reminder comment so we revisit this next sprint - @mtpultz

prime-dotnet-webapi/Services/AuthorizedUserService.cs
mtpultz: @james-hollinger reminder to revisit and find out if there are indeed regressions, and what it would take to update it throughout - @mtpultz

prime-dotnet-webapi/Controllers/AuthorizedUsersController.cs
mtpultz: @james-hollinger guessing your looking for `authorizedUser.Party.PermissionsRecord().AccessableBy(User)`, and looking through it a bit more guess I could also be using `!authorizedUser.Validate(User)`.  Have one question then will update this tomorrow.

prime-dotnet-webapi/Controllers/AuthorizedUsersController.cs
james-hollinger: All this does is generate swagger documentation, there is nothing strongly typed here at all.

prime-dotnet-webapi/Services/AuthorizedUserService.cs
```diff
@@ -0,0 +1,133 @@
...
+            {
+                authorizedUser = new AuthorizedUser
+                {
+                    Party = new Party
+                    {
+                        Addresses = new List<PartyAddress>()
+                    },
+                    Status = AccessStatusType.UnderReview
+                };
+                _context.AuthorizedUsers.Add(authorizedUser);
```
james-hollinger, line 75: While this is now more succinct, it still throws if the User exists in the DB as a Party but not as an Authorized user.

mtpultz: Yep, haven't updated it yet.

PR #1267, opened 2021-05-12
PRIME-1270-confirm-automatic-submissions

AnaisHebert:


mtpultz:
No code comments, but a visual comment.  Feels like we could vertically align the text and icon using flexbox:
image

mtpultz (approved):
image

prime-angular-frontend/src/app/modules/adjudication/shared/components/contextual-enrolment-confirmation/contextual-enrolment-confirmation.component.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  public onEscalate() {
+    const data: DialogOptions = {
+      data: {
+        id: this.enrolleeId,
+        escalationType: EscalationType.ENROLLEE
+      }
+    };
+
+    this.dialog.open(EscalationNoteComponent, { data }).afterClosed()
```
hogmann, line 64: After testing, it doesn't appear hitting the cancel button is properly handled and throws an error in the console.

mtpultz:
@AnaisHebert feels like we should clear the checkboxes when the context closes.  If you open it, check them, and then open escalate and cancel the modal the checkboxes are still checked when reopening the context.

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html
```diff
@@ -15,16 +15,23 @@
...
               *matCellDef="let row;">
             <div class="d-flex flex-wrap triage-container">
+
               <app-triage *ngIf="row.currentStatusCode === EnrolmentStatus.UNDER_REVIEW"
                           class="align-self-start"
                           [enrolleeId]="row.id"
                           [assigned]="(row.adjudicatorIdir)"
                           (reload)="onReload()"></app-triage>
 
+              <app-contextual-enrolment-confirmation *ngIf="row.requiresConfirmation"
```
hogmann, line 25: Im assuming this might be a corner case, but if a user submits and is automatically adjudicated. Then submits again and is under review. Is there a case where requiresConfirmation && EnrolmentStatus = UnderReview?

if so maybe just add "&& row.currentStatusCode !== EnrolmentStatus.UNDER_REVIEW" to this condition so we can cover that case and having to many objects here.

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html
```diff
@@ -142,6 +149,11 @@
                       matTooltip="Enrollee flagged for manual adjudication"
                       matTooltipPosition="after">flag
             </mat-icon>
+            <mat-icon *ngIf="row.confirmed"
```
hogmann, line 152: image
Just a knit picky thing but this icon seems to shift the text down and make it miss-aligned. Im assuming the icon above causes the same thing but might be something worth addressing.

hogmann (commented):
Just the one console error and some knit picking. Everything else LGTM. Added a push with some little fixes.

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html
mtpultz: Hahaha beat you to the nitpicky :)

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html
AnaisHebert: No this shouldnt be possible because you can only get requiresConfirmation if your previous status was RequiresTOA, so if you were Under Review your last status would be editable

PR #1273, opened 2021-05-18
PRIME-1543 - (Start and ) Stop, Banner Time

james-hollinger:
Start/End Date has been expanded with Start/End time for banners.

Postgres does not support DateTimeOffset operations, so the Banners now have DateTimes. We should try using NodaTime when we can, but our current available version of NpgSql / Noda does not allow BCL types and NodaTime types in the same DB Context.


prime-angular-frontend/src/app/shared/models/banner.model.ts
```diff
@@ -1,9 +1,61 @@
...
+    const start = startDate.set({
+      hour: startTime.get('hour'),
+      minute: startTime.get('minute')
+    });
+    const end = endDate.set({
+      hour: endTime.get('hour'),
+      minute: endTime.get('minute')
+    });
+
+    const vm = new BannerViewModel();
```
mtpultz, line 53: This is why I tend to add a constructor so other people don't have to type out the assignments each time they new-up an object, but that's just personal preference since I find assignments like this to be tedious when replicated and can instead use destructuring to pass as params in 2 lines.  In this case what are the chances anyone else is going to be populate a `BannerViewModel` with these keys outside of this factory method... really low... so completely up to you there's no wrong answer.

mtpultz (approved):
Grayson_Wow

johnnyl6:
Might want to adjust layout for date time controls to be more fluid on a lower resolution screen.

On 1920X1080 without zooming there is overlapping. When zoom-out @ 80% it's good.

image

PR #1274, opened 2021-05-18
Feature/prime 1566 view email/toa

hogmann:
Going to touch up some of the front end test files tomorrow as a lot of components added but tests haven't been updated.

PR #1275, opened 2021-05-19
PRIME-1565 admin previewing TOA

johnnyl6:
added ability for admin to preview current enrollee TOA template


prime-dotnet-webapi/Services/AgreementVersionService.cs
```diff
@@ -0,0 +1,47 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.EntityFrameworkCore;
+using Prime.Models;
+using Prime.ViewModels;
+using AutoMapper;
+using AutoMapper.QueryableExtensions;
+
+namespace Prime.Services
+{
+    public class AgreementVersionService : BaseService, IAgreementVersionService
```
mtpultz, line 14: Similarly, this feels like it could be folded into the AgreementService

mtpultz:
@johnnyl6 can you coordinate with Nathan you're tickets overlap, and it appears you have overlap in your tickets.

prime-dotnet-webapi/Services/AgreementVersionService.cs
```diff
@@ -0,0 +1,47 @@
...
+            foreach (var type in AgreementTypeExtensions.EnrolleeAgreementTypes())
+            {
+                agreementVersionList.Add(await GetLatestEnrolleeAgreementVersionsByTypeAsync(type));
+            }
+            return agreementVersionList.AsQueryable()
+            .ProjectTo<AgreementVersionViewModel>(_mapper.ConfigurationProvider)
+            .ToList();
+        }
+
+        private async Task<AgreementVersion> GetLatestEnrolleeAgreementVersionsByTypeAsync(AgreementType type)
```
mtpultz, line 39: This could be `FetchNewestAgreementVersionOfType` to complement `FetchNewestAgreementVersionIdOfType` in the `AgreementService`

prime-dotnet-webapi/Services/AgreementService.cs
```diff
@@ -117,7 +117,7 @@ public async Task CreateEnrolleeAgreementAsync(int enrolleeId)
             var agreement = new Agreement
             {
                 EnrolleeId = enrolleeId,
-                AgreementVersionId = await FetchNewestAgreementVersionIdOfType(dto.NewestAssignedAgreement.Value),
+                AgreementVersionId = (await FetchNewestAgreementVersionOfTypeAsync(dto.NewestAssignedAgreement.Value)).Id,
```
mtpultz, line 120: Do you know that this will always have an `Id`?  You'll either need to maintain both methods or perform a null check.

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-toa-maintenance/enrollee-toa-maintenance.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  ) {
+    this.routeUtils = new RouteUtils(route, router, AdjudicationRoutes.routePath(AdjudicationRoutes.ENROLLEES));
+  }
+
+  public previewToa(agreementVersion: AgreementVersion): void {
+    this.previewingToa = agreementVersion;
+  }
+
+  public onBack(): void {
+    this.previewingToa ?
```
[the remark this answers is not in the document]
johnnyl6: > Do you expect to have `onBack` triggered and not route back? Doesn't feel like we should be using back to clear the preview. They can just choose another TOA to view or go back.

When previewing TOA `onBack` is triggered to show the list of TOA cards and hide the previewing TOA, not really route back.

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-toa-maintenance/enrollee-toa-maintenance.component.html
```diff
@@ -0,0 +1,45 @@
...
+          (back)="onBack()">
+  <app-page-header>TOA Maintenance</app-page-header>
+  <app-page-subheader *ngIf="!previewingToa">
+    <ng-container appPageSubheaderTitle>View TOA</ng-container>
+    <ng-container appPageSubheaderSummary>
+      Select what TOA you would like to view
+    </ng-container>
+  </app-page-subheader>
+
+  <div class="row"
```
mtpultz, line 12: You won't want to use the `[hidden]` binding since it doesn't remove the list markup from the DOM, and instead you'll want to use ngIfElse and templates.

Feels like you'll want to use `*ngIf="!previewingToa; else preview" `, and then 

```
<ng-container *ngIf="!previewingToa; else preview">
  <div class="row">
    <!-- List content -->
  </div>
</ng-container>

<ng-template *ngIf="previewingToa">
  <div class="row">
    <!-- Preview -->
  </div>
</ng-template>
```

johnnyl6: Somehow using ngIf triggers the observable async http call when set previewingToa=null

prime-dotnet-webapi/ViewModels/AutoMapping.cs
```diff
@@ -80,5 +80,9 @@ public AutoMapping()
         CreateMap<AuthorizedUser, AuthorizedUserViewModel>()
             .IncludeMembers(src => src.Party);
         CreateMap<Party, AuthorizedUserViewModel>();
+
+        CreateMap<AgreementVersion, AgreementVersionViewModel>()
+            .ForMember(dest => dest.UpdatedDate, opt => opt.MapFrom(src => src.UpdatedTimeStamp))
+            .ForMember(dest => dest.AgreementContent, opt => opt.MapFrom(src => src.Text));
```
james-hollinger, line 86: I'm not sure why these properties are named differently in the DB vs the ViewModel? 
`Agreements` have a property called AgreementContent, while `AgreementVersion`s have Text. The main difference is that AgreementContent is also rendered with Razor and includes styling, additional clauses, and could include other variables such as today's date or the person's name, etc. To me, it seems like keeping the VM property Text instead of AgreementContent is more accurate?

james-hollinger: For reference, we want to generally avoid using the audit columns whenever possible, since they are more like an implementation detail than actual business data. We could, as a crazy hypothetical example, have a database trigger that pushes a log whenever we update a row. Or implement temporal tables. In this case however, it's fine for now at least since its what they want in the existing mock-ups. I feel we should probably still make another column called PublishedDate or something like that rather than using the audit columns

james-hollinger: Also for reference, once published we won't be updating the AgreementVersions, we will only be inserting new records. These represent records of legal agreements, so we cant have the admins go back and ret-con an agreement someone already signed.

johnnyl6: > For reference, we want to generally avoid using the audit columns whenever possible, since they are more like an implementation detail than actual business data. We could, as a crazy hypothetical example, have a database trigger that pushes a log whenever we update a row. Or implement temporal tables. In this case however, it's fine for now at least since its what they want in the existing mock-ups. I feel we should probably still make another column called PublishedDate or something like that rather than using the audit columns

Totally agree with the db trigger implementation approach for audit date (or any other columns possible) its less error prone so developer won't forget assigning value to them. I think right now it's okay since the ticket is just previewing and hopefully more implementation will be added in future ticket.

prime-dotnet-webapi/Services/AgreementService.cs
```diff
@@ -278,6 +278,18 @@ public async Task<SignedAgreementDocument> AddSignedAgreementDocumentAsync(int a
...
 
+        public async Task<IEnumerable<AgreementVersionViewModel>> GetLatestEnrolleeAgreementVersionsAsync()
+        {
+            var agreementVersionList = new List<AgreementVersion>();
+            foreach (var type in AgreementTypeExtensions.EnrolleeAgreementTypes())
+            {
+                agreementVersionList.Add(await FetchNewestAgreementVersionOfTypeAsync(type));
+            }
+            return agreementVersionList.AsQueryable()
+                .ProjectTo<AgreementVersionViewModel>(_mapper.ConfigurationProvider)
```
james-hollinger, line 289: You should almost never have to call `AsQueryable()` on something in this project. IQueryable more or less represents the application boundary that Entity Framework interfaces with to translate our C# code into SQL.  Automapper's `ProjectTo()` turns our ViewModel definitions into `Select()` statements that EF can then turn into SQL to retrieve our Models as ViewModels.

In this case, you have already materialized all the `AgreementVersion`s from the DB on line 286, so this collection already exists in memory. You should be able to just use

```C#
return _mapper.Map<IEnumerable<AgreementViewModel>>(agreementVersionList);
```
to do this.

prime-dotnet-webapi/Models/Agreements/AgreementType.cs
```diff
@@ -19,5 +23,12 @@ public static bool IsEnrolleeAgreement(this AgreementType agreementType)
         {
             return agreementType != AgreementType.CommunityPracticeOrgAgreement && agreementType != AgreementType.CommunityPharmacyOrgAgreement;
         }
+
+        public static IEnumerable<AgreementType> EnrolleeAgreementTypes()
+        {
+            return Enum.GetValues(typeof(AgreementType))
+                .Cast<AgreementType>()
+                .Where(t => t.IsEnrolleeAgreement());
```
james-hollinger, line 31: This isn't really an extension method, this is a static helper method. From a technical perspective, there isn't much difference but the methods are called using a different syntax. Extension methods have additional syntactic sugar that makes them look like instance methods on objects rather than static methods. `IsEnrolleeAgreement()`, for instance, can be called like 

```C#
var agreementType = AgreementType.OboToa;
agreementType.IsEnrolleeAgreement();
```
rather than

```C#
var agreementType = AgreementType.OboToa;
AgreementTypeExtensions.IsEnrolleeAgreement(agreementType);
```

This is just as an aside, this code is fine for now

prime-dotnet-webapi/Services/AgreementService.cs
```diff
@@ -278,6 +278,18 @@ public async Task<SignedAgreementDocument> AddSignedAgreementDocumentAsync(int a
             return signedAgreement;
         }
 
+        public async Task<IEnumerable<AgreementVersionViewModel>> GetLatestEnrolleeAgreementVersionsAsync()
+        {
+            var agreementVersionList = new List<AgreementVersion>();
+            foreach (var type in AgreementTypeExtensions.EnrolleeAgreementTypes())
+            {
+                agreementVersionList.Add(await FetchNewestAgreementVersionOfTypeAsync(type));
```
james-hollinger, line 286: As an FYI, this is going to hit the DB once for every AgreementType

johnnyl6: > As an FYI, this is going to hit the DB once for every AgreementType

yep aware of that. Thought about one single query but seems linq is limited to achieve it.

prime-dotnet-webapi/Models/Agreements/AgreementType.cs
johnnyl6: Since enum could not hold any method so borrowed some estate from the extension class for the helper method😀

johnnyl6:
Merged to PR #1274,  closing PR.

PR #1272, opened 2021-05-18
PRIME-1571 PLR integration testing

neophyte57:
Ways to test:

1. Test locally without need for Nginx (note that the client certificate is still expected in the HTTP header) by executing `curl --insecure --header "@../Utils/Auth/client-pem-url-encoded-headers.txt" --data-binary "@PRPM_IN303030CA-2.xml" https://127.0.0.1:5001/api/PLRHL7` when in prime-dotnet-webapi-tests/Integration directory.  

The `dotnet` process needs to be running with the environment variable `PLR_INTEGRATION_CLIENT_CERT_THUMBPRINT` set to 6FA357305008C8EA814EA98768C21783DCCA374E

2. Test in OpenShift environment:  `curl --insecure --cert client.crt:changeme --key client.key --data-binary "@PRPM_IN303030CA.xml" https://pr-1272-mutualauth.pharmanetenrolment.gov.bc.ca/api/v1/PLRHL7`.  The files `client.crt`, `client.key` and `PRPM_IN303030CA.xml` (among other test data files) are in subfolders of `prime-dotnet-webapi-tests`.

PR #1277, opened 2021-05-19
Temporary fix for `The Route ... is invalid` error

neophyte57:
Fixes `The Route "angular-frontend-develop-mutualauth" is invalid: spec.host: Invalid value: "mutualauth-develop.pharmanetenrolment.gov.bc.ca": field is immutable`

FYI, for PLR integration testing, removed route will be re-added in the relevant PR (#1272 currently)


mtpultz (approved):
image

PR #1280, opened 2021-05-26
Rename Submission Actions

james-hollinger:
`SubmissionAction` is an increasingly confusing name since we have a `Submission` table, so I've renamed it to `EnrolleeStatusAction` and changed some URLs.  

I have not moved any service or controller methods to keep this simple for now. We should consider further breaking up or condensing of controller/service methods in the future.


mtpultz (approved):
image

johnnyl6 (approved):

PR #1284, opened 2021-05-27
Feature/prime 1566 view email/toa

hogmann:
Had to make a new pr to cover the fact there were migration changes


prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-toa-maintenance-view-page/enrollee-toa-maintenance-view-page.component.html
```diff
@@ -0,0 +1,21 @@
...
+    </ng-container>
+  </app-page-subheader>
+
+  <app-page-subheader>
+    <ng-container appPageSubheaderTitle>
+      Preview
+    </ng-container>
+  </app-page-subheader>
+
+  <div [innerHtml]="enrolleeAgreementVersion?.text"></div>
```
johnnyl6, line 20: Note that the TOA template HTML has custom CSS class, which is in `@import 'mixins/agreements';` The final rendering style is slightly different if the css is not imported into `enrollee-toa-maintenance-view-page.component.scss`

johnnyl6 (approved):
LGTM 🎉

prime-dotnet-webapi/Services/EmailRenderingService.cs
```diff
@@ -28,17 +32,17 @@ public async Task<Email> RenderBusinessLicenceUploadedEmailAsync(string recipien
                 from: PrimeEmail,
                 to: recipientEmail,
                 subject: "Site Business Licence Uploaded",
-                body: await _razorConverterService.RenderTemplateToStringAsync(RazorTemplates.Emails.BusinessLicenceUploaded, viewModel)
```
james-hollinger, line 31: for several of these, the `body:` named attribute is removed, but some are not. Why is that?

hogmann: That is really weird since i tested these and they were there once
but fixed

james-hollinger (approved):
LGTM

PR #1292, opened 2021-05-29
PRIME-1582 Convert Excel Document to SQL for loading into PRIME PROD DB

johnnyl6:
Convert an excel file containing paper site info and remote user to a SQL file to be loaded into prod db.


utilities/SiteExcelToSql/ExcelReaderExtension.cs
```diff
@@ -0,0 +1,134 @@
...
+                if (hour != null)
+                {
+                    site.BusinessHours.Add(hour);
+                }
+            }
+
+            return site;
+        }
+
+        public static RemoteUser PasseRemoteUser(this IExcelDataReader reader)
```
neophyte57, line 123: Should that be `ParseRemoteUser`?

mtpultz (commented):
LGTM, but worth having a second set of eyes

mtpultz (dismissed):
image

utilities/SiteExcelToSql/Program.cs
```diff
@@ -0,0 +1,160 @@
...
+            // data section
+            for (var i = 0; i < siteList.Count; i++)
+            {
+                // output site SQL
+                writer.WriteLine("\t-- Site {0} - {1} | {2}", i + 1, siteList[i].PEC, siteList[i].DoingBusinessAs);
+                writer.WriteLine("\tINSERT INTO public.\"Site\" (\"CreatedUserId\", \"CreatedTimeStamp\", \"UpdatedUserId\", \"UpdatedTimeStamp\", \"PEC\", \"Completed\", \"DoingBusinessAs\", \"OrganizationId\", \"Status\") "
+                                + "VALUES(vCreatorUUID, now(), vCreatorUUID, now(), '{0}', TRUE, '{1}', vOrganizationID, 1) "
+                                + "RETURNING \"Id\" INTO vSiteId;",
+                                siteList[i].PEC, siteList[i].DoingBusinessAs.Replace("'", "''"));
+
```
neophyte57, line 114: Shouldn't `CareSettingCode` be populated, even though it is nullable?  It appears to be missing from the spreadsheet?  Can we assume they are all Community Health Practice?  

Also after executing the generated SQL, I see a lot of rows in the database but they don't appear in the /admin interface?  It's not explicitly mentioned in PRIME-1576 but I'm assuming that that is the expectation?

neophyte57 (changes requested):

neophyte57:
Unfortunately with the upcoming `SiteStatus` table addition, the generated SQL will need to change  :-(   Also, my understanding is that this program needs to take the output of PRIME-1583 and add the association with the business licence to the database appropriately; apologies I don't know if this has been communicated "in writing" anywhere ...

utilities/SiteExcelToSql/Program.cs
johnnyl6: > Shouldn't `CareSettingCode` be populated, even though it is nullable? It appears to be missing from the spreadsheet? Can we assume they are all Community Health Practice?
> 
> Also after executing the generated SQL, I see a lot of rows in the database but they don't appear in the /admin interface? It's not explicitly mentioned in PRIME-1576 but I'm assuming that that is the expectation?

- `CareSettingCode` is missing in the excel.
- The site needs to have organization and signing authority data together in the db to be able to show up in /admin UI.

johnnyl6:
> Unfortunately with the upcoming `SiteStatus` table addition, the generated SQL will need to change :-( Also, my understanding is that this program needs to take the output of PRIME-1583 and add the association with the business licence to the database appropriately; apologies I don't know if this has been communicated "in writing" anywhere ...

yes agreed that this will have to be revisited after site table model is changed

utilities/SiteExcelToSql/Program.cs
johnnyl6: Org ID should be not imported as per Sarah. I don't think we are could import everyting exactly as the workflow in PRIME but only to import as much as we could.

PR #1293, opened 2021-05-29
PRIME-1583 Auto Upload Business Licences

neophyte57:


prime-angular-frontend/nginx.template.conf
```diff
@@ -36,6 +36,7 @@ server {
     }
     location /api/docman/ {
         proxy_pass        http://document-manager$SUFFIX-backend:6001/;
+        client_max_body_size  20m;
```
james-hollinger, line 39: Why was 20 MB chosen here? The document manager has a maximum file size of 400 MB:
image

james-hollinger: While the frontend restricts uploads to 3MB:
image

neophyte57: Someone was saying that we may need to deal with 20 MB business licence files so I was just testing that it is possible

utilities/business-licence-upload/upload-business-licences.sh
```diff
@@ -0,0 +1,15 @@
...
+  -H "Content-Type: application/x-www-form-urlencoded" \
+  -d "grant_type=client_credentials" \
+  -d "client_id=${KEYCLOAK_CLIENT_ID}" \
+  -d "client_secret=${KEYCLOAK_CLIENT_SECRET}" | python -c "import sys, json; print(json.load(sys.stdin)['access_token'])")
+echo -e ${TOKEN}
+
+FILE=$1
+
+echo -e "-------- Uploading ${FILE} --------\n"
+DOC_GUID=$(curl -X POST --data-binary "@${FILE}" -H "Authorization: Bearer ${TOKEN}" "${DOCMAN_DOCUMENTS_URL}?folder=licences&filename=${FILE}" | python -c "import sys, json; print(json.load(sys.stdin)['document_guid'])")
```
james-hollinger, line 14: Currently, Business License Documents are uploaded to the folder "business_licences"

image

This should probably upload to the same folder. If we would rather save them in a new folder it should be named more explicitly like "paper_form_business_licenses" or "manually_uploaded_business_licenses" or something

prime-angular-frontend/nginx.template.conf
james-hollinger: Ah, I see. In that case it should probably be set to the maximum size the Document Manager can accept, 400MB.

neophyte57: I've removed this change since it seems the scanned business licence files are not expected to be larger than 1 MB

mtpultz (approved):

neophyte57:
FYI, to conserve OCP 4 resources, the server environment for this pull request may not be available.  To test, the script can be pointed at the `develop` environment (using `DOCMAN_DOCUMENTS_URL` environment variable for Docker)

johnnyl6 (approved):
alt text

PR #1294, opened 2021-05-31
PRIME-1547 no adjudication box for editable enrolments

johnnyl6:
hide adjudication box for editable enrolments


mtpultz (approved):

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.ts
```diff
@@ -25,6 +26,7 @@ export class EnrolleeOverviewComponent extends AdjudicationContainerComponent im
   public enrollee: HttpEnrollee;
   public enrolment: Enrolment;
   public enrolleeNavigation: EnrolleeNavigation;
+  public hideAdjudication: boolean;
```
mtpultz, line 29: My only suggestion would be to make this `showAdjudication`, which by default will be false, and then let the data show it.  Currently as the page loads it should be hidden and then we decide to show it.

PR #1298, opened 2021-06-03
PRIME-1511 GIS Authentication Finalized

mtpultz:
PR is ready for code review, but can't be tested without proper test cards for PHSA. The only way this PR can be tested is through local development and setting the idpHint to IDIR in the gis-login-page.component.ts.


AnaisHebert (approved):
🍟

PR #1316, opened 2021-06-04
Consolidate Configuration

james-hollinger:
- Removed all of the entity configuration from `ApiDbContext` and moved it into Configuration files (where all future configuration should go)

- Pruned the DbSets with zero references.  This does not drop any tables, they are sill in our DB since they are referenced from another model. If we need to query one of these tables directly we can add them back in as needed. Kept the `Job` DbSet since we want to keep the table even though its not referenced anywhere (and so would have been dropped if we removed the DbSet).

- Removed redundant configuration from several entities. EF is smart enough to infer relationships if they are simple and/or follow the naming conventions. The following entities had redundant configuration removed partially or entirely:
  - AssignedPrivilege
  - BusinessEvent
  - Enrollee
  - EnrolleeCredential
  - EnrolleeNote
  - EnrolmentStatusReason
  - EnrolmentStatus
  - RemoteAccessLocation
  - RemoteAccessSite
  - RemoteUser
  - Site
  - Submission
 
- Sorted the configuration files into seeded and unseeded folders to make finding seed data easier.

- No changes to the DB should appear if we make a new migration.


mtpultz (approved):
image

PR #1340, opened 2021-06-14
Updated to allow maintenance page to load

mtpultz:


AnaisHebert (approved):
🦆🦆🦆🦆

itchison:
Can we quickly update the message? It currently says:
If you require immediate assistance, but if the problem persists feel free to

PR #1363, opened 2021-06-15
PRIME-1621 Renew Enrolment UI

johnnyl6:


prime-angular-frontend/src/app/modules/enrolment/pages/pharmanet-enrolment-summary/pharmanet-enrolment-summary.component.ts
```diff
@@ -63,6 +64,8 @@ export class PharmanetEnrolmentSummaryComponent extends BaseEnrolmentPage implem
     this.showHealthAuthority = true;
     this.form = this.buildVendorEmailGroup();
     this.careSettingConfigs = [];
+    const navigation = this.router.getCurrentNavigation();
```
mtpultz, line 67: You can definitely pass state, but you're only passing a single value and not a complex object.  It isn't actually state as much as it's a flag indicating that this is a renewal.  I'd probably pass this as a queryParam then you only have to do `this.route.snapshot.queryParams?.isRenewal === 'true'` as a check.  As a queryParam it also means they could refresh their screen and not have it switch to the non-renewal view.

mtpultz (approved):
image

PR #1369, opened 2021-06-17
Fix Broken OPTIONS Request

james-hollinger:
W H O O P S


neophyte57:
How does the problem manifest itself?

hogmann (approved):
LGTM

mtpultz (approved):

AnaisHebert (approved):
M E R G E   IT   I N

james-hollinger:
> 
> 
> How does the problem manifest itself?

In local development, the browser makes a pre-flight OPTIONS request before any regular request. ASP tries to enumerate the actions available on the controller by inspecting the return values of the methods. The custom Ok(), BadRequest(), and NotFound() methods I added are seen as actions, but with no routes assigned to them. This causes the API to throw an error saying they are ambiguous since they all have the same route and it can't tell them apart. The fix is to add the attributes I missed tagging the methods as not being actions (they are ActionResults).

PR #1372, opened 2021-06-21
Use Effective Date instead of Updated Date in TOA Maintenance page

james-hollinger:
The "Last Modified Date" on the TOA maintenance page shows September 19th, 2019 for every TOA. This is because it comes from the "Updated Date" column, which is set to Sept 19th for every version. A far more useful date for the current state of the app is the "Effective Date", which is the date that the Agreement Version became effective in the application.

image
 
 Becomes
 
image

We can revisit what data should be displayed on the cards when/if adjudicators have the ability to edit agreements.


johnnyl6 (approved):
LGTM alt text

mtpultz (approved):

PR #1383, opened 2021-06-23
PRIME-1586 Add state to sites

neophyte57:


prime-dotnet-webapi-tests/Unit Tests/SubmissionStateEngine.Tests.cs
```diff
@@ -13,107 +13,108 @@ public class SubmissionStateEngineTests
...
+        public void TestAllowableActions_Editable(EnrolleeStatusAction action)
         {
             // Arrange
             var status = EnrolmentStatus.FromType(StatusType.Editable, 0);
             var expectedAllowed = new[]
             {
-                SubmissionAction.LockProfile,
-                SubmissionAction.DeclineProfile
+                EnrolleeStatusAction.LockProfile,
+                EnrolleeStatusAction.DeclineProfile
```
james-hollinger, line 31: Thank you for fixing all these.

prime-dotnet-webapi/Services/SiteRegistrationService.cs
```diff
@@ -0,0 +1,35 @@
...
+                    newStatus = SiteStatusType.Locked;
+                    break;
+                case SiteRegistrationAction.RequestChange:
+                case SiteRegistrationAction.Unapprove:
+                    newStatus = SiteStatusType.Active;
+                    break;
+                default:
+                    return false;
+            }
+            return SiteStatusStateEngine.AllowableStatusChange(currentStatus, newStatus);
```
james-hollinger, line 32: What is the purpose of having both this Service with `CanPerformSiteStatusAction()` and the `SiteStatusEngine` with `AllowableStatusChange()`? 
It feels like you only need this one method; the important question is can you perform the given action in the given status.

johnnyl6: > What is the purpose of having both this Service with `CanPerformSiteStatusAction()` and the `SiteStatusEngine` with `AllowableStatusChange()`?
> It feels like you only need this one method; the important question is can you perform the given action in the given status.

Can defintely merge them into one place. The original idea was just to make it the same pattern of the SubmissionStateEngine which also has a wrapping service layer.

prime-dotnet-webapi/Services/SiteRegistrationService.cs
```diff
@@ -0,0 +1,35 @@
+using Prime.Models.Api;
+using Prime.Engines;
+using Prime.Models;
+
+namespace Prime.Services
+{
+    public class SiteRegistrationService : ISiteRegistrationService
```
james-hollinger, line 7: The Services in our app are registered with the Dependency Injection framework to handle creating dependencies. This does not have any dependencies, so isn't really a service and has no benefit being registered as such.

johnnyl6: > The Services in our app are registered with the Dependency Injection framework to handle creating dependencies. This does not have any dependencies, so isn't really a service and has no benefit being registered as such.

Similar reason for the service/engine pattern. can remove the service layer and call the engine directly.

prime-dotnet-webapi/Models/Statuses/SiteStatus.cs
```diff
@@ -0,0 +1,35 @@
...
+    [Table("SiteStatus")]
+    public class SiteStatus : BaseAuditable
+    {
+        [Key]
+        public int Id { get; set; }
+
+        public int SiteId { get; set; }
+
+        // [JsonIgnore]
+        // public Site Site { get; set; }
```
james-hollinger, line 17: Is this commented code intentional?

prime-dotnet-webapi/Models/Statuses/SiteStatus.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public SiteStatusType StatusType { get; set; }
+
+        public DateTimeOffset StatusDate { get; set; }
+
+
+        // TODO: Why not constructor?
```
james-hollinger, line 24: There are two reasons this was done like this for Enrollee statuses:
1. I generally prefer to use factory methods rather than constructors when properties are being set other than the given parameters. In this case, StatusDate is being set but isn't a parameter.
2. De-Serializing an object uses, by default, the empty constructor of the class. The curly brace object construction syntax also uses the empty constructor. This:
```C#
var object = new SiteStatus
{
    SiteId = siteId,
    StatusType = siteStatusType,
    StatusDate = DateTimeOffset.Now
}
```
is actually equivalent to
```C#
var object = new SiteStatus();
object.SiteId = siteId;
object.StatusType = siteStatusType;
object.StatusDate = DateTimeOffset.Now;
```

So we want the empty constructor on the class. If no constructor is specified, the empty one is added implicitly to the class. If *any* other constructor is specified, however, the implicit empty constructor is removed and you have to explicitly add it to the class if you want it. So you would have to add both the empty constructor and one that takes SiteStatusType and SiteId. In terms of lines of code, I went with 

```C#
public static SiteStatus FromType(SiteStatusType siteStatusType, int siteId)
{
    return new SiteStatus
    {
        SiteId = siteId,
        StatusType = siteStatusType,
        StatusDate = DateTimeOffset.Now
    };
}
```
vs
```C#
public SiteStatus() {}

public SiteStatus(SiteStatusType siteStatusType, int siteId)
{
    SiteId = siteId;
    StatusType = siteStatusType;
    StatusDate = DateTimeOffset.Now;
}
```

prime-dotnet-webapi/Models/SiteRegistration/Site.cs
```diff
@@ -84,5 +84,33 @@ public IEnumerable<DayOfWeek> DaysOpen(DateTimeOffset? atTime = null)
...
+        }
+
+        /// <summary>
+        /// Gets the most recent Status of the Site.
+        /// </summary>
+        [NotMapped]
+        [Computed]
+        public SiteStatusType Status
+        {
+            get => (SiteStatuses != null ?
```
AnaisHebert, line 108: image
im getting this error when trying to run this locally

prime-dotnet-webapi/Services/OrganizationService.cs
```diff
@@ -78,15 +78,19 @@ public async Task<IEnumerable<OrganizationListViewModel>> GetOrganizationsByPart
                 .Include(o => o.SigningAuthority)
                         .ThenInclude(sa => sa.Addresses)
                             .ThenInclude(pa => pa.Address)
+                // .Include(o => o.Sites)
+                //     .ThenInclude(s => s.SiteStatuses)
```
james-hollinger, line 82: Commented out code?

prime-dotnet-webapi/Models/SiteRegistration/Site.cs
neophyte57: How to reproduce this?

prime-dotnet-webapi/Models/Statuses/SiteStatus.cs
neophyte57: With the code enabled, I get this, despite the `[JsonIgnore]`?
image

My use of the class doesn't seem impeded without the property?

prime-dotnet-webapi/Services/OrganizationService.cs
neophyte57: Yeah, we should talk about `DecompileAsync` again and then I'll remove

prime-dotnet-webapi/Models/SiteRegistration/Site.cs
james-hollinger: FirstOrDefault().StatusType throws if there is nothing in the list.

prime-dotnet-webapi/Models/Statuses/SiteStatus.cs
AnaisHebert: wrong Json import.
using Newtonsoft.Json;

prime-dotnet-webapi/Migrations/20210614235905_AddSiteStatusHistory.cs
```diff
@@ -0,0 +1,62 @@
...
+    public partial class AddSiteStatusHistory : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            // Existing data needs to be migrated before column is dropped.
+            // However, new Site rows cannot run into not-nullable constraint.
+            migrationBuilder.AlterColumn<int>(
+                name: "Status",
+                table: "Site",
+                nullable: true);
```
james-hollinger, line 16: Are you intending to add the migration in this PR?

prime-dotnet-webapi/Models/Statuses/SiteStatus.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+using System.Text.Json.Serialization;
```
james-hollinger, line 4: This is the wrong JSON import. We use Newtonsoft.Json

james-hollinger: I would assume this is why you are seeing the circular reference error in your code; `Site` isn't being properly JsonIgnored.

prime-dotnet-webapi/Migrations/20210614235905_AddSiteStatusHistory.cs
neophyte57: Data migration?  It's in https://bcgovmoh.atlassian.net/browse/PRIME-1646

prime-dotnet-webapi/Models/SiteRegistration/Site.cs
neophyte57: Not sure how I never ran into this lately ... different data?

prime-dotnet-webapi/Models/Statuses/SiteStatus.cs
neophyte57: Valid points

mtpultz (dismissed):
image

prime-dotnet-webapi/Migrations/20210614235905_AddSiteStatusHistory.cs
james-hollinger: What is the status of this SQL? Is it still being tested?
I hope that the intention is to add it to this migration rather than running it against prod directly.

prime-dotnet-webapi/Models/Statuses/SiteStatus.cs
james-hollinger: The greatest accidental haiku of all time

mtpultz (dismissed):

prime-dotnet-webapi/Migrations/20210614235905_AddSiteStatusHistory.cs
johnnyl6: Data SQL has been tested locally and was preferred by Jason to run separately from .NET migration.

johnnyl6 (approved):
LGTM.alt text

prime-dotnet-webapi/Migrations/20210614235905_AddSiteStatusHistory.cs
itchison: After looking at the migration I dont see any PII data in it. Apologies if I confused the issue.. but if there's no PII it should be run in a migration. Thanks!

prime-dotnet-webapi/Migrations/20210614235905_AddSiteStatusHistory.cs
```diff
@@ -44,6 +37,30 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
+                    + " FROM public.\"Site\""
+                    + " WHERE \"SubmittedDate\" IS NOT NULL"
+            );
+
+            // the old status 1 without SubmittedDate becomes the new Active
+            migrationBuilder.Sql(
+                  "INSERT INTO public.\"SiteStatus\" (\"CreatedUserId\", \"CreatedTimeStamp\", \"UpdatedUserId\", \"UpdatedTimeStamp\", \"SiteId\", \"StatusType\", \"StatusDate\")"
+                + " SELECT \"CreatedUserId\", \"CreatedTimeStamp\", \"UpdatedUserId\", \"UpdatedTimeStamp\", \"Id\" AS \"SiteId\", 1 AS \"StatusType\", \"UpdatedTimeStamp\" AS \"StatusDate\""
+                + " FROM public.\"Site\""
+                + " WHERE \"Status\" = 1 AND \"SubmittedDate\" IS NULL"
```
james-hollinger, line 56: All of our existing examples of calling bare SQL have been cleaned up from the repo, unfortunately. If you would like to see some examples, try checking out any commit on develop before PR#1367 (such as 990d624211a22a5eb3b8ca0e2ca98b85ca96e5f2) and looking at 20200609020423_UpdateVendorTableNameAndVendorIdToCode.cs. 

I have two suggestions for how to format these SQL queries
1. Please use the string literal syntax `@""`. Among other reasons, it means you do not have to use the '+'s to string concatenate these strings across multiple lines.
2. Please make sure you terminate your SQL with a semicolon. This code might run fine in DBeaver but may cause issues when made into an update script in our deployment pipeline. A missing semicolon like this has caused data loss in production, so please take care.

Here is an example from the above referenced migration:
```C#
migrationBuilder.Sql(@"
    ALTER TABLE ""Vendor""
    RENAME COLUMN ""Id"" TO ""Code"";

    ALTER TABLE ""Site""
    RENAME COLUMN ""VendorId"" TO ""VendorCode"";
");
```

mtpultz (dismissed):
image

prime-dotnet-webapi/Models/SiteRegistration/Site.cs
```diff
@@ -84,5 +92,33 @@ public IEnumerable<DayOfWeek> DaysOpen(DateTimeOffset? atTime = null)
...
+        [Computed]
+        public SiteStatusType Status
+        {
+            get => (SiteStatuses.Count > 0 ?
+                SiteStatuses
+                .OrderByDescending(s => s.StatusDate)
+                .ThenByDescending(s => s.Id)
+                .FirstOrDefault().StatusType :
+                SiteStatusType.Active);
+        }
```
james-hollinger, line 122: This version of the Property still has a ternary in it, so the generated SQL is still rather complicated:
original

You could remove the ternary by doing a Select() instead:
```C#
public SiteStatusType Status
{
    get => SiteStatuses
        .OrderByDescending(s => s.StatusDate)
        .ThenByDescending(s => s.Id)
        .Select(x => (SiteStatusType?)x.StatusType)
        .FirstOrDefault() ?? SiteStatusType.Active;
}
```
but the SQL looks like this:
Untitled
which isn't much better.

I personally think this Property probably shouldn't specify a default value for Status if there are none. Since this is a Property on the data model, it should really be a `SiteStatusType?` so it can return null if there is no current status, which more accurately represents the data rather than the business logic surrounding the data. To be clear, I think this is just something to consider going forward, it's not a change I'm requesting for this PR. 

With this change to the code:
```C#
public SiteStatusType? Status
{
    get => SiteStatuses
        .OrderByDescending(s => s.StatusDate)
        .ThenByDescending(s => s.Id)
        .Select(x => (SiteStatusType?)x.StatusType)
        .FirstOrDefault();
}
```
the generated SQL becomes:
nullable
which is much cleaner.

james-hollinger:
It looks like the chmod settings for Startup.cs have been changed (I assume accidentally): 
image

To avoid any unintended side effects, I've reverted the file back to what is on develop.
For reference, the command to revert a file like this is `git checkout develop -- C:\full\file\path\Startup.cs`

james-hollinger (approved):
👍

mtpultz (approved):

prime-dotnet-webapi/Models/SiteRegistration/Site.cs
neophyte57: Thanks but I'm seeing even more complicated SQL being generated for the ternary situation ... do I have my breakpoint at the wrong spot?  I'll need to show you on Monday @james-hollinger

neophyte57:
> It looks like the chmod settings for Startup.cs have been changed (I assume accidentally):
> image
> 
> To avoid any unintended side effects, I've reverted the file back to what is on develop.
> For reference, the command to revert a file like this is `git checkout develop -- C:\full\file\path\Startup.cs`

Not sure how that happened!  Thanks.

PR #1390, opened 2021-06-25
PRIME-1665-removed collection component from overview

Paahn:
- Removed collection notice component from overview


mtpultz (approved):
image

AnaisHebert (approved):
🦝

PR #1393, opened 2021-06-29
update readme to include alternate way of running app locally

Paahn:
- included alternate way of running app that worked for me

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,11 @@ Docker command:
 
 	docker-compose up --build	
 
+or
+  run containers with Docker desktop
+  cd prime-angular-frontend
+  ng serve
+
 To update the database schema, first update the model file in the
 [Models](prime-dotnet-webapi/Models) folder, and rebuild using:
 
--- a/README.md
+++ b/README.md
@@ -15,10 +15,23 @@ Docker command:
 
 	docker-compose up --build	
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,6 +15,11 @@ Docker command:
 
 	docker-compose up --build	
 
+or
```
mtpultz, line 18: I might be a bit more verbose to help guide the next developer.  You could add a sub-section for local development, and indicate steps and requirements, like:

### LOCAL DEVELOPMENT

For local development where the web client and server are run directly off the host system some of the containers should be stopped to avoid port collisions and to reduce resources on containers that aren't required.

**Required Containers**

* primedb
* mailhog
* document-manager-backend
* prime-redis
 
To run the backend open `prime-dotnet-webapi` in VSCode and launch the API using the debugger (or F5).  For the web client run the web client locally type:

```
cd prime-angular-frontend
ng serve -o
```

I added this using markdown so if you want the markdown let me know, and I can pass it through Slack.

Paahn: That makes more sense, I love it. I got the markup from using edit on your comment :))

mtpultz (approved):
image

AnaisHebert (approved):
🎂🎂

PR #1404, opened 2021-06-29
PRIME-1654 Multiple Banners

hogmann:
Just cleaning up some front end test, will push shortly.


mtpultz (approved):

prime-dotnet-webapi/Services/BannerService.cs
```diff
@@ -42,35 +52,45 @@ public async Task<BannerViewModel> GetBannerAsync(BannerLocationCode locationCod
...
+                .ToListAsync();
+
+            return banners.Select(banner =>
+            {
+                // Banner time stamps are stored as UTC, but are materialized as DateTimeKind.Unspecified
+                // Once we upgrade our EF / Npgsql, we can use NodaTime instead
+                banner.StartTimestamp = DateTime.SpecifyKind(banner.StartTimestamp, DateTimeKind.Utc);
+                banner.EndTimestamp = DateTime.SpecifyKind(banner.EndTimestamp, DateTimeKind.Utc);
+                return banner;
+            });
```
james-hollinger, line 69: image

While its not _wrong_ to use a `Select()` statement this way, one main goal of functional programming is to avoid causing side-effects. So using functional programming to intentionally cause side effects this way _feels_ bad to me. I would prefer a `foreach()` loop when you want to mutate the source objects or cause other side effects. `List` actually already has a `.ForEach()` method, so this code can be:

```C#
banners.ForEach(banner =>
{
    // Banner time stamps are stored as UTC, but are materialized as DateTimeKind.Unspecified
    // Once we upgrade our EF / Npgsql, we can use NodaTime instead
    banner.StartTimestamp = DateTime.SpecifyKind(banner.StartTimestamp, DateTimeKind.Utc);
    banner.EndTimestamp = DateTime.SpecifyKind(banner.EndTimestamp, DateTimeKind.Utc);
});

return banners;
```

prime-dotnet-webapi/Controllers/BannersController.cs
```diff
@@ -24,111 +25,163 @@ public BannersController(IBannerService bannerService)
...
         /// <param name="viewModel"></param>
-        [HttpPut("enrolment-landing", Name = nameof(SetEnrolleeLandingBanner))]
+        /// <param name="locationCode"></param>
+        [HttpPost("", Name = nameof(CreateBanner))]
         [Authorize(Roles = Roles.PrimeSuperAdmin)]
         [ProducesResponseType(StatusCodes.Status401Unauthorized)]
         [ProducesResponseType(StatusCodes.Status403Forbidden)]
         [ProducesResponseType(typeof(ApiResultResponse<BannerViewModel>), StatusCodes.Status200OK)]
-        public async Task<ActionResult> SetEnrolleeLandingBanner(BannerViewModel viewModel)
+        public async Task<ActionResult> CreateBanner(BannerViewModel viewModel, [FromQuery] BannerLocationCode locationCode)
```
james-hollinger, line 39: Do we still use this endpoint when we also have the more specific ones?

prime-dotnet-webapi/Controllers/BannersController.cs
```diff
@@ -24,111 +25,163 @@ public BannersController(IBannerService bannerService)
...
-        [HttpGet("site-landing", Name = nameof(GetSiteLandingBanner))]
+        /// <param name="bannerId"></param>
+        /// <param name="viewModel"></param>
+        [HttpPut("{bannerId}", Name = nameof(UpdateBanner))]
         [Authorize(Roles = Roles.PrimeSuperAdmin)]
         [ProducesResponseType(StatusCodes.Status401Unauthorized)]
         [ProducesResponseType(StatusCodes.Status403Forbidden)]
         [ProducesResponseType(typeof(ApiResultResponse<BannerViewModel>), StatusCodes.Status200OK)]
-        public async Task<ActionResult> GetSiteLandingBanner()
+        public async Task<ActionResult> UpdateBanner(int bannerId, BannerViewModel viewModel)
```
james-hollinger, line 88: Since Id is in the View Model now, pragmatically we should be performing a little validation here to make sure the Id matches the Id in the View Model.

prime-dotnet-webapi/Controllers/BannersController.cs
```diff
@@ -24,111 +25,163 @@ public BannersController(IBannerService bannerService)
...
+        public async Task<ActionResult> GetSiteLandingBanners()
+        {
+            var banners = await _bannerService.GetBannersAsync(BannerLocationCode.SiteRegistrationLandingPage);
+            return Ok(banners);
         }
 
         // GET: api/banners/active?locationCode=4
         /// <summary>
-        /// Gets the active Banner by location code. Returns null result if no active banner
+        /// Gets the active Banners by location code. Returns null result if no active banner
```
james-hollinger, line 174: I believe this returns an empty array if there are no active banners.

prime-dotnet-webapi/Controllers/BannersController.cs
```diff
@@ -24,111 +25,163 @@ public BannersController(IBannerService bannerService)
...
+        /// <summary>
+        /// Get banners
+        /// </summary>
+        /// <param name="locationCode"></param>
+        [HttpGet("", Name = nameof(GetBanners))]
+        [Authorize(Roles = Roles.PrimeSuperAdmin)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiResultResponse<IEnumerable<BannerViewModel>>), StatusCodes.Status200OK)]
+        public async Task<ActionResult> GetBanners([FromQuery] BannerLocationCode locationCode)
```
james-hollinger, line 136: Do we need this endpoint if we have the more specific ones?

prime-dotnet-webapi/Controllers/BannersController.cs
hogmann: Other way around, need this one not the other ones.

prime-dotnet-webapi/Services/BannerService.cs
hogmann: Good to know, Javascript mind is always on autopilot

mtpultz (approved):

james-hollinger (approved):
LGTM

PR #1407, opened 2021-07-05
Fix envars and nginx logging

i5okie:


mtpultz (approved):

infrastructure/prime-app-template.yml
```diff
@@ -177,8 +185,16 @@ objects:
               - containerPort: 8890
                 protocol: TCP
             volumeMounts:
-              - name: nginx-logs
-                mountPath: /var/log/nginx/
+            - name: vanity-tls-certs
```
itchison, line 188: Im surprised that the certs was missing?

infrastructure/prime-app-template.yml
```diff
@@ -371,7 +387,7 @@ objects:
                   secretKeyRef:
                     name: ${SVC_NAME}-postgres
                     key: database-password
-              - name: POSTGRESQL_USERNAME
+              - name: POSTGRESQL_USER
```
itchison, line 390: Does this secret exist already? Do we have to time the deployments to prod?

prime-angular-frontend/openshift.nginx.conf
```diff
@@ -32,6 +32,19 @@ server {
     proxy_pass http://$SVC_NAME-webapi:8080/api/PLRHL7;
     #        proxy_set_header  X-SSL-CERT $ssl_client_escaped_cert;
   }
+  location /nginx_status {
```
itchison, line 35: Am I right that this will only be accessible on it's on pod? Shouldn't the cluster have access to this for a status check?

infrastructure/prime-app-template.yml
i5okie: We load them directly into Route in prod and test. but not dev.

infrastructure/prime-app-template.yml
i5okie: There was a change to the Entrypoint. Along with that, POSTGRESQL_USERNAME got added with POSTGRESQL_USER (duplicate) before. this is a fix for that.
Won't have to time the deploy.

itchison (approved):
lgtm thanks for addressing my comments.

mtpultz (approved):

PR #1409, opened 2021-07-06
PRIME 1676 Paper Enrolment Guard

Paahn:
For ticket #[1676](https://bcgovmoh.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=PRIME&modal=detail&selectedIssue=PRIME-1676)

- Guarding the paper enrolment routing module and redirecting to next-steps if the user has already been approved.
- Added Authentication and Adjudication guards


prime-angular-frontend/src/app/modules/paper-enrolment/shared/guards/finalized-enrolment.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+    const enrolleeId = +params.eid;
+    if (enrolleeId) {
+      return this.paperEnrolmentResource.getEnrolleeById(enrolleeId)
+        .pipe(
+          map((enrollee: HttpEnrollee) => {
+            return this.routeDestination(routePath, enrollee, params);
+          })
+        );
+    } else {
+      return new Promise(async (resolve, reject) => resolve(true));
```
mtpultz, line 46: You can simplify this by dropping the `else` and returning `of(true)`.  Where `of` will emit a list of values then complete so in this case emit `true` and complete.  With respect to the else there's nothing else to run so it's superfluous and can be omitted.

```
if(someStatement) {
  return this.paperEnrolmentResource.getEnrolleeById(...)
    .pipe(...);
}

return of(true);
```

prime-angular-frontend/src/app/modules/paper-enrolment/shared/guards/finalized-enrolment.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+    return this.checkAccess(url, next.params);
+  }
+
+  protected checkAccess(routePath: string = null, params: Params): Observable<boolean> | Promise<boolean> {
+    const enrolleeId = +params.eid;
+    if (enrolleeId) {
+      return this.paperEnrolmentResource.getEnrolleeById(enrolleeId)
+        .pipe(
+          map((enrollee: HttpEnrollee) => {
+            return this.routeDestination(routePath, enrollee, params);
```
mtpultz, line 42: You can definitely pass `enrolleeId` if that's what you want to do.  When you showed your code the other day I think `enrolleeId` was out of screen, but passing an `number` is easier to check for falsey if this code is expanded on later.

prime-angular-frontend/src/app/modules/paper-enrolment/shared/guards/finalized-enrolment.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+          })
+        );
+    } else {
+      return new Promise(async (resolve, reject) => resolve(true));
+    }
+  }
+
+  private routeDestination(routePath: string, httpEnrollee: HttpEnrollee, params) {
```
mtpultz, line 50: You'll want to add types to everything, otherwise the benefits of TypeScript over vanilla ES6 are lost.  Using types everywhere is like adding tests to your code.

prime-angular-frontend/src/app/modules/paper-enrolment/shared/guards/finalized-enrolment.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+    return false;
+  }
+
+  private getUrl(routeParam: UrlSegment[] | RouterStateSnapshot): string {
+    return (Array.isArray(routeParam))
+      ? routeParam.reduce((path, segment) => `${path}/${segment.path}`, '')
+      : routeParam.url;
+  }
+
+  private navigate(routePath: string, params): boolean {
```
mtpultz, line 67: You'll want to add types to everything

prime-angular-frontend/src/app/modules/paper-enrolment/shared/guards/finalized-enrolment.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+    const modulePath = this.config.routes.paperEnrolment;
+    const comparePath = `/${modulePath}/${params}/${routePath}`;
```
mtpultz, line 69: This might not work as you expect based on your use of `params`.  Does `comparePath` equal what you expect?

prime-angular-frontend/src/app/modules/paper-enrolment/shared/guards/finalized-enrolment.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  private routeDestination(routePath: string, httpEnrollee: HttpEnrollee, params) {
+    if (!httpEnrollee?.approvedDate) {
+      return true;
+    }
+    if (routePath.includes('demographic')) {
+      routePath = routePath.replace('demographic', '');
+    }
+    this.navigate(PaperEnrolmentRoutes.NEXT_STEPS, params);
```
mtpultz, line 57: You'll want to return the value of `this.navigate` to gate access, otherwise you're always returning false

prime-angular-frontend/src/app/modules/paper-enrolment/shared/guards/finalized-enrolment.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+    } else {
+      return new Promise(async (resolve, reject) => resolve(true));
+    }
+  }
+
+  private routeDestination(routePath: string, httpEnrollee: HttpEnrollee, params) {
+    if (!httpEnrollee?.approvedDate) {
+      return true;
+    }
+    if (routePath.includes('demographic')) {
```
mtpultz, line 54: Not sure of this section of code that replaces demographic.  Were you having an issue with the routing?

Paahn: /demographic was always getting appended to the end of routePath

mtpultz: You'll want to find out the reason for that instead curing the symptom.  If it is always getting appended without knowing more I would assume that you're entering the enrolment since that's the default route.

prime-angular-frontend/src/app/modules/paper-enrolment/shared/guards/paper-enrolment.guard.ts
```diff
@@ -26,29 +27,40 @@ export class PaperEnrolmentGuard extends BaseGuard {
   }
 
   protected checkAccess(routePath: string = null, params: Params): Observable<boolean> | Promise<boolean> {
-    const enrolleeId = params.eid;
-    if (!enrolleeId) {
+    const enrolleeId = +params.eid;
+    if (enrolleeId) {
```
mtpultz, line 31: To help compress the code a bit, and provide a bit of readability I would use a ternary:

```
return (enrolleeId)
  ? this.paperEnrolmentResource.getEnrolleeById(enrolleeId)
      .pipe(map((enrollee: HttpEnrollee) => this.routeDestination(routePath, enrollee, params)))
  : of(true);
```

prime-angular-frontend/src/app/modules/paper-enrolment/shared/guards/paper-enrolment.guard.ts
```diff
@@ -26,29 +27,40 @@ export class PaperEnrolmentGuard extends BaseGuard {
...
   }
 
   /**
    * @description
    * Determine the route destination based on the enrolment.
    */
-  private routeDestination(routePath: string, enrolment: HttpEnrollee) {
-    return !!enrolment;
+  private routeDestination(routePath: string, httpEnrollee: HttpEnrollee, params: Params) {
+    if (!httpEnrollee?.approvedDate) {
```
mtpultz, line 45: You can do the same thing with a ternary in this method as well, or use null coalesce:

```
return (httpEnrollee?.approvedDate)
  ? this.navigate(...)
  : true;
```

```
return !httpEnrollee?.approvedDate ?? this.navigate(...);
```

prime-angular-frontend/src/app/modules/paper-enrolment/shared/guards/paper-enrolment.guard.ts
```diff
@@ -26,29 +27,39 @@ export class PaperEnrolmentGuard extends BaseGuard {
...
+    return of(true);
   }
 
   /**
    * @description
    * Determine the route destination based on the enrolment.
    */
-  private routeDestination(routePath: string, enrolment: HttpEnrollee) {
-    return !!enrolment;
+  private routeDestination(routePath: string, httpEnrollee: HttpEnrollee, params: Params) {
```
mtpultz, line 44: Use `enrollee` as the param.  Normally you would follow the name of the type, but in this case `HttpEnrollee` is a temporary name of a model due to naming collision that we haven't had time to fix.

prime-angular-frontend/src/app/modules/paper-enrolment/shared/guards/paper-enrolment.guard.ts
```diff
@@ -26,29 +27,39 @@ export class PaperEnrolmentGuard extends BaseGuard {
...
+    if (!httpEnrollee?.approvedDate) {
+      return true;
+    } else {
+      return this.navigate(`/${this.config.routes.paperEnrolment}/${+params.eid}/${PaperEnrolmentRoutes.NEXT_STEPS}`, `${routePath}`, params);
```
mtpultz, line 48: Don't need to wrap `routePath` with ticks since it is already a string.  Similarly don't need to cast `eid` since it will be a string so you're converting it to a number then back to a string.

prime-angular-frontend/src/app/modules/paper-enrolment/shared/guards/paper-enrolment.guard.ts
mtpultz: Missing a return type of `params: Params): boolean {`

mtpultz (approved):

AnaisHebert (approved):
🦝🦝🦝🍕

PR #1424, opened 2021-07-08
PRIME-1655 Claim an existing organization as a signing authority

johnnyl6:


prime-dotnet-webapi/Startup.cs
```diff
@@ -76,13 +76,18 @@ public void ConfigureServices(IServiceCollection services)
...
+            services.AddScoped<IAuthorizedUserService, AuthorizedUserService>();
+            services.AddScoped<IEnrolleePaperSubmissionService, EnrolleePaperSubmissionService>();
+            services.AddScoped<IOrganizationClaimService, OrganizationClaimService>();
             services.AddScoped<IVerifiableCredentialService, VerifiableCredentialService>();
+            services.AddScoped<ISubmissionRulesService, SubmissionRulesService>();
+            services.AddScoped<ISubmissionService, SubmissionService>();
+            services.AddScoped<ISoapService, SoapService>();
+            services.AddScoped<ISiteService, SiteService>();
+            services.AddScoped<IRazorConverterService, RazorConverterService>();
+            services.AddScoped<IPrivilegeService, PrivilegeService>();
```
james-hollinger, line 90: I alphabetized these services in a PR yesterday, which is why you are seeing this diff.

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
+            var organization = await _organizationService.ClaimOrganizationAsync(claimOrganization.PartyId, claimOrganization.PEC, claimOrganization.ClaimDetail);
+
+            return Ok(organization);
+        }
+
+        // GET: api/Organizations/claim
+        /// <summary>
+        /// Gets organization claim count by a given search criteria.
+        /// </summary>
+        [HttpGet("claim", Name = nameof(GetOrganizationClaimCount))]
```
james-hollinger, line 151: Bit of a nuance here with nomenclature: the word "claim" can be either a verb or a noun. In REST, resources are generally nouns (that you do HTTP verbs to), they are collections, and are plural in the URL. Actions are verbs, are generally singular, and really only make sense with POST (how do you verb a verb?).

If there was only a POST endpoint that was performing the "claim" action, I could see these endpoints using the singular. However, if you are doing a GET against it, claim is now a _noun_, and you are getting a collection of them. It makes more sense to me to see these endpoints with "claim" being a noun; where we are creating, approving, and getting lists of them. To that end, I think these URLs should be /claims rather than /claim.

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
+            var result = await _organizationService.GetOrganizationClaimCountAsync(search);
+
+            return Ok(result);
+        }
+
+        // GET: api/Organizations/claim/5
+        /// <summary>
+        /// Find OrganizationClaim by Organization ID.
+        /// </summary>
+        [HttpGet("claim/{organizationId}", Name = nameof(GetOrganizationClaimByOrgId))]
```
james-hollinger, line 175: Based on my other comment, it makes more sense to think of claims as (noun) collections on Organizations, so this URL should be /organizations/{organizationId}/claims

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
+        // POST: api/Organizations/claim
+        /// <summary>
+        /// Claim an existing Organization.
+        /// </summary>
+        [HttpPost("claim", Name = nameof(ClaimOrganization))]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(typeof(ApiResultResponse<int>), StatusCodes.Status200OK)]
+
+        public async Task<ActionResult<Organization>> ClaimOrganization(ClaimOrganizationViewModel claimOrganization)
```
james-hollinger, line 135: Typing ActionResult this way probably does not work the way you might think. What is does is allow you to return _either_ any `ActionResult`, _or_ an object of the specified type. So theoretically with `ActionResult<Organization>` you could do 
```C#
return(organization);
```
rather than 
```C#
return Ok(organization);
```

But this is not actually whats happening: our API responses are wrapped in an `ApiResultResponse` object, so for this to be accurate it would have to be `public async Task<ActionResult<ApiResultResponse<Organization>>> ClaimOrganization`, which is really long and messy. And we aren't even using it, since we call `Ok(whatever)` everywhere.

So I have been pushing to just use `Task<ActionResult>` everywhere since its shorter and we aren't actually using the typed `ActionResult`s

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
+        {
+            if (!await _partyService.PartyExistsAsync(claimOrganization.PartyId, PartyType.SigningAuthority))
+            {
+                return BadRequest("Could not claim an organization, the passed in SigningAuthority does not exist.");
+            }
+
+            var organization = await _organizationService.ClaimOrganizationAsync(claimOrganization.PartyId, claimOrganization.PEC, claimOrganization.ClaimDetail);
+
+            return Ok(organization);
```
james-hollinger, line 144: I _very strongly_ feel this endpoint should not be returning the whole `Organization` back to the user. The annotation for this endpoint reads `[ProducesResponseType(typeof(ApiResultResponse<int>), StatusCodes.Status200OK)]`, was the original intention for this endpoint to return just the OrganizationId of the org? What does the FE actually require?

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(typeof(ApiResultResponse<int>), StatusCodes.Status200OK)]
+
+        public async Task<ActionResult<Organization>> ClaimOrganization(ClaimOrganizationViewModel claimOrganization)
+        {
+            if (!await _partyService.PartyExistsAsync(claimOrganization.PartyId, PartyType.SigningAuthority))
+            {
+                return BadRequest("Could not claim an organization, the passed in SigningAuthority does not exist.");
+            }
```
james-hollinger, line 140: In addition to checking that the Party exists, we also need to check that the current user is the party in the payload, else users could submit claims for other parties.

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
 
+        // POST: api/Organizations/claim
+        /// <summary>
+        /// Claim an existing Organization.
+        /// </summary>
+        [HttpPost("claim", Name = nameof(ClaimOrganization))]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(typeof(ApiResultResponse<int>), StatusCodes.Status200OK)]
+
```
james-hollinger, line 134: Please remove newline

prime-dotnet-webapi/Services/OrganizationService.cs
```diff
@@ -64,6 +67,15 @@ public async Task<IEnumerable<OrganizationSearchViewModel>> GetOrganizationsAsyn
                 .DecompileAsync()
                 .ToListAsync();
 
+            // TODO: Fix "A second operation started on this context before a previous operation completed." error
```
james-hollinger, line 70: Is this error still occurring?

prime-dotnet-webapi/Controllers/OrganizationsController.cs
johnnyl6: It is true that organization returned is not being used if it is a claim. This was a quick workaround after discussion with @mtpultz. The FE expects the Organization object to be returned to satisfy the `afterSubmitIsSuccessful` method of the OrganizationSigningAuthorityPage (inherits from `AbstractEnrolmentPage`).

> I _very strongly_ feel this endpoint should not be returning the whole `Organization` back to the user. The annotation for this endpoint reads `[ProducesResponseType(typeof(ApiResultResponse<int>), StatusCodes.Status200OK)]`, was the original intention for this endpoint to return just the OrganizationId of the org? What does the FE actually require?

prime-dotnet-webapi/ViewModels/SiteRegistration/OrganizationListViewModel.cs
```diff
@@ -18,6 +18,7 @@ public class OrganizationListViewModel
         public bool Completed { get; set; }
         public bool HasAcceptedAgreement { get; set; }
         public bool HasSubmittedSite { get; set; }
+        public bool IsUnderReview { get; set; }
```
james-hollinger, line 21: While the JIRA ticket says "the organization is under review", keep in mind the text in the ticket is often the product of brainstorming / imprecise language during ticket refinement. In my opinion, the organization is not under review in any way by having a claim on it. Calling it "IsUnderReview" is rather misleading and suggests that Organizations can be approved / have state.  I would prefer if this property was "HasClaim" or something similar. 
HasClaim"

prime-dotnet-webapi/Migrations/20210712052354_OrgClaimApprovalNotificationEmailTemplate.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DeleteData(
+                table: "EmailTemplate",
+                keyColumn: "Id",
+                keyValue: 14);
+        }
+    }
+}
```
james-hollinger, line 24: Please combine your migrations into one migration

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
+        public async Task<ActionResult<Organization>> ClaimOrganization(ClaimOrganizationViewModel claimOrganization)
+        {
+            if (!await _partyService.PartyExistsAsync(claimOrganization.PartyId, PartyType.SigningAuthority))
+            {
+                return BadRequest("Could not claim an organization, the passed in SigningAuthority does not exist.");
+            }
+
+            var organization = await _organizationService.ClaimOrganizationAsync(claimOrganization.PartyId, claimOrganization.PEC, claimOrganization.ClaimDetail);
```
james-hollinger, line 142: You can pass in the View Model here rather than pulling out the parts. This reduces the method signature as well as reducing churn if we have to add/remove parameters from the VM.

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
         }
 
+        // POST: api/Organizations/claim
+        /// <summary>
+        /// Claim an existing Organization.
+        /// </summary>
+        [HttpPost("claim", Name = nameof(ClaimOrganization))]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(typeof(ApiResultResponse<int>), StatusCodes.Status200OK)]
```
james-hollinger, line 133: This does not reflect what is being returned by the method.

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
             );
         }
 
+        // POST: api/Organizations/claim
+        /// <summary>
+        /// Claim an existing Organization.
+        /// </summary>
+        [HttpPost("claim", Name = nameof(ClaimOrganization))]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
```
james-hollinger, line 132: All of our (Authenticated) endpoints return 403 as well as 401, so the unfortunate bit of additional boilerplate is to add

```C#
```

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
+            // check if the organization exists with the given PEC
+            if (!string.IsNullOrEmpty(search.Pec))
+            {
+                var organization = await _organizationService.GetOrganizationByPecAsync(search.Pec);
+                if (organization == null)
+                {
+                    return BadRequest("Organization does not exist with the passed in site PEC.");
+                }
+            }
+            var result = await _organizationService.GetOrganizationClaimCountAsync(search);
```
james-hollinger, line 166: I am unsure why this is returning a count of claims rather than just if any exist? In both the places the frontend is currently using this endpoint it doesn't care about count.

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
+        [HttpGet("claim", Name = nameof(GetOrganizationClaimCount))]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(typeof(ApiResultResponse<OrganizationClaim>), StatusCodes.Status200OK)]
+        public async Task<ActionResult<int>> GetOrganizationClaimCount([FromQuery] OrganizationClaimSearchOptions search)
+        {
+            // check if the organization exists with the given PEC
+            if (!string.IsNullOrEmpty(search.Pec))
+            {
+                var organization = await _organizationService.GetOrganizationByPecAsync(search.Pec);
```
james-hollinger, line 160: This is fetching the entire Organization then throwing it away to see if it exists. Not a huge deal, but that's a bit wasteful. It feels like the `GetOrganizationClaimCountAsync` method could be written to return null if the org does not exist.

prime-dotnet-webapi/Services/OrganizationService.cs
neophyte57: No the error doesn't occur with the code as-is but I have a feeling the code could be improved.

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
+        {
+            var claim = await _organizationClaimService.GetOrganizationClaimAsync(organizationId);
+            if (claim == null)
+            {
+                return NotFound("No claim by a SigningAuthority exists for given Organization.");
+            }
+            else
+            {
+                return Ok(claim);
+            }
```
james-hollinger, line 190: Dont really need the `eslse` here

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,116 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
+        /// <summary>
+        /// Approve claim for an existing Organization.
+        /// </summary>
+        [HttpPost("claim/approve", Name = nameof(ApproveOrganizationClaim))]
+        [Authorize(Roles = Roles.EditSite)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        public async Task<IActionResult> ApproveOrganizationClaim(OrganizationClaimApprovalViewModel claimApproval)
```
james-hollinger, line 203: This endpoint makes more sense to me to not use a view model. Rather, The URL for this endpoint could be api/Organizations/{organizationId}/claims/{claimId}/approve

prime-dotnet-webapi/ViewModels/SiteRegistration/ClaimOrganizationViewModel.cs
```diff
@@ -0,0 +1,9 @@
+namespace Prime.ViewModels
+{
+    public class ClaimOrganizationViewModel
```
james-hollinger, line 3: Since "claim" reads more as a noun than a verb, this reads better as `OrganizationClaimViewModel`

prime-dotnet-webapi/Services/OrganizationService.cs
james-hollinger: If you change this back to 
```C#
var result = await _context.Organizations
``` 
does the error return? Or convention is to await `Task`s rather than do task.Result.

prime-dotnet-webapi/Controllers/OrganizationsController.cs
johnnyl6: The FE expects a BadRequest to be caught so it can't proceed to next step to claim the organization. If returning null, it means the ClaimOrganization does not exist instead. and the user can proceed. The alternative approach is to return a composite object containing both ClaimOrganizationExists & OrganizationExists which can be checked by the FE.

> This is fetching the entire Organization then throwing it away to see if it exists. Not a huge deal, but that's a bit wasteful. It feels like the `GetOrganizationClaimCountAsync` method could be written to return null if the org does not exist.

prime-dotnet-webapi/Services/OrganizationService.cs
neophyte57: I reverted line 55 and `/adjudication/site-registrations` loads fine.  But if I have the following code enabled, I get that error again.  Assuming it's related to https://docs.microsoft.com/en-us/ef/core/dbcontext-configuration/#asynchronous-operation-pitfalls but not sure how to correct it without using `task.Result`   

image

prime-dotnet-webapi/Startup.cs
neophyte57: I'll resolve conflicts

prime-dotnet-webapi/Services/OrganizationService.cs
james-hollinger: The error `A second operation started on this context before a previous operation completed.` suggests that multiple DB connections are being opened simultaneously. I strongly suspect the issue was the line 
```C#
results.ForEach(async r => r.IsUnderReview = await organizationClaimService.IsOrganizationUnderReviewAsync(r.Id));
```
which is opening a db connection for each organization being returned. The better way to do this is to add the mapping to the automapper configuration so the Property gets filled out when you do `.ProjectTo<OrganizationListViewModel>`, rather than doing another DB call for each item.

james-hollinger: There are a few ways you could do this, with minor tradeoffs between them. One way is to directly supply the `OrganizationClaims` table to Automapper:
```C#
// In AutoMapping.cs
IQueryable<OrganizationClaim> claims = null;
CreateMap<Organization, OrganizationListViewModel>()
    .ForMember(dest => dest.HasAcceptedAgreement, opt => opt.MapFrom(src => src.Agreements.Any(a => a.AcceptedDate.HasValue)))
    .ForMember(dest => dest.HasSubmittedSite, opt => opt.MapFrom(src => src.Sites.Any(s => s.SubmittedDate.HasValue)))
    .ForMember(dest => dest.Sites, opt => opt.MapFrom(src => src.Sites.Where(s => careSettingCode == null || s.CareSettingCode == careSettingCode)))
    .ForMember(dest => dest.HasClaim, opt => opt.MapFrom(src => claims.Any(c => c.OrganizationId == src.Id)));

// In OrganizationService.cs, line 66
.ProjectTo<OrganizationListViewModel>(_mapper.ConfigurationProvider, new { careSettingCode = searchOptions.CareSettingCode, claims = _context.OrganizationClaims })
```
This has the upside of not having to change any Models or View Models etc at the cost of making the mapping / fetching code a little uglier.

Another option would be to add the `Claims` navigation property to `Organization`:
```C#
// In AutoMapping.cs
CreateMap<Organization, OrganizationListViewModel>()
    .ForMember(dest => dest.HasAcceptedAgreement, opt => opt.MapFrom(src => src.Agreements.Any(a => a.AcceptedDate.HasValue)))
    .ForMember(dest => dest.HasSubmittedSite, opt => opt.MapFrom(src => src.Sites.Any(s => s.SubmittedDate.HasValue)))
    .ForMember(dest => dest.Sites, opt => opt.MapFrom(src => src.Sites.Where(s => careSettingCode == null || s.CareSettingCode == careSettingCode)))
    .ForMember(dest => dest.HasClaim, opt => opt.MapFrom(src => src.Claims.Any()));
    
// In Organization.cs
[JsonIgnore]
public ICollection<OrganizationClaim> Claims { get; set; }
```
This has the upside of being the simplest code change with the downside of adding more stuff to the `Organization` Model.

another option would be to take that last one and expand on it:
```C#
// In Organization.cs
[NotMapped]
[Computed]
public bool HasClaim
{
    get => Claims.Any();
}
```
This has the upside of not needing any change to Automapper at all, and it solves the problem of needing the property on both the `OrganizationListViewModel` and `Organization` itself (since we still return the whole `Organization` on GET organizations/{id} rather than a View Model). This has the downside of having *another* not mapped, computed property on our models. I would expect this would go away when we get a proper `OrganizationViewModel`.

My preference would be either option 2 or 3, fairly equally.

prime-dotnet-webapi/ViewModels/SiteRegistration/OrganizationClaimViewModel.cs
```diff
@@ -0,0 +1,9 @@
+namespace Prime.ViewModels
+{
+    public class OrganizationClaimViewModel
+    {
+        public int PartyId { get; set; }
+        public string PEC { get; set; }
+        public string ClaimDetail { get; set; }
```
james-hollinger, line 7: We added fluent validator a few PRs ago and have been very slowly adding validation to our View Models that represent incoming user data. I'd love to see some basic validation here :)

`PrivacyOfficeViewModel` was the first one to use it, the pattern would look something like:
```C#
using FluentValidation;

namespace Prime.ViewModels
{
    public class OrganizationClaimViewModel
    {
        public int PartyId { get; set; }
        public string PEC { get; set; }
        public string ClaimDetail { get; set; }
    }

    public class OrganizationClaimValidator : AbstractValidator<OrganizationClaimViewModel>
    {
        public OrganizationClaimValidator()
        {
            RuleFor(x => x.PartyId).NotEmpty();
            RuleFor(x => x.PEC).NotEmpty();
            RuleFor(x => x.ClaimDetail).NotEmpty();
        }
    }
}
```
You don't even have to register the validator anywhere, the app automatically finds all classes that inherit from `AbstractValidator` in the assembly due to line 99 in `Startup.cs`:
```C#
.AddFluentValidation(options => options.RegisterValidatorsFromAssemblyContaining<PrivacyOfficeValidator>())
```

prime-dotnet-webapi/Services/interfaces/IOrganizationService.cs
```diff
@@ -25,5 +28,13 @@ public interface IOrganizationService
...
         Task<SignedAgreementDocument> GetLatestSignedAgreementAsync(int organizationId);
         AgreementType OrgAgreementTypeForSiteSetting(int careSettingCode);
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="organizationId"></param>
+        /// <param name="newSigningAuthorityId"></param>
+        /// <returns><c>true</c> if can successfully switch signing authority, <c>false</c> otherwise.</returns>
+        Task<bool> SwitchSigningAuthorityAsync(int organizationId, int newSigningAuthorityId);
```
james-hollinger, line 38: I would prefer the comment be in the comment body, rather than in the return value of the Method

prime-dotnet-webapi/Services/interfaces/IOrganizationService.cs
```diff
@@ -15,7 +15,10 @@ public interface IOrganizationService
         Task<IEnumerable<OrganizationSearchViewModel>> GetOrganizationsAsync(OrganizationSearchOptions search);
         Task<IEnumerable<OrganizationListViewModel>> GetOrganizationsByPartyIdAsync(int partyId);
         Task<Organization> GetOrganizationAsync(int organizationId);
+        Task<Organization> GetOrganizationByPecAsync(string pec);
         Task<int> CreateOrganizationAsync(int signingAuthorityId);
+        Task<Organization> ClaimOrganizationAsync(OrganizationClaimViewModel claimOrganization);
+        Task<bool> OrganizationClaimExistsAsync(OrganizationClaimSearchOptions search);
```
james-hollinger, line 21: These two methods seem like they could be on the `OrganizationClamService`, rather than this ervice

prime-dotnet-webapi/Services/OrganizationClaimService.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace Prime.Services
+{
+    public class OrganizationClaimService : BaseService, IOrganizationClaimService
+    {
+        public OrganizationClaimService(
+            ApiDbContext context,
+            IHttpContextAccessor httpContext)
+            : base(context, httpContext) { }
+
+        public async Task<bool> DoesOrganizationHaveClaimAsync(int organizationId)
```
james-hollinger, line 14: Name can be simplified to `HasClaimAsync`

prime-dotnet-webapi/Services/OrganizationClaimService.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                return false;
+            }
+
+            _context.OrganizationClaims.Remove(claim);
+            int numAffected = await _context.SaveChangesAsync();
+            return (numAffected == 1);
+        }
+
+        public Task<Models.OrganizationClaim> GetOrganizationClaimAsync(int organizationId)
```
james-hollinger, line 35: I would prefer `using Prime.Models` at the top rather then in-lining the namespace here.

prime-dotnet-webapi/Services/OrganizationClaimService.cs
```diff
@@ -0,0 +1,42 @@
...
+            : base(context, httpContext) { }
+
+        public async Task<bool> DoesOrganizationHaveClaimAsync(int organizationId)
+        {
+            return await _context.OrganizationClaims
+                .AsNoTracking()
+                .AnyAsync(oc => oc.OrganizationId == organizationId);
+        }
+
+        public async Task<bool> DeleteClaimAsync(int organizationId, int partyId)
```
james-hollinger, line 21: The thing about delete(s) is that usually a successful delete and the thing not existing are the same. In the end, the thing doesn't exist. Does this need to report success/fail?

james-hollinger: Also, I feel it would be better /safer to delete by ClaimId, rather than OrdId + PartyId

prime-dotnet-webapi/Services/SiteService.cs
```diff
@@ -593,7 +593,7 @@ public async Task<IEnumerable<BusinessEvent>> GetSiteBusinessEventsAsync(int sit
         {
             return await _context.BusinessEvents
                 .Include(e => e.Admin)
-                .Where(e => e.SiteId == siteId && businessEventTypeCodes.Any(c => c == e.BusinessEventTypeCode))
+                .Where(e => e.SiteId == siteId && businessEventTypeCodes.Any(c => c == e.BusinessEventTypeCode) || e.Organization.Sites.Any(s => s.Id == siteId))
```
james-hollinger, line 596: Why was this change necessary?

prime-dotnet-webapi/Services/OrganizationService.cs
```diff
@@ -289,5 +357,12 @@ private IQueryable<Organization> GetBaseOrganizationQuery()
                     .ThenInclude(sa => sa.Addresses)
                         .ThenInclude(pa => pa.Address);
         }
+
+        public async Task<bool> SwitchSigningAuthorityAsync(int organizationId, int newSigningAuthorityId)
+        {
+            var organization = await GetOrganizationAsync(organizationId);
```
james-hollinger, line 363: We don't need the entire `Organization` here with all its related objects, you can just do 
```C#
var organization = await _context.Organizations
    .SingleAsync(o => o.Id == organizationId);
```

prime-dotnet-webapi/Services/OrganizationService.cs
```diff
@@ -118,6 +140,52 @@ public async Task<int> CreateOrganizationAsync(int partyId)
...
+            {
+                return false;
+            }
+
+            var userId = Guid.Empty;
+            Guid.TryParse(searchOptions.UserId, out userId);
+
+            return await _context.OrganizationClaims
+                .AsNoTracking()
+                .If(!string.IsNullOrEmpty(searchOptions.Pec), q => q
```
james-hollinger, line 180: These `.If()` statements are not needed, you have the guard clause on line 169 ensuring they are not null or empty

prime-dotnet-webapi/Services/OrganizationService.cs
```diff
@@ -118,6 +140,52 @@ public async Task<int> CreateOrganizationAsync(int partyId)
...
+        public async Task<bool> OrganizationClaimExistsAsync(OrganizationClaimSearchOptions searchOptions)
+        {
+            // return false if searchOptions is invalid
+            if (searchOptions == null || string.IsNullOrEmpty(searchOptions.Pec) && string.IsNullOrEmpty(searchOptions.UserId))
+            {
+                return false;
+            }
+
+            var userId = Guid.Empty;
+            Guid.TryParse(searchOptions.UserId, out userId);
```
james-hollinger, line 176: Rather than accepting a string and then trying to parse it into a `Guid`, you can make the type on the View Model a Guid

prime-dotnet-webapi/Services/SiteService.cs
neophyte57: Business events related to Organizations don't show up otherwise

prime-dotnet-webapi/Startup.cs
```diff
@@ -75,8 +75,9 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<IHealthAuthoritySiteService, HealthAuthoritySiteService>();
             services.AddScoped<ILookupService, LookupService>();
             services.AddScoped<IMetabaseService, MetabaseService>();
-            services.AddScoped<IOrganizationService, OrganizationService>();
             services.AddScoped<IOrganizationAgreementService, OrganizationAgreementService>();
+            services.AddScoped<IOrganizationClaimService, OrganizationClaimService>();
+            services.AddScoped<IOrganizationService, OrganizationService>();
```
james-hollinger, line 80: I decided to order the services alphabetically assuming the word "Service" is not included; the suffix "service" is a convention and I personally think about it more like "Organization Service" than an "OrganizationService" if that makes any sense.  Sounds a bit crazy when I write it out that way 😛.

This also has the side effect of "bubbling up" the "base" service to the top of the group:

DocumentService
DocumentAccessTokenService

EmailService
EmailDocumentsService
EmailRenderingService
EmailTemplateService

EnrolleeService
EnrolleeAgreementService
EnrolleePaperSubmissionService
EnrolleeSubmissionService

HealthAuthorityService
HealthAuthoritySiteService

OrganizationService
OrganizationAgreementService
OrganizationClaimService

SubmissionService
SubmissionRulesService

But if it makes more sense to you to do a strict alphabetical ordering I can see the merit.

neophyte57: Just thinking it's easier for someone to keep them ordered by e.g. copying and pasting into NotePad++ and doing a Edit > Line Operations > Sort ...

mtpultz:
@johnnyl6 @neophyte57 looking at PR we should consider moving the claim organization page after a user has created themselves as a signing authority.  It would simplify the the process by creating the user that will claim before requesting that information, and encapsulating functionality to specific pages instead of passing it forward.

prime-dotnet-webapi/Controllers/OrganizationsController.cs
james-hollinger: I suspect that this method's return type is going to change on further revision so I'm un-resolving the comment so I remember to check again later.

prime-dotnet-webapi/Services/OrganizationService.cs
johnnyl6: The guard is to ensure both PEC and userId cannot be empty at the same time, but one can be empty. When PEC is empty and userId is non-empty, we do not want the where clause to search PEC.

prime-angular-frontend/src/app/modules/site-registration/pages/organization-signing-authority-page/organization-signing-authority-page.component.ts
```diff
@@ -171,12 +179,20 @@ export class OrganizationSigningAuthorityPageComponent extends AbstractEnrolment
...
     if (redirectPath) {
       routePath = [redirectPath, SiteRoutes.SITE_REVIEW];
-    } else {
+    }
+    else if (this.route.snapshot.queryParams.claimOrg === 'true') {
+      if (organization) {
+        routePath = ['../../', SiteRoutes.ORGANIZATION_CLAIM_CONFIRMATION];
+      }
+      else {
+        this.toastService.openErrorToast('Could not claim the organization.');
```
mtpultz, line 188: The message indicating that it is already claimed should appear in the page, and not in a toast that disappears.  We should move the claim page after the signing authority is created so they already exist, and we don't need to use queryParams to indicate state or potentially lose data if they refresh.

prime-angular-frontend/src/app/modules/site-registration/pages/organization-claim-page/organization-claim-page.component.ts
```diff
@@ -0,0 +1,94 @@
...
+    private route: ActivatedRoute,
+    router: Router
+  ) {
+    this.title = this.route.snapshot.data.title;
+    this.routeUtils = new RouteUtils(route, router, SiteRoutes.MODULE_PATH);
+    this.isClaimExistingOrg = false;
+  }
+
+  public onBack(): void {
+    this.routeUtils.routeWithin(SiteRoutes.COLLECTION_NOTICE);
```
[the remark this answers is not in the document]
mtpultz: Moving the claim page after signing authority creation would have a back button to signing authority

prime-dotnet-webapi/Services/OrganizationService.cs
james-hollinger: ah, I see now, thanks

prime-angular-frontend/src/app/modules/site-registration/pages/organization-claim-page/organization-claim-page.component.ts
```diff
@@ -0,0 +1,94 @@
...
+            }
+          });
+      }
+      else {
+        // invalid form
+        return;
+      }
+    }
+    else {
+      this.routeUtils.routeRelativeTo([SiteRoutes.SITE_MANAGEMENT], { queryParams: { claimOrg: false } });
```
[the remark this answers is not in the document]
johnnyl6: Kind of still need the claimOrg=false explicitly since the guard is using it to determine the routes for first time visiter.  Thoughts?

In manageNoOrganizationRouting of OrganizationGuard
1. claimOrg=undefined, go to organization claim page
2. claimOrg query param exists (true or false), go to signing authority page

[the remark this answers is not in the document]
mtpultz: I would move the claim page to the other side of organization signing authority then the flow would be:

1. Access signing authority and create the signing authority, and then route to claim using initial route flow - `/organizations/0/organization-signing-authority`
2. Allow the now existing signing authority to claim an existing organization, otherwise route to organization name - `/organizations/0/claim`.  Using this route pattern will allow us to reuse the claim page in the future and not just on initial workflow.
3. Allow the signing authority to fill out the form, and on submit create the organization instead of creating it in the organization signing authority page - `/organizations/0/organization-name`.  Which once the organization is created would replace the oid with the proper oid from creation and carry on with registration.

We'll be allowing signing authority updates in the future, and this would keep each page with a single responsibility.  Also, we are claiming an organization so the pages should be within the organization routes.

prime-angular-frontend/src/app/modules/site-registration/site-registration-routing.module.ts
```diff
@@ -46,6 +48,16 @@ const routes: Routes = [
         component: CollectionNoticePageComponent,
         data: { title: 'Collection Notice' }
       },
+      {
+        path: SiteRoutes.ORGANIZATION_CLAIM,
```
mtpultz, line 52: With this relating to an organization this might be better located within the organization route loop.

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,117 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        public async Task<IActionResult> ApproveOrganizationClaim(int organizationId, int claimId)
+        {
+            var orgClaim = await _organizationClaimService.GetOrganizationClaimAsync(organizationId);
+            if (orgClaim == null || (orgClaim.Id != claimId))
+            {
+                return NotFound("Cannot locate Claim for given Organization.");
+            }
+            var userId = HttpContext.User.GetPrimeUserId();
```
james-hollinger, line 219: You dont need the `HttpContext` here, you only need `User.GetPrimeUserId()`

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,117 @@ public async Task<ActionResult<Organization>> CreateOrganization(CreateOrganizat
...
+
+            if (!await _organizationService.SwitchSigningAuthorityAsync(orgClaim.OrganizationId, orgClaim.NewSigningAuthorityId))
+            {
+                // TODO: Properly communicate error
+                return BadRequest("Could not assign Organization to new SigningAuthority.");
+            }
+
+            await _emailService.SendOrgClaimApprovalNotificationAsync(orgClaim);
+
+            await _businessEventService.CreateOrganizationEventAsync(orgClaim.OrganizationId, orgClaim.NewSigningAuthorityId, $"Organization Claim (Site ID/PEC provided: {orgClaim.ProvidedSiteId}, Reason: {orgClaim.Details}) approved by {admin.LastName}, {admin.FirstName}");
```
james-hollinger, line 230: I don't know if you really need to fetch the Admin here to add their name to the log; the business event has a direct reference to the Admin who created the event:
image

neophyte57: OK, agreed ... when I read description of https://bcgovmoh.atlassian.net/browse/PRIME-1655, I interpreted it differently at the time

prime-angular-frontend/src/app/modules/adjudication/pages/site-overview/site-overview.component.html
```diff
@@ -13,6 +13,35 @@
...
+                {{ orgClaim?.details }}
+              </ng-container>
+            </app-page-subheader>
+
+            <app-overview-section title="New Signing Authority"
+                                  (route)="onRoute($event)">
+              <app-bcsc-profile [user]="newSigningAuthority"></app-bcsc-profile>
+            </app-overview-section>
+
+            <!-- TODO: Change color and position properly -->
```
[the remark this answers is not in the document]
neophyte57: If "Decline" functionality is requested, can revisit this then

prime-dotnet-webapi/Services/SiteService.cs
james-hollinger: Looking through the code, it doesn't appear that we show organization events anywhere in our app currently.
I'm not sure that showing all of the Organization evens on each site is the best place for it, but it's not really in scope for this ticket to make a screen for just Organization events. As it stands, this is probably the best place for it for now.

However, this method now behaves in two unexpected ways:
1) It returns both Site events and Organization events despite being called `GetSiteBusinessEventsAsync` and being on the `SiteService`. This is super minor, but maybe renaming it to `GetSiteRelatedBusinessEventsAsync` or something like that might be more clear? Makes the name longer though. Maybe add a short comment?

2) This service accepts a list of `BusinessEventType` codes to filter the return, but Organization Events wont be filtered out due to the `&&` operation having precedence over the `||` operation. This method currently returns events that 
 a) Match the site Id and are in the list of codes.
 or
 b) are org events on the site's parent org

to keep the filtering behavior, the line can be changed to :
```C#
.Where(e => businessEventTypeCodes.Any(c => c == e.BusinessEventTypeCode) && (e.SiteId == siteId || e.Organization.Sites.Any(s => s.Id == siteId))
```

prime-dotnet-webapi/Services/interfaces/IOrganizationClaimService.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        Task<bool> HasClaimAsync(int organizationId);
+
+        /// <summary>
+        /// Returns <c>true</c> if can successfully delete claim, <c>false</c> otherwise.
+        /// </summary>
+        Task<bool> DeleteClaimAsync(int claimId);
+
+        Task<OrganizationClaim> GetOrganizationClaimAsync(int organizationId);
+        Task<Organization> ClaimOrganizationAsync(OrganizationClaimViewModel claimOrganization, Organization organization);
```
james-hollinger, line 19: This method's name is a bit misleading; it sounds like this method might re-assign the Signing Authority. Could you please change it to `CreateOrganizationClaimAsync`?

prime-dotnet-webapi/Startup.cs
```diff
@@ -77,10 +77,10 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<IMetabaseService, MetabaseService>();
             services.AddScoped<IOrganizationService, OrganizationService>();
             services.AddScoped<IOrganizationAgreementService, OrganizationAgreementService>();
+            services.AddScoped<IOrganizationClaimService, OrganizationClaimService>();
             services.AddScoped<IPartyService, PartyService>();
             services.AddScoped<IPdfService, PdfService>();
             services.AddScoped<IPlrProviderService, PlrProviderService>();
-            services.AddScoped<ISubmissionRulesService, SubmissionRulesService>();
```
james-hollinger, line 83: Thanks for catching this

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,115 @@ public async Task<ActionResult> CreateOrganization(CreateOrganizationViewModel c
             );
         }
 
+        // POST: api/Organizations/claim
+        /// <summary>
+        /// Claim an existing Organization.
+        /// </summary>
+        [HttpPost("claim", Name = nameof(ClaimOrganization))]
```
james-hollinger, line 130: As before, we are creating `Claim`s, so this URL should be /claims

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,115 @@ public async Task<ActionResult> CreateOrganization(CreateOrganizationViewModel c
...
+            if (organization == null)
+            {
+                return BadRequest("Could not claim an organization, the passed in PEC did not locate an organization.");
+            }
+            organization = await _organizationClaimService.CreateOrganizationClaimAsync(organizationClaim, organization);
+
+            return Ok(organization);
+        }
+
+        // GET: api/Organizations/claim
```
james-hollinger, line 158: claims

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,115 @@ public async Task<ActionResult> CreateOrganization(CreateOrganizationViewModel c
...
+        /// </summary>
+        [HttpPost("{organizationId}/claims/{claimId}/approve", Name = nameof(ApproveOrganizationClaim))]
+        [Authorize(Roles = Roles.EditSite)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        public async Task<ActionResult> ApproveOrganizationClaim(int organizationId, int claimId)
+        {
+            var orgClaim = await _organizationClaimService.GetOrganizationClaimAsync(organizationId);
```
james-hollinger, line 214: I don't see why you would have to get the claim by organizationId when you have the claimId already. I could see passing both IDs into the service method to make sure that the specified claim is actually on the specified organization.

prime-dotnet-webapi/Services/OrganizationClaimService.cs
```diff
@@ -0,0 +1,89 @@
...
+            var claim = await _context.OrganizationClaims
+                .SingleOrDefaultAsync(oc => oc.Id == claimId);
+            if (claim == null)
+            {
+                return false;
+            }
+
+            _context.OrganizationClaims.Remove(claim);
+            int numAffected = await _context.SaveChangesAsync();
+            return (numAffected == 1);
```
james-hollinger, line 43: You do not need to surround boolean expressions with parentheses like this

prime-dotnet-webapi/Controllers/OrganizationsController.cs
johnnyl6: > I suspect that this method's return type is going to change on further revision so I'm un-resolving the comment so I remember to check again later.

changed the return type to int (organizationId)

prime-angular-frontend/src/app/modules/site-registration/shared/guards/organization.guard.ts
```diff
@@ -37,22 +38,24 @@ export class OrganizationGuard extends BaseGuard {
...
         ),
-        map((organizations: Organization[] | null) =>
-          (organizations?.length)
-            ? organizations.shift()
-            : null
+        map((result: [Organization[] | null, Party, boolean]) =>
+          [result[0]?.length ? result[0].shift() : null, result[1], result[2]]
         ),
-        map((organization: Organization | null) => {
+        map((result: [Organization | null, Party, boolean]) => {
```
mtpultz, line 50: Use array destructuring instead of array indexing:

```
map(({ organization, party, claimed }: [Organization | null, Party, boolean]) => ...
```

prime-angular-frontend/src/app/modules/site-registration/shared/guards/organization.guard.ts
```diff
@@ -37,22 +38,24 @@ export class OrganizationGuard extends BaseGuard {
...
           // Store the organization for access throughout registration, which
           // will allows be the most up-to-date organization
-          this.organizationService.organization = organization;
+          this.organizationService.organization = result[0];
```
mtpultz, line 53:
```
this.organizationService.organization = organization;
```

prime-angular-frontend/src/app/modules/site-registration/shared/guards/organization.guard.ts
```diff
@@ -37,22 +38,24 @@ export class OrganizationGuard extends BaseGuard {
...
 
           // Determine the next route based on whether this is the initial
           // registration of an organization and site, or subsequent
           // registration of sites under an existing organization
-          return this.routeDestination(routePath, params, organization);
+          return this.routeDestination(routePath, params, result[0], result[1], result[2]);
```
mtpultz, line 58:
```
return this.routeDestination(routePath, params, organization, party, claimed);
```

prime-angular-frontend/src/app/modules/site-registration/shared/guards/organization.guard.ts
```diff
@@ -127,10 +130,14 @@ export class OrganizationGuard extends BaseGuard {
    * Manage routing when an organization does not exist, or initial
    * registration has not been completed.
    */
-  private manageNoOrganizationRouting(routePath: string) {
+  private manageNoOrganizationRouting(routePath: string, party: Party, hasOrgClaim: boolean) {
+    const destPath = hasOrgClaim ? SiteRoutes.ORGANIZATION_CLAIM_CONFIRMATION :
```
mtpultz, line 134: When using ternaries unless they are really short they should always be structured vertically for readability:

```
const destPath = (claimed) 
  ? SiteRoutes.ORGANIZATION_CLAIM_CONFIRMATION 
  : (party) 
    ? SiteRoutes.ORGANIZATION_CLAIM 
    : SiteRoutes.ORGANIZATION_SIGNING_AUTHORITY;
```

prime-angular-frontend/src/app/modules/site-registration/pages/organization-claim-page/organization-claim-page.component.ts
```diff
@@ -0,0 +1,116 @@
...
+          if (this.isClaimExistingOrg) {
+            return this.organizationResource.claimOrganization(party.id, this.formState.json)
+              .pipe(
+                catchError((error, caught) => {
+                  this.hasOrgClaimError = true;
+                  return of(0);
+                })
+            );
+          }
+          return this.organizationResource.createOrganization(party.id)
```
mtpultz, line 88: We want to create the organization just in time.  You'll want to put organization creation on the OrganizationNamePage and have it create if the `oid` param doesn't exist (zero aka falsey), or update if the `oid` param exists (not zero aka truthy).

prime-angular-frontend/src/app/modules/site-registration/site-registration-routing.module.ts
```diff
@@ -67,6 +69,18 @@ const routes: Routes = [
                 canDeactivate: [CanDeactivateFormGuard],
                 data: { title: 'Signing Authority' }
               },
+              {
+                path: SiteRoutes.ORGANIZATION_CLAIM,
+                component: OrganizationClaimPageComponent,
+                canDeactivate: [CanDeactivateFormGuard],
+                data: { title: 'Create or Claim Organization' }
```
mtpultz, line 76: We don't need to indicate the organization creation in the title.  Also, we want to perform creation on OrganizationNamePage.

prime-angular-frontend/src/app/modules/site-registration/pages/organization-claim-page/organization-claim-page.component.html
```diff
@@ -0,0 +1,76 @@
+<app-page [busy]="busy">
+  <app-page-header>PharmaNet Site Registration</app-page-header>
+  <app-site-progress-indicator [inProgress]="!isCompleted"></app-site-progress-indicator>
+
+  <section class="mb-3">
```
mtpultz, line 5: You can update `<section class="mb-3">` to a component `<app-page-section>`

prime-angular-frontend/src/app/modules/site-registration/pages/organization-claim-page/organization-claim-page.component.html
```diff
@@ -0,0 +1,68 @@
...
+      Claim an existing organization.
+    </mat-slide-toggle>
+  </section>
+
+  <form *ngIf="claimExistingOrg.checked"
+        [formGroup]="formState.form"
+        (ngSubmit)="onContinue()"
+        novalidate
+        autocomplete="off">
+    <section class="mb-3">
```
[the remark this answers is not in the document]
mtpultz: Looks like you resolve this, but didn't make the change.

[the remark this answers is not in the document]
mtpultz: You can update `<section class="mb-3">` to a component `<app-page-section>`

prime-angular-frontend/src/app/modules/site-registration/pages/organization-claim-confirmation-page/organization-claim-confirmation-page.component.html
```diff
@@ -0,0 +1,27 @@
...
+    <app-alert type="success"
+               icon="check_circle_outline">
+      <ng-container #alertTitle
+                    class="alert-title">
+        You have claimed an existing Organization
+      </ng-container>
+    </app-alert>
+  </div>
+
+  <section class="mb-3">
```
mtpultz, line 15: You can update `<section class="mb-3">` to a component `<app-page-section>`

prime-dotnet-webapi/Startup.cs
neophyte57: Added comment re: ordering convention

prime-dotnet-webapi/Services/OrganizationService.cs
neophyte57: @james-hollinger , thanks for the ideas.  It became something like Options 2 and 3.  Please have a look.

prime-dotnet-webapi/Services/OrganizationClaimService.cs
```diff
@@ -0,0 +1,106 @@
...
+            return numAffected == 1;
+        }
+
+        public Task<OrganizationClaim> GetOrganizationClaimAsync(int claimId)
+        {
+            return _context.OrganizationClaims
+                .SingleOrDefaultAsync(oc => oc.Id == claimId);
+        }
+
+        public Task<OrganizationClaim> GetOrganizationClaimByOrgIdAsync(int organizationId)
```
james-hollinger, line 52: These two methods are not marked `async`, and are instead returning an awaitable Task rather than awaiting the query and returning the value. There is a really interesting discussion to be had here on the nuance of each; was this done on purpose or was it an oversight?

prime-dotnet-webapi/Services/OrganizationService.cs
```diff
@@ -287,7 +301,15 @@ private IQueryable<Organization> GetBaseOrganizationQuery()
...
                 .Include(o => o.SigningAuthority)
                     .ThenInclude(sa => sa.Addresses)
-                        .ThenInclude(pa => pa.Address);
+                        .ThenInclude(pa => pa.Address)
+                .Include(o => o.Claims);
+        }
+
+        public async Task<bool> SwitchSigningAuthorityAsync(int organizationId, int newSigningAuthorityId)
+        {
+            var organization = await _context.Organizations.Where(o => o.Id == organizationId).SingleAsync();
```
james-hollinger, line 310: Because this pattern of doing `.Where(predicate)` followed by an execution statement is so common, these LINQ methods
```C#
AnyAsync()
FirstAsync()
FirstOrDefaultAsync()
SingleAsync()
SingleOrDefaultAsync()
```

all have an overloaded version that accepts the filtering predicate to simplify the code. So this line can be
```C#
var organization = await _context.Organizations
    .SingleAsync(o => o.Id == organizationId);
```

prime-dotnet-webapi/Services/OrganizationClaimService.cs
neophyte57: Yeah, let's talk a bit later today  :-)

prime-dotnet-webapi/Services/OrganizationService.cs
```diff
@@ -60,6 +63,7 @@ public async Task<IEnumerable<OrganizationSearchViewModel>> GetOrganizationsAsyn
                         o => o.Sites.Select(s => s.PEC))
                     .Containing(searchOptions.TextSearch)
                 )
+                .Include(o => o.Claims)
```
james-hollinger, line 66: This might be surprising, but you don't need to `.Include()` the Claims here for `HasClaim` to appear on the View Model. I think it's worth knowing why it works this way so you can have a better understanding of how AutoMapper + Entity Framework works.

As a baseline, instead of using AutoMapper, we could project to a View Model by doing a `.Select()` statement. Entity Framework takes that `.Select()` statement and uses it to generate SQL; so the `.ProjectTo()` could be re-written without AutoMapper by doing:
```C#
.Select(org => new OrganizationListViewModel
{
    Id = org.Id,
    Name = org.Name,
    ...
    HasClaim = org.Claims.Any()
})
```
EF knows that the `Claims` property is on another table, so it will automatically generate the SQL it needs. Sometimes this is a join, but in this case the SQL is something like 
```SQL
SELECT EXISTS (
    SELECT 1
    FROM "OrganizationClaim" AS oc
    WHERE o."Id" = oc."OrganizationId") AS "HasClaim"
FROM "Organization" AS o
```
Adding in AutoMapper: the mapping configuration, by default, copies Properties if their name and type match. So AM sees that the View Model has a Property called `HasClaim`, and the Model has a matching property, so it generates
```C#
.Select(org => new OrganizationListViewModel
{
    Id = org.Id,
    Name = org.Name,
    ...
    HasClaim = org.HasClaim
})
```
But this wont work; EF doesn't know how to look inside the `HasClaim` property since it is effectively a black box. This is where Delegate Decompiler comes in: calling `.DecompileAsync()` turns all the `[Computed]` Properties into their expressions, making the above `.Select()` into:
```C#
.Select(org => new OrganizationListViewModel
{
    Id = org.Id,
    Name = org.Name,
    ...
    HasClaim = org.Claims.Any()
})
```
as we had before.

In contrast, using `.Include()` does an eager load and will, usually, do a full SELECT * + JOIN to the other table. It is kinda in direct opposition to the View Models we are building: VMs declare the (often flattened) subset of Properties the we are interested in, while `.Include()` loads everything. In general, if we are projecting to a VM, we don't need (or want) to use `.Include()`.

neophyte57: FYI @james-hollinger , I added a link to this conversation into my OneNote collection.  Plus we are 1 more comment closer to triple digit comments on this Pull Request!  😊

prime-angular-frontend/src/app/modules/adjudication/pages/site-overview/site-overview.component.ts
```diff
@@ -79,20 +84,40 @@ export class SiteOverviewComponent extends SiteRegistrationContainerComponent im
...
-      organization: this.organizationResource.getOrganizationById(oid),
-      site: this.siteResource.getSiteById(sid)
-    }).subscribe(({ organization, site }) => {
-      this.organization = organization;
-      this.site = site;
-      this.form.get('pec').setValue(site.pec);
+    this.busy = forkJoin([
+      this.organizationResource.getOrganizationById(oid),
+      this.siteResource.getSiteById(sid),
+      this.organizationResource.getOrganizationClaimByOrgId(oid)
```
james-hollinger, line 109: HasClaim is already on the organization model returned on line 107, you don't need this call.

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,112 @@ public async Task<ActionResult> CreateOrganization(CreateOrganizationViewModel c
...
+        // GET: api/Organizations/5/claims
+        /// <summary>
+        /// Find OrganizationClaim by Organization ID.
+        /// </summary>
+        [HttpGet("{organizationId}/claims", Name = nameof(GetOrganizationClaimByOrgId))]
+        [Authorize(Roles = Roles.ViewSite)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiResultResponse<OrganizationClaim>), StatusCodes.Status200OK)]
+        public async Task<ActionResult> GetOrganizationClaimByOrgId(int organizationId)
```
james-hollinger, line 189: As alluded to in the previous comment, we don't need this endpoint.

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -110,6 +123,112 @@ public async Task<ActionResult> CreateOrganization(CreateOrganizationViewModel c
...
+        [ProducesResponseType(typeof(ApiResultResponse<int>), StatusCodes.Status200OK)]
+        public async Task<ActionResult> ClaimOrganization(OrganizationClaimViewModel organizationClaim)
+        {
+            var party = await _partyService.GetPartyAsync(organizationClaim.PartyId, PartyType.SigningAuthority);
+            if (party == null)
+            {
+                return BadRequest("Could not claim an organization, the passed in SigningAuthority does not exist.");
+            }
+
+            if (party.UserId != HttpContext.User.GetPrimeUserId())
```
james-hollinger, line 143: You dont need the `HttpContext` here, you only need User.GetPrimeUserId()

prime-angular-frontend/src/app/modules/adjudication/pages/site-overview/site-overview.component.ts
neophyte57: We still need the claim details

AnaisHebert (approved):
Lgtm 🦆

prime-angular-frontend/src/app/modules/site-registration/pages/organization-claim-page/organization-claim-page.component.ts
```diff
@@ -0,0 +1,112 @@
...
+                  this.hasOrgClaimError = true;
+                  return of(false);
+                })
+          ))
+        )
+    }
+    return of(true);
+  }
+
+  protected afterSubmitIsSuccessful(success: boolean): void {
```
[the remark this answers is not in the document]
mtpultz:
```ts
  protected performSubmission(): NoContent {
    return (this.isClaimExistingOrg)
      ? this.authService.getUser$()
        .pipe(
          exhaustMap((bcscUser: BcscUser) => this.organizationResource.getSigningAuthorityByUserId(bcscUser.userId)),
          exhaustMap((party: Party) =>
            this.organizationResource.claimOrganization(party.id, this.formState.json)
              .pipe(
                catchError(_ => {
                  this.hasOrgClaimError = true;
                  return EMPTY;
                })
              )),
          NoContentResponse
        )
      : of(NoContentResponse);
  }
```

prime-angular-frontend/src/app/modules/site-registration/pages/organization-claim-page/organization-claim-page.component.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  protected createFormInstance(): void {
+    this.formState = this.organizationFormStateService.organizationClaimPageFormState;
+  }
+
+  protected patchForm(): void {
+  }
+
+  protected performSubmission(): Observable<boolean> {
```
mtpultz, line 74: I'm getting this error on submission of a claim for an approved site on the PR.  A site with a PEC of 12345 was submitted by PRIMET003 and has been approved, but PRIMET004 (other test card) can't claim the organization:

image

prime-dotnet-webapi/Controllers/OrganizationsController.cs
mtpultz: @johnnyl6 @neophyte57 has this been addressed?  Trying to resolve remaining issues so we can get this PR out the door.

neophyte57: As per the other conversation, we still need this endpoint:

image

prime-angular-frontend/src/app/modules/site-registration/pages/organization-claim-page/organization-claim-page.component.ts
neophyte57: As discussed on Zoom, the (data) problem is that there are 3 sites, each belonging to a separate organization, that have the same Site Id "12345"

image

mtpultz: https://bcgovmoh.atlassian.net/browse/PRIME-1762

prime-angular-frontend/src/app/modules/site-registration/pages/organization-signing-authority-page/organization-signing-authority-page.component.html
```diff
@@ -200,9 +200,8 @@
   </form>
 
   <app-page-footer [isInitialEnrolment]="!isCompleted"
-                   [hasSecondaryAction]="true"
+                   [hasSecondaryAction]="false"
```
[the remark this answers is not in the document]
johnnyl6: I think you referred to the wrong line, this is back button leading to collection of notice, which you'd like to remove previously. 

> We'll want a back button to allow updates to the signing authority if they made a mistake prior to reaching the first site creation pages. You should be able to move forwards and backwards between Signing Authority, Claim Organization, and Organization Name pages like before, but with the addition of Claim Organization.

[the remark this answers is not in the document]
mtpultz: Thanks @johnnyl6 you're absolutely right.  This comment should have been on the claim page.

mtpultz (approved):

PR #1428, opened 2021-07-09
Feature/prime 1325 frontend logging errors to web api

hogmann:


neophyte57 (commented):
@hogmann I haven't investigated why but when the web-api starts up, it immediately receives many calls to the error API.  See screen-shots.  

image

image

This is when testing locally ... I wanted to see the behavior in the PR environment

prime-dotnet-webapi/ViewModels/Logging/FrontEndLogViewModel.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using System.Collections.Generic;
+using Prime.Models;
+
+namespace Prime.ViewModels
+{
+    public class FrontEndLogViewModel
+    {
+        public string Msg { get; set; }
+        public string Data { get; set; }
```
james-hollinger, line 10: Seems a bit simpler to have the log type in the view model rather than as a route parameter.

prime-dotnet-webapi/Services/LogService.cs
```diff
@@ -0,0 +1,29 @@
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Prime.Models;
+using Prime.ViewModels;
+
+namespace Prime.Services
+{
+    public class LogService : BaseService, ILogService
```
james-hollinger, line 8: Perhaps we should name this something other than LogService so it doesn't get confused with a generic logger that the backend uses. Perhaps `ClientLogService`?

neophyte57:
> @hogmann I haven't investigated why but when the web-api starts up, it immediately receives many calls to the error API. See screen-shots.
> 
> image
> 
> image
> 
> This is when testing locally ... I wanted to see the behavior in the PR environment

Never mind, not happening in PR environment.

prime-dotnet-webapi/Models/Logging/FrontEndLog.cs
```diff
@@ -0,0 +1,22 @@
...
+        [Key]
+        public int Id { get; set; }
+
+        [Required]
+        public string Msg { get; set; }
+
+        public string Data { get; set; }
+
+        [Required]
+        public LogType LogType { get; set; }
```
james-hollinger, line 20: It feels a little redundant, but I would actually prefer to see a Date on the log separate from the audit columns using a UTC `DateTime` rather than `DateTimeOffset`.

prime-dotnet-webapi/Models/Logging/FrontEndLog.cs
```diff
@@ -0,0 +1,22 @@
...
+
+    [Table("FrontEndLog")]
+    public class FrontEndLog : BaseAuditable
+    {
+        [Key]
+        public int Id { get; set; }
+
+        [Required]
+        public string Msg { get; set; }
+
```
neophyte57, line 16: Can this be `Message` instead?

prime-angular-frontend/src/app/core/services/logger.service.ts
```diff
@@ -1,12 +1,15 @@
 import { Injectable } from '@angular/core';
+import { LoggerResource } from '@core/resources/logger-resource.service';
 
 import { environment } from '@env/environment';
 
 @Injectable({
   providedIn: 'root'
 })
 export class LoggerService {
```
neophyte57, line 9: @hogmann Not related to your changes exactly but @mtpultz , can we get rid of the `log` and `trace` methods?  Don't appear to be used and they are confusing.

neophyte57 (changes requested):

prime-dotnet-webapi/Services/LogService.cs
neophyte57: Would `FrontEndLogService` be better because in case the Angular client isn't the only client of PRIME's web-api?

prime-dotnet-webapi/ViewModels/Logging/FrontEndLogViewModel.cs
hogmann: The reason i chose this path is it reduced the work and code needed in the front end. Since i'm just using the default object generated in our logging service and sending that up as a sudo view model. This also reflects the naming Msg/Data.

prime-dotnet-webapi/Models/Logging/FrontEndLog.cs
hogmann: Also seems extra redundant since the data is only accessible through the DB. But can throw in there so down the road we don't have to do a tiny data migration.

prime-angular-frontend/src/app/core/services/logger.service.ts
hogmann: Going to have to resolve.

prime-dotnet-webapi/Services/ClientLogService.cs
```diff
@@ -0,0 +1,32 @@
...
+        { }
+
+        public async Task CreateLogAsync(ClientLogViewModel log)
+        {
+            var newLog = new ClientLog
+            {
+                LogType = log.LogType,
+                Msg = log.Message,
+                Data = log.Data,
+                CreatedDate = DateTime.Now
```
james-hollinger, line 24: Please use `DateTime.UtcNow` here

prime-dotnet-webapi/Startup.cs
```diff
@@ -72,6 +72,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<IEnrolmentCertificateService, EnrolmentCertificateService>();
             services.AddScoped<IGisService, GisService>();
             services.AddScoped<IHealthAuthorityService, HealthAuthorityService>();
+            services.AddScoped<IClientLogService, ClientLogService>();
```
james-hollinger, line 75: In a PR last week I organized service declarations in alphabetical order, please adjust

prime-dotnet-webapi/ApiDbContext.cs
```diff
@@ -51,6 +51,8 @@ IHttpContextAccessor context
             _context = context;
         }
 
+        public DbSet<ClientLog> FrontEndLogs { get; set; }
```
james-hollinger, line 54: I assume this was just missed in the rename

prime-dotnet-webapi/Controllers/ClientLogsController.cs
```diff
@@ -0,0 +1,37 @@
...
+        // POST /api/client-logs
+        /// <summary>
+        /// Creates a client log
+        /// </summary>
+        [HttpPost("", Name = nameof(CreateLog))]
+        [AllowAnonymous]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        public async Task<ActionResult> CreateLog(ClientLogViewModel log)
+        {
+            await _logService.CreateLogAsync(log);
```
james-hollinger, line 33: normally I would never advocate for a `try ... catch` to swallow all exceptions, but I think it might be warranted here. If the frontend errors we probably don't want to make the situation worse by accidentally returning another error.

prime-dotnet-webapi/Startup.cs
hogmann: Yeah forgot to shift after name change

prime-dotnet-webapi/Controllers/ClientLogsController.cs
hogmann: So i went the route more that this call on te front-end won't trigger itself, but can double down here too.

james-hollinger (dismissed):
LGTM

hogmann:
@neophyte57 you have requested a change but as far as im aware i have fixed what you referenced.

neophyte57:
> @neophyte57 you have requested a change but as far as im aware i have fixed what you referenced.

Sorry for delay, will look again today

prime-angular-frontend/src/app/core/services/abstract-logger.service.ts
```diff
@@ -0,0 +1,52 @@
...
+    this.send('error', { msg, data });
+  }
+
+  /**
+   * @description
+   * Outputs a stack trace.
+   */
+  public trace(msg: string, ...data: any[]) {
+    this.send('error', { msg, data });
+  }
```
neophyte57, line 48: `trace` and `log` are not referenced anywhere ... do we still need these?

neophyte57:
I'm getting some weird behavior locally as in the PR environment, the database migration didn't run properly  
image

neophyte57:
My test is to try to go to the Banner pages without the required Keycloak permissions.  It logs the error the first time but not subsequent visits to http://localhost:4200/adjudication/enrollees/banners or http://localhost:4200/adjudication/site-registrations/banners

image

prime-angular-frontend/src/app/core/services/http-logger.service.ts
```diff
@@ -0,0 +1,21 @@
...
+
+@Injectable({
+  providedIn: 'root'
+})
+export class HttpLoggerService extends AbstractLoggerService {
+  constructor(
+    private loggerResource: LoggerResource,
+  ) {
+    super();
+  }
```
neophyte57, line 16: Would this be better named `WebApiLoggerService`?

hogmann:
> I'm getting some weird behavior locally as in the PR environment, the database migration didn't run properly

Ill have to recreate the PR enviroment i wiped the migrations and remade and haven't relaunched.

prime-angular-frontend/src/app/core/services/http-logger.service.ts
hogmann: I have no reserves for either name. Both seem to get the point across. Is there a reason you feel WebApi is better suited?

prime-angular-frontend/src/app/core/services/abstract-logger.service.ts
hogmann: Just cloned the code from the previous logger. If they aren't called I can remove

prime-angular-frontend/src/app/core/services/http-logger.service.ts
neophyte57: For `WebApiLoggerService`, I think it's more clear where the logs are going to.  E.g. `ConsoleLoggerService`, the logs are going to the browser console

AnaisHebert (approved):
🦝🦝🦝

Paahn (approved):
all for that

PR #1443, opened 2021-07-14
Unify ActionResult

james-hollinger:
There are three options for the `ActionResult` returned from our controller methods:
```C#
ActionResult
ActionResult<Result>
IActionResult
```

In the beginning of our app, we typed all of our `ActionResult`s because that's what we thought we had to do. Turns out, typing `ActionResult` does not work the way one might intuitively assume. It does **not** specify that the method will return an object of that type.  What is does is allow you to return *either* an `ActionResult`, *or* an object of the specified type. So theoretically with `ActionResult<Enrollee>` we could do
```C#
return(enrollee);
```
rather than
```C#
return Ok(enrollee);
```

But this is not actually whats happening: our API responses are all wrapped in an `ApiResultResponse` object, so for this to be accurate it would have to be `ActionResult<ApiResultResponse<Enrollee>>`, which is really long and messy. And we aren't even using it, since we call Ok(whatever) everywhere.

I have been pushing to just use `Task<ActionResult>` everywhere since it's shorter, we aren't actually using the typed `ActionResult`s for their intended purpose, and we have helper methods that automatically wrap the return object in `ApiResultResponse`. To that end, here is a PR changing them all.

### DIFF
--- a/prime-dotnet-webapi/Controllers/AdminsController.cs
+++ b/prime-dotnet-webapi/Controllers/AdminsController.cs
@@ -35,7 +35,7 @@ namespace Prime.Controllers
         [ProducesResponseType(StatusCodes.Status403Forbidden)]
         [ProducesResponseType(typeof(ApiResultResponse<Admin>), StatusCodes.Status200OK)]
         [ProducesResponseType(typeof(ApiResultResponse<Admin>), StatusCodes.Status201Created)]
-        public async Task<ActionResult<Admin>> CreateAdmin(Admin admin)
+        public async Task<ActionResult> CreateAdmin(Admin admin)
         {
             if (admin == null)
             {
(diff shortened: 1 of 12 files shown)


mtpultz (approved):

neophyte57 (approved):
I ran the e2e tests against the PR environment.  Although they don't provide full coverage yet, nothing failed.  Somehow this "flexibility" and our misunderstanding seems like it's due to a deficiency in the ASP.NET Core API

james-hollinger:
> 
> 
> I ran the e2e tests against the PR environment. Although they don't provide full coverage yet, nothing failed. Somehow this "flexibility" and our misunderstanding seems like it's due to a deficiency in the ASP.NET Core API

This isn't so much a deficiency in the ASP.NET Core API as it was a complete misunderstanding on our part of how a small part of ASP.NET works.

PR #1444, opened 2021-07-14
PRIME-1454 Error messaging  UX

johnnyl6:


neophyte57:
My test is to try to go to the Banner pages without the required Keycloak permissions.  The UI needs some tuning I think: 

image

prime-angular-frontend/src/app/core/services/error-handler.service.ts
```diff
@@ -29,16 +32,7 @@ export class ErrorHandlerService implements ErrorHandler {
...
       // Client error has occurred (Angular Error, ReferenceError...)
+      dialogLogger.log(error);
     }
-
-    const message = (error.message)
-      ? error.message
-      : error.toString();
-    const url = router.url;
-
-    logger.error(message, { url });
```
neophyte57, line 39: Should the error still go to `LoggerService`?  I'm thinking about PRIME-1325 which would record it to web-api.

prime-angular-frontend/src/app/shared/components/dialogs/content/error-logger/error-logger.component.html
```diff
@@ -0,0 +1,15 @@
+<mat-dialog-content>
+  <p>URL: {{ this.router.url }}</p>
+  <p>Environment: {{ this.config.environmentName }}</p>
+  <p>Version: {{ this.enviornment.version }}</p>
+  <mat-expansion-panel *ngIf="this.error.stack">
+    <mat-expansion-panel-header>
+      <mat-panel-title>
+        Details
+      </mat-panel-title>
```
neophyte57, line 9: I think we need some better messaging to the end user, as per "Need UX messaging and next steps" in the Jira ticket.  Get during demo?

prime-angular-frontend/src/app/shared/components/dialogs/content/error-logger/error-logger.component.ts
```diff
@@ -0,0 +1,28 @@
...
+
+@Component({
+  selector: 'app-error-logger',
+  templateUrl: './error-logger.component.html',
+  styleUrls: ['./error-logger.component.scss']
+})
+export class ErrorLoggerComponent implements OnInit {
+  public error: any;
+  public enviornment = environment;
+
```
neophyte57, line 17: `enviornment` typo

prime-angular-frontend/src/app/shared/components/dialogs/content/error-logger/error-logger.component.ts
```diff
@@ -0,0 +1,28 @@
...
+  constructor(
+    @Inject(MAT_DIALOG_DATA) public data: DialogOptions,
+    @Inject(APP_CONFIG) public config: AppConfig,
+    public router: Router
+  ) {
+    this.error = data.data;
```
neophyte57, line 23: Although I see the same naming in `ClaimNoteComponent` e.g., can the variable `data` be called something else, possibly `options` so there isn't code like `data.data`?

prime-angular-frontend/src/app/core/services/error-handler.service.ts
johnnyl6: > Should the error still go to `LoggerService`? I'm thinking about PRIME-1325 which would record it to web-api.

Yes, and I believe PRIME-1325 would also log the client error in the else statement

prime-angular-frontend/src/app/core/services/error-handler.service.ts
```diff
@@ -29,16 +32,7 @@ export class ErrorHandlerService implements ErrorHandler {
...
-
-    // IMPORTANT: Rethrow the error, otherwise it gets swallowed
-    throw error;
```
hogmann, line 42: I'm pretty sure we need to keep this in there right?

johnnyl6: yep deleted by accident.

prime-angular-frontend/src/app/shared/components/dialogs/content/error-logger/error-logger.component.html
```diff
@@ -0,0 +1,15 @@
+<mat-dialog-content>
+  <p>URL: {{ this.router.url }}</p>
+  <p>Environment: {{ this.config.environmentName }}</p>
+  <p>Version: {{ this.environment.version }}</p>
+  <mat-expansion-panel *ngIf="this.error.stack">
```
mtpultz, line 5: Don't use `this` in templates.  The `public` variables are already scoped to the template.

prime-angular-frontend/src/app/core/services/error-handler.service.ts
```diff
@@ -29,6 +32,7 @@ export class ErrorHandlerService implements ErrorHandler {
       }
     } else {
       // Client error has occurred (Angular Error, ReferenceError...)
+      dialogLogger.log(error);
```
mtpultz, line 35: Running this locally and creating an error it appears to invoke the dialog twice

prime-angular-frontend/src/app/shared/components/dialogs/content/error-logger/error-logger.component.html
mtpultz: Also, the user will be confused by the stack trace.  I would probably only display outside of production.

prime-angular-frontend/src/app/shared/classes/dialog-logger.ts
```diff
@@ -0,0 +1,36 @@
...
+
+@Injectable({
+  providedIn: 'root'
+})
+export class DialogLogger {
+  constructor(
+    private dialog: MatDialog
+  ) { }
+
+  public log(error: Error): void {
```
mtpultz, line 16: The messaging on this dialog is a bit hard for users to understand.  We need a generic message indicating an "unknown error" has occurred at a "date and time", and provide a unique error ID they can pass off to a "specific contact".  I provided Pilar with the quoted strings to construct the message with Sarah.  

Can you update the API logger to respond with the ID of the row it just added and we'll use that as the unique error ID to help locate errors easier when reported, and then it can be passed as a parameter to `log`.

prime-angular-frontend/src/app/shared/components/dialogs/content/error-logger/error-logger.component.html
```diff
@@ -0,0 +1,16 @@
+<mat-dialog-content>
+  <p>URL: {{ this.router.url }}</p>
```
mtpultz, line 2: Looks like you still have a couple more `this`

prime-angular-frontend/src/app/core/resources/logger-resource.service.ts
```diff
@@ -15,8 +16,11 @@ export class LoggerResource {
     private http: HttpClient
   ) { }
 
-  public createLog(log: Log): Observable<HttpResponse<any>> {
+  public createLog(log: Log): Observable<number> {
     return this.http
-      .post(`${this.config.apiEndpoint}/client-logs`, log, { observe: 'response' });
+      .post(`${this.config.apiEndpoint}/client-logs`, log, { observe: 'response' })
```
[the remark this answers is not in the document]
johnnyl6: > Is there a reason not to dependency inject the ApiResource? Seems like we're just duplicating what is provide through the service and using the HttpClient directly.
I have no idea. This is code from another PR but can definitely change it

prime-angular-frontend/src/app/shared/components/dialogs/content/error-logger/error-logger.component.ts
```diff
@@ -0,0 +1,30 @@
...
+import { DialogOptions } from '../../dialog-options.model';
+
+@Component({
+  selector: 'app-error-logger',
+  templateUrl: './error-logger.component.html',
+  styleUrls: ['./error-logger.component.scss']
+})
+export class ErrorLoggerComponent implements OnInit {
+  public error: any;
+  public environment = environment;
```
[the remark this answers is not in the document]
johnnyl6: To simplify it, I've removed the details panel so no environment information needed for now.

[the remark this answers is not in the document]
mtpultz: That works out well since with the changes to `config-map` we now get full stack traces in console in dev and test so we don't need the panel.

prime-angular-frontend/src/app/shared/components/dialogs/content/error-logger/error-logger.component.html
```diff
@@ -0,0 +1,14 @@
+<mat-dialog-content>
+  <p>An unknown error has occurred at {{ currentDate | date:'medium' }}</p>
```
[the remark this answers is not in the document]
mtpultz: @johnnyl6 I've updated the dialog content and messaging a bit, and removed the rethrow on the error handler.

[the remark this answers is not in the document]
mtpultz: Also, I remove the use of the native Date object and used Moment to generate the date.  We want to avoid using the native date object in JS.

mtpultz (approved):
LGTM

AnaisHebert (approved):
🦆🦆

PR #1451, opened 2021-07-15
OPS: Trigger bespoke cleanup action on PR close event

i5okie:
- Move cleanup step into its own action
- Trigger only on PR close event.


i5okie:
testing pr.

mtpultz (approved):

itchison (approved):

PR #1452, opened 2021-07-16
PRIME-1729 Move environment config out of docker image

mtpultz:
In PR won't work until `config-map.json` is injected by the pipeline with the appropriate environment information.

Locally everything works as before using the cascade of environment defaults, and `config-map.json` is .gitignored to prevent it from being accidentally committed if used to test locally.

Using the `APP_INITIALIZER` injection token and provider hierarchy causes a lot of issues around circular dependencies, or refactoring away from using AppConfig and instead using a service.  This solution pre-fetches the config map prior to bootstrapping the application in `main.ts` and moves the injection token creation out of `AppConfigModule` side stepping issues and refactoring.


hogmann (approved):
LGTM

johnnyl6 (approved):

PR #1453, opened 2021-07-16
Feature/prime 1731 make filters persistant

hogmann:


prime-angular-frontend/src/app/modules/adjudication/shared/components/search-form/search-form.component.ts
```diff
@@ -22,15 +23,22 @@ export class SearchFormComponent implements OnInit {
...
     private configService: ConfigService,
+    private localStorage: LocalStorageService
   ) {
     this.statuses = this.configService.statuses;
     this.search = new EventEmitter<string>();
     this.filter = new EventEmitter<EnrolmentStatusEnum>();
     this.refresh = new EventEmitter<void>();
+
+    this.textSearchKey = 'search-form-textSearch';
+    this.statusCodeKey = 'search-form-statusCode';
```
johnnyl6, line 41: Since these two keys reside in a shared component, suspect it would cause problem when being used in multiple pages, and values could be overwritten/loaded accidentally.

AnaisHebert (changes requested):
I set my status filter to 'Locked', and had 1 enrollee in my list who was 'Editable'. When I reloaded the page, the Locked filter was secleted, but it did not filter out my editable enrollee. If i pressed the refresh table button, it still did not filter. Seems like the filter only works on change maybe?

AnaisHebert (approved):
🦝🦝🦝

Paahn (approved):
hashtagsuccess

PR #1454, opened 2021-07-17
Health Auth and Paper Enrolment Updates

mtpultz:
Brings HealthAuthorityModule and PaperEnrolmentModule inline with the pattern for overview components being encapsulated within the associated page component, form state, and form models; to assist with decoupling overview pages.  

Also, aligning response model with client models, and only convert when necessary within the form state/form model.  For example, using givenNames instead of middleName except within the form state/form model for consistency throughout the application.


AnaisHebert (approved):
🦝🦝🦝

Paahn (approved):
successhighfive

PR #1473, opened 2021-07-28
Update README

Paahn:
- Amendment to README to inform new developers verifiable credentials are not needed to run the project


documentation/VerifiableCredentials.md
```diff
@@ -1,5 +1,7 @@
 ## Running the Aries Agent Locally for Verifiable Credential Testing and Development
 
+**Please Note** The following process is only necessary when working with Verifiable Credential Testing and Development, ie not required when initially setting up the project.
```
mtpultz, line 3: Looks great!!!  

I would probably use `**NOTE:**`, and replace `ie` with `, and is not necessary for general development purposes.  For example, when initially setting up the project.`

mtpultz (approved):
Woot! Someone in the future will thank you!

documentation/VerifiableCredentials.md
neophyte57: Can we eliminate the sentence fragment, i.e. `... general development purposes, for example, when ...`

AnaisHebert (approved):
🦆🦆🦆

documentation/VerifiableCredentials.md
mtpultz: If you want to eliminate the sentence fragment it would be `purposes; for example, when initially`

PR #1477, opened 2021-07-28
Remove direct use of environment for config

mtpultz:
With the recent update to allow swapping of environment variables through OCP4 config maps we should not be access environment files directly, and should only use config injection token for access to these values.  

Will create another update that essentially removes import access of the environment file, but this interim update will fix the issues that will occur outside of our local environments.

### DIFF
--- a/prime-angular-frontend/src/app/core/services/console-logger.service.ts
+++ b/prime-angular-frontend/src/app/core/services/console-logger.service.ts
@@ -1,13 +1,15 @@
-import { Injectable } from '@angular/core';
+import { Inject, Injectable } from '@angular/core';
 
-import { environment } from '@env/environment';
+import { APP_CONFIG, AppConfig } from 'app/app-config.module';
 import { AbstractLoggerService } from '@core/services/abstract-logger.service';
 
 @Injectable({
   providedIn: 'root'
 })
 export class ConsoleLoggerService extends AbstractLoggerService {
-  constructor() {
+  constructor(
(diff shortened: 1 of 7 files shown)


prime-angular-frontend/src/app/core/services/console-logger.service.ts
```diff
@@ -24,7 +26,7 @@ export class ConsoleLoggerService extends AbstractLoggerService {
    * Prints the logging information, but ONLY if not in production.
    */
   protected send(type: string, params: { msg?: string, data?: any[] }) {
-    if (!environment.production || type === 'error' || type === 'warn') {
+    if (this.config.environmentName !== 'prod' || type === 'error' || type === 'warn') {
```
mtpultz, line 29: Bonus: using config allows us to know the actual environment versus whether the web client was build using `--prod` so we will now have stack traces in dev and test.

prime-angular-frontend/src/main.ts
```diff
@@ -17,6 +17,7 @@ fetch('/assets/config-map.json')
     let appConfig = defaultAppConfig;
 
     if (configMap) {
+      // TODO mohKeyCloakConfig will eventually have configuration applied throughout the environments
```
mtpultz, line 20: `mohKeyCloak` config will be added to OCP4 when the GIS team deploys out their test environment at the end of next week or early the following week.

AnaisHebert (approved):
🍕🍕🍕

Paahn (approved):
✌️

PR #1493, opened 2021-08-04
PRIME-1697 PLR Information

johnnyl6:
To view PLR information on enrollees with a college licence to help with adjudication decisions in Admin

In order to test, PLR data need to be manually inserted into database table PlrProvider with matching collegeId of the enrollee.

Sample data (2 records) SQL script with collegeID `12345`, the enrollee must have a matching college licence number `123456`

```
INSERT INTO public."PlrProvider"
("Id", "CreatedUserId", "CreatedTimeStamp", "UpdatedUserId", "UpdatedTimeStamp", "Ipc", "IdentifierType", "CollegeId", "ProviderRoleType", "MspId", "NamePrefix", "FirstName", "SecondName", "ThirdName", "LastName", "Suffix", "Gender", "DateOfBirth", "StatusCode", "StatusReasonCode", "StatusStartDate", "StatusExpiryDate", "Expertise", "Address1Line1", "Address1Line2", "Address1Line3", "City1", "Province1", "Country1", "PostalCode1", "Address1StartDate", "Credentials", "TelephoneAreaCode", "TelephoneNumber", "FaxAreaCode", "FaxNumber", "Email", "ConditionCode", "ConditionEndDate", "ConditionStartDate", "Cpn")
VALUES(1, '00000000-0000-0000-0000-000000000000'::uuid, '2021-08-05 15:09:56.199', '00000000-0000-0000-0000-000000000000'::uuid, '2021-08-05 15:09:56.199', 'IPC.00003638.BC.PRS', 'CPSID', '12345', 'MD', '121466115', 'Dr.', 'Martin', 'Owen', NULL, 'Davis', NULL, 'M', '1994-09-28 00:00:00.000', 'SUSPENDED', 'RET', '2014-06-01 00:00:00.000', NULL, '["165","S31","302","240"]', '15406 Kilkee Pl', NULL, NULL, 'Surrey', 'BC', 'CA', 'V3S 6E8', '2011-01-31 00:00:00.000', '["MD"]', '866', '2818165', NULL, '866123456', NULL, NULL, NULL, NULL, 'CPN.00003638.BC.PRS');

```


prime-dotnet-webapi/Models/Plr/PlrExpertise.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace Prime.Models.Plr
+{
+    [Table("PlrExpertiseLookup")]
+    public class PlrExpertise : ILookup<string>
+    {
+        [Key]
+        public string Code { get; set; }
```
AnaisHebert, line 11: We usually have a autoincrementing integer id for our lookup tables. Not sure if it is necessary in this case, but maybe would cause some issues doing it this way if we want to link to this table in the future?

johnnyl6: Followed same pattern of other *Lookup tables like LicenseLookup, Country, Province etc. they do not have incrementing Id either.

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html
```diff
@@ -53,25 +53,29 @@ <h2>Adjudication</h2>
...
-                               (assignToa)="onAssignToa($event)"
-                               (enableEnrollee)="onEnableEnrollee($event)"
-                               (enableEditing)="onEnableEditing($event)"
-                               (cancelToa)="onCancelToa($event)"
-                               (rerunRules)="onRerunRules($event)"
-                               (toggleManualAdj)="onToggleManualAdj($event)"
-                               (reload)="onRefresh()">
-      </app-adjudicator-actions>
-
+      <div class="row justify-content-end">
```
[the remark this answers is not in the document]
johnnyl6: Good question. The TOA selection dropdown and actions menu are currently inside one component, which would affect any other pages using it. Will confirm with UX designer.

prime-dotnet-webapi/Configuration/Seeded Tables/PlrExpertiseConfiguration.cs
```diff
@@ -0,0 +1,351 @@
...
+                    new PlrExpertise { Code = "580",      Name = "Pediatric Radiology" },
+                    new PlrExpertise { Code = "582",      Name = "Pediatric Emergency Medicine" },
+                    new PlrExpertise { Code = "584",      Name = "Developmental Pediatrics" },
+                    new PlrExpertise { Code = "474",      Name = "Rheumatology(Int.Med or Ped.)" },
+                    new PlrExpertise { Code = "900",      Name = "Pediatric Gastroenterology" },
+                    new PlrExpertise { Code = "905",      Name = "Pediatric Nephrology" },
+                    new PlrExpertise { Code = "910",      Name = "Pediatric Neurology" },
+                    new PlrExpertise { Code = "915",      Name = "Pediatric Cardiovascular Surgery" },
+                    new PlrExpertise { Code = "A110",     Name = "dministrative Medicine" },
+                    new PlrExpertise { Code = "A115",     Name = "Adolescent Medicine" },
```
neophyte57, line 160: There is a typo with A110

prime-dotnet-webapi/ViewModels/AutoMapping.cs
```diff
@@ -114,5 +119,29 @@ public AutoMapping()
         CreateMap<PlrProvider, PlrProvider>()
             .ForMember(dest => dest.Id, opt => opt.Ignore())
             .ForMember(dest => dest.Ipc, opt => opt.Ignore());
+
+        // ProjectTo does not work for mapping expertise array, use workaround below and call mapper.Map
+        CreateMap<PlrProvider, PlrViewModel>()
```
hogmann, line 124: For this instead of having the complicated mapping, we can go into the PlrProvider model and explicitly  set foreign key references to the new lookups you just made.

[ForeignKey("PlrRoleType")]
public string ProviderRoleType { get; set; }

[JsonIgnore]
public PlrRoleType PlrRoleType { get; set; }

Not sure if you need the JsonIgnore, but this will simplify the mapping not to need DB calls in the mapper. And helps describe the relationship in the models. Which is better than hiding this relationship in the mapper.

The only thing we have to watch for is if there is data outside that defined in the lookups. Though the mapper would fail in that case too, just the way above would fail on data insertion to the DB

prime-dotnet-webapi/Startup.cs
```diff
@@ -100,6 +100,8 @@ public void ConfigureServices(IServiceCollection services)
                 .AddNewtonsoftJson(options =>
                 {
                     options.SerializerSettings.Converters.Add(new EmptyStringToNullJsonConverter());
+                    // required to fix the error "Self referencing loop detected" caused by PlrViewModel Expertise property
+                    options.SerializerSettings.ReferenceLoopHandling = ReferenceLoopHandling.Ignore;
```
hogmann, line 104: Not 100% sure what this is doing, but the other comment might solve this error. But if not we should look into this more since this looks like something we should not be ignoring.

hogmann (changes requested):

prime-dotnet-webapi/ViewModels/AutoMapping.cs
hogmann: Usually we use our ORM to generate these FK relationships but like everything we can explicitly describe them since we don't want to change the names of the columns from an external source.

prime-dotnet-webapi/Startup.cs
johnnyl6: Correct, this line is no longer needed. Deleted.

prime-dotnet-webapi/ViewModels/AutoMapping.cs
johnnyl6: > Usually we use our ORM to generate these FK relationships but like everything we can explicitly describe them since we don't want to change the names of the columns from an external source.

This is the reason I am hesitated to use foreignkey since these PLR and lookup data are from external source and could change over time. Do we really want to maintain and keep them in sync?

hogmann: Did we generate these lookup table values or are they defined by PLR?

hogmann: Looking at the ticket it seems these were given to us from PLR, So I think it's safe to say if their data breaks these definitions then they are on the hook for that. I'd have to look at how we do insertion, but realistically if should just not insert rows with no foreign key constraint. In my mind this isn't a case we should try and make a work around in our code for, unless we have been explicitly told that PLR change these type def.

hogmann: One thing to note is 
``` .FirstOrDefault().Name ``` is going to throw an error when code is not in the lookup on 126/127 so you will have to handle the case.

You can do this by defining a default or use ?.

Comment (hogmann) [reply prime-dotnet-webapi/ViewModels/AutoMapping.cs:125]: Also was just talking to Alan and he is going to reach out to PLR as it may be that they would notify us when they are adding new values so that having the foreign key is a reasonable solution.

Comment (johnnyl6) [reply prime-dotnet-webapi/ViewModels/AutoMapping.cs:125]: > One thing to note is
> `.FirstOrDefault().Name` is going to throw an error when code is not in the lookup on 126/127 so you will have to handle the case.
> 
> You can do this by defining a default or use ?.

Tested the null case, and surprisingly no exception is thrown. might be due to it's a IQueryable?

Comment (johnnyl6) [reply prime-dotnet-webapi/ViewModels/AutoMapping.cs:123]: > Also was just talking to Alan and he is going to reach out to PLR as it may be that they would notify us when they are adding new values so that having the foreign key is a reasonable solution.

@hogmann Foreignkey is added to PlrProvider model for RoleType and StatusReason. However Expertise is an array property deserialized from JSON string in db so could not be mapped using foreignkey.

Comment (johnnyl6) [reply prime-dotnet-webapi/ViewModels/AutoMapping.cs:125]: > Also was just talking to Alan and he is going to reach out to PLR as it may be that they would notify us when they are adding new values so that having the foreign key is a reasonable solution.

Hi @hogmann , per feedback from @itchison  we'd like to always accept PLR messages so FK might cause rejection if the lookup table is not updated in a timely manner or data out of sync. I've reverted the code. Can you take another look?

**Review by hogmann — APPROVED** [review]
LGTM

**Review by neophyte57 — CHANGES_REQUESTED** [review]

File: prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html [hunk 6]
```diff
@@ -53,25 +53,29 @@ <h2>Adjudication</h2>
...
+                                 (cancelToa)="onCancelToa($event)"
+                                 (rerunRules)="onRerunRules($event)"
+                                 (toggleManualAdj)="onToggleManualAdj($event)"
+                                 (reload)="onRefresh()">
+        </app-adjudicator-actions>
+      </div>
+      <div *ngIf="plrInfo?.length">
+        <ng-container *ngFor="let plr of plrInfo">
+          <app-plr-info [plrData]="plr"></app-plr-info>
+        </ng-container>
```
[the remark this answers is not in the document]
johnnyl6: > Does this also need to be added to Adjudication screen as well?
> 
> image

Good point. Had a discussion with Pilar about it, we will just have the PLR info on overview page for MVP now as the mockup shows. Might need the adjudication based on future feedback.

neophyte57 (approved):
LGTM!

PR #1495, opened 2021-08-05
Fix - Hide green alert when within the 90 days of renewal

Paahn:
Fix - Feedback from demo
- Hide the green alert "You have approval to access PharmaNet"  when we are within the 90 day period of renewal, we only need to display the blue alert notifying the user of the renewal.


hogmann (approved):
LGTM

johnnyl6 (approved):
🎉

PR #1519, opened 2021-08-12
Fix to nginx.conf for PLR connections 

neophyte57:
Similar to #1498 but cleaner PR 

My email to CGI PLR team:  "FYI, the problem was that our Nginx configuration was missing `ssl_client_certificate` which specifies the Certificate Authorities the server trusts (CN=TestRootCA in this case).  During the SSL handshake, PLR then looks up in its distribution.jks keystore the client certificate that has that CA in its certificate chain, and responds with the matching client certificate (CN=TestClient) to complete the mutual authentication."


infrastructure/prime-app-ephemeral-template.yml
```diff
@@ -1037,7 +1037,7 @@ objects:
     namespace: ${OC_NAMESPACE}-${OC_APP}
     labels:
       app.kubernetes.io/part-of: ${SVC_NAME}
-    annotations: 
+    annotations:
```
AMiKaH, line 1040: Good catch on all the trailing spaces!

AMiKaH (approved):

i5okie (approved):
lgtm

PR #1532, opened 2021-08-17
PRIME-1755 GC Collect Endpoint

johnnyl6:


AMiKaH (approved):

prime-dotnet-webapi/Controllers/UtilController.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Prime.Controllers
+{
+    [Produces("application/json")]
+    [Route("api/[controller]")]
+    [ApiController]
+    public class UtilController : PrimeControllerBase
+    {
+        [HttpGet("gccollect", Name = nameof(GCCollect))]
+        [Authorize(Roles = Roles.ManageEnrollee)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
```
neophyte57, line 17: Is this the correct Role to require?

itchison: I asked for manager admin.

hogmann: Since we don't have a way to access this endpoint through the app is there a point in adding perm? Like i guess if you export your token and use it through postman you can get this to work?

mtpultz (approved):

PR #1536, opened 2021-08-18
Fix multiple enrollee call on admin view that wipes filter

hogmann:
This bug only presents it self with large enrollee counts when the getEnrollees call takes time to return. On the enrollee page there was two calls being fired since we were watching the query params and the route params. So even though param.id was undefined we were then firing off a get all enrollees call. So the duplication in two spots caused the slower of the calls to return after the filtered call and then repopulate the data source with all enrollees.


neophyte57 (approved):
Seems reasonable

AMiKaH (approved):

PR #1547, opened 2021-08-23
PRIME-1774 Pharmacy Site Reg - email notification to Provider Enrolment team

johnnyl6:


prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -753,6 +753,32 @@ public async Task<ActionResult> SendRemoteUsersEmailUser(int siteId, IEnumerable
...
         }
 
+        // POST: api/Sites/5/site-reviewed-email-user
+        /// <summary>
+        /// Send site reviewed notification email to provider enrolment team
+        /// </summary>
+        /// <param name="siteId"></param>
+        /// <param name="note"></param>
+        /// <returns></returns>
+        [HttpPost("{siteId}/site-reviewed-email-user", Name = nameof(SendSiteReviewedNotificationEmail))]
```
james-hollinger, line 763: I assume this URL was copied from the above email endpoint? A more clear URL would just be `/site-reviewed-email`

prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -753,6 +753,32 @@ public async Task<ActionResult> SendRemoteUsersEmailUser(int siteId, IEnumerable
...
+        [Authorize(Roles = Roles.ViewSite)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        public async Task<ActionResult> SendSiteReviewedNotificationEmail(int siteId, FromBodyText note)
+        {
+            var site = await _siteService.GetSiteAsync(siteId);
```
james-hollinger, line 771: I would like to stop fetching the entire `Site` just to check that it exists when we can. Looking ahead, the only information on the `Site` we need for this email is the PEC, so we are throwing away a lot of retrieved data here.
I would prefer to see a `SiteService.SiteExists(int siteId)` method, and then only fetching the PEC in `SendSiteReviewedNotificationAsync(int siteId, note)`

prime-dotnet-webapi/Migrations/ApiDbContextModelSnapshot.cs
```diff
@@ -17939,111 +17949,6 @@ the Province.
...
-                            CareSettingCode = 1,
-                            Email = "",
-                            Name = "Meditech"
-                        },
-                        new
-                        {
-                            Code = 35,
-                            CareSettingCode = 1,
-                            Email = "",
-                            Name = "Cerner"
```
james-hollinger, line 18046: All of these removed lines should be in the snapshot. I'm not sure how you local got into this state, but could you please regenerate your migration? After removing the migration you may have to do `git checkout develop -- C:\Path\to\the\repo\prime-dotnet-webapi\Migrations\ApiDbContextModelSnapshot.cs` to fetch the snapshot that's currently on develop.

prime-dotnet-webapi/ViewModels/Emails/SiteNotificationEmailViewModel.cs
```diff
@@ -0,0 +1,8 @@
+namespace Prime.ViewModels.Emails
+{
+    public class SiteNotificationEmailViewModel
```
james-hollinger, line 3: I would rather this ViewModel be specific to the one email using it rather than having a more generic name. How about `SiteReviewedEmailViewModel`.

PR #1563, opened 2021-08-30
PRIME-1793 Choosing 24 hours in hours of operation for all days causes error

mtpultz:
Choosing 24 hours disables the associated fields which is expected, but the status of valid and disabled are not mutually exclusive so when a form is disabled validations are not run and it is considered invalid when every form field is disabled.

On this particular page disabled fields are known to be valid so when all fields are disabled then we don't need validation so checkValidity was overwritten locally to account for validations not being required if everything is disabled.


prime-angular-frontend/src/app/shared/components/forms/fields/doing-business-as-form-field/doing-business-as-form-field.component.html
```diff
@@ -1,13 +1,13 @@
...
+         [formControl]="doingBusinessAs"
+         [matAutocomplete]="auto">
+  <mat-autocomplete #auto="matAutocomplete">
+    <mat-option *ngFor="let doingBusinessAsName of doingBusinessAsNames"
+                [value]="doingBusinessAsName">
+      {{ doingBusinessAsName }}
+    </mat-option>
+  </mat-autocomplete>
+  <mat-error>Required</mat-error>
+</mat-form-field>
```
Paahn, line 13: there's a failing test on this file, checking commit history made me run the tests to see this. Previously the html was commented out to avoid the test failure when we were getting the tests up and running

mtpultz: Okay, will have a look.  Commenting this out broke health authority.  Couldn't enter a site name and therefore not submit the page.

johnnyl6 (approved):
🎊

prime-angular-frontend/src/app/shared/components/forms/fields/doing-business-as-form-field/doing-business-as-form-field.component.html
johnnyl6: the test is missing code to initialize the property ` @Input() public doingBusinessAs: FormControl;`
so the following would make it happy
```typescript
  beforeEach(() => {
    fixture = TestBed.createComponent(DoingBusinessAsFormFieldComponent);
    component = fixture.componentInstance;
    component.doingBusinessAs = new FormControl('parent control');
    fixture.detectChanges();
  });
```

mtpultz: Thanks Johnny!

Paahn (approved):
🥳 🤳

PR #1573, opened 2021-08-31
PRIME 1762 Site Registration Duplicate PEC

johnnyl6:


prime-angular-frontend/src/app/lib/validators/unique-pec.validator.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from '@angular/core';
+import { AbstractControl, AsyncValidator, ValidationErrors } from '@angular/forms';
+
+import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
+
+import { SiteResource } from '@core/resources/site-resource.service';
+
+@Injectable({ providedIn: 'root' })
```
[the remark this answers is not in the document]
johnnyl6: Neat implementation. Thanks!🙏

AMiKaH (approved):

prime-dotnet-webapi/Services/interfaces/ISiteService.cs
```diff
@@ -51,5 +51,6 @@ public interface ISiteService
         Task<SiteRegistrationNoteViewModel> GetSiteRegistrationNoteAsync(int siteId, int siteRegistrationNoteId);
         Task<IEnumerable<int>> GetNotifiedSiteIdsForAdminAsync(ClaimsPrincipal user);
         Task<bool> SiteExists(int siteId);
+        Task<bool> PecExistsInNonHealthAuthoritySites(string pec);
```
james-hollinger, line 54: I would advise against putting business logic in the name of a method. I makes the name really long and it means that if we want to change the business logic later we have to alter the name. I would prefer something like `PecIsAvailibleAsync` or `ValidatePecAsync` or `CheckPecConflictAsync` or something like that.

prime-dotnet-webapi/Services/SiteService.cs
```diff
@@ -282,6 +287,11 @@ public async Task<Site> UpdatePecCode(int siteId, string pecCode)
             var site = await GetBaseSiteQuery()
                 .SingleOrDefaultAsync(s => s.Id == siteId);
 
+            if (await IsNonHaSiteAndPecNotUnique(site, pecCode))
+            {
+                throw new InvalidOperationException($"Could not update the site.");
```
james-hollinger, line 292: We should not be throwing an error in a service method like this due to business logic; especially if the controller is not validating this information. If you try to update the PEC by calling PUT /sites/5/pec and the PEC happens to exist the server will respond with a `500 Internal Service Error` due to this uncaught exception. The controller should be validating this at a minimum.

mtpultz (approved):
image

PR #1597, opened 2021-09-09
Automapper Profiles

james-hollinger:
Organize the AutoMapper profiles before undertaking fixing the EnrolleeViewModel


mtpultz (approved):
MTGL

prime-dotnet-webapi/ViewModels/Agreements/_AgreementMappingProfile.cs
```diff
@@ -0,0 +1,23 @@
+using AutoMapper;
+
+using Prime.Models;
+
+namespace Prime.ViewModels.Profiles
+{
+    public class AgreementMappingProfile : Profile
```
johnnyl6, line 7: any benefit using underscore prefix for the class filename like `_AgreementMappingProfile.cs`?

hogmann (approved):
Looks Good To Marty

prime-dotnet-webapi/ViewModels/Agreements/_AgreementMappingProfile.cs
james-hollinger: Putting the underscore makes the profile bubble up to the top of the file list, making it easier to find. 
image

vs

image

PR #1600, opened 2021-09-10
PRIME-1813-re-order-banners

AMiKaH:


prime-angular-frontend/src/app/modules/adjudication/pages/site-banner-list-page/site-banner-list-page.component.ts
```diff
@@ -37,6 +37,6 @@ export class SiteBannerListPageComponent implements OnInit {
 
   private getBanners(): void {
     this.bannerResource.getSiteLandingBanners()
-      .subscribe((banners: Banner[]) => this.banners = banners);
+      .subscribe((banners: Banner[]) => this.banners = banners.reverse());
```
mtpultz, line 40: Though we can do this in the web client.  It would be better to have the order determined in the API.  There are two pages that get banners based on location in adjudication - sites and enrolments.  So the order can be determined within the querying of `GetBannersAsync`

AMiKaH: Sure, I will reverse that in the API then.

AMiKaH: Moved

prime-dotnet-webapi/Services/BannerService.cs
```diff
@@ -67,6 +67,8 @@ public async Task<IEnumerable<BannerViewModel>> GetBannersAsync(BannerLocationCo
                 banner.EndTimestamp = DateTime.SpecifyKind(banner.EndTimestamp, DateTimeKind.Utc);
             });
 
+            banners.Reverse();
```
mtpultz, line 70: I would look at EF Core's query API and use `OrderByDescending(...)`.  You should be able to do all the work in the DB query.

AMiKaH: It seems that we don't have an attribute for when the banner was created in order to sort using the `OrderByDescending()`
The closest I can find to sort them based on when they were posted would be the `Banner.Id`. We have StartTime  EndTime attributes but those are irrelevant to the posting time of the banner.

mtpultz: That's what I would have done since don't have a CreatedDate or UpdatedDate to adjust the resulting set of banners.

mtpultz (approved):
LGTM

prime-dotnet-webapi/Services/BannerService.cs
mtpultz: And they haven't asked to have banners change order based on changes/updates

johnnyl6 (approved):

james-hollinger (approved):
👍

PR #1604, opened 2021-09-11
Refactor Dependency Injection

james-hollinger:
Currently we are injecting `IHttpContextAccessor` into all of our services despite not using it anywhere. It has now been removed from `BaseService` and replaced by `ILogger` so we can have a logger in every service by default.

I have alphabetized the dependency injection parameters as well as the locals. Since the import ordering depends on the suffix (we could DI both an `EnrolleeService` and `ErolleeClient` for example), the ordering includes the suffix. For consistency, I have reordered the DI registration lines in `Startup.cs` to follow.

I have also alphabetized the `using`s in the services, abandoning my previous arbitrary scheme of "loosely sorted by length to make a nice triangle". Maybe one day I'll get to every file.

Tests have been updated to generate service mocks using reflection rather than explicit generator methods. This should reduce maintenance burden.


mtpultz (approved):
MTGL

AMiKaH (approved):
Looks good, any way we can bring back the 0.5% drop in coverage?

james-hollinger:
> 
> 
> Looks good, any way we can bring back the 0.5% drop in coverage?

Presumably that drop is due to us having a `CreateService()` method that instantiated concrete services and so contributed to code coverage a tiny bit. These have been replaced by the reflection based `MockDependenciesFor` method

PR #1605, opened 2021-09-13
PRIME-1796 Create a new role for "maintenance"

AMiKaH:
Added role prime_maintenance in keycloak, the role can view and edit banners, the role can edit email templates

### DIFF
--- a/prime-angular-frontend/src/app/modules/adjudication/shared/components/notification-email-view/notification-email-view.component.html
+++ b/prime-angular-frontend/src/app/modules/adjudication/shared/components/notification-email-view/notification-email-view.component.html
@@ -18,7 +18,7 @@
     <mat-icon>edit</mat-icon>
   </button>
 
-  <button *ngIf="(Role.SUPER_ADMIN | inRole) && editable"
+  <button *ngIf="((Role.SUPER_ADMIN | inRole) || (Role.MAINTENANCE | inRole)) && editable"
           mat-icon-button
           matTooltip="Cancel Editing"
           (click)="toggleEdit(false)">
--- a/prime-angular-frontend/src/app/modules/adjudication/shared/components/notification-email-view/notification-email-view.component.html
+++ b/prime-angular-frontend/src/app/modules/adjudication/shared/components/notification-email-view/notification-email-view.component.html
@@ -11,7 +11,7 @@
     {{(editable) ? "Editing" : "Preview"}}
   </ng-container>
 
-  <button *ngIf="(Role.SUPER_ADMIN | inRole) && !editable"
+  <button *ngIf="((Role.SUPER_ADMIN | inRole) || (Role.MAINTENANCE | inRole)) && !editable"
           mat-icon-button
           matTooltip="Edit Email Template"
           (click)="toggleEdit(true)">
--- a/prime-angular-frontend/src/app/modules/adjudication/shared/components/notification-email-view/notification-email-view.component.html
+++ b/prime-angular-frontend/src/app/modules/adjudication/shared/components/notification-email-view/notification-email-view.component.html
@@ -11,18 +11,11 @@
     {{(editable) ? "Editing" : "Preview"}}
   </ng-container>
 
-  <button *ngIf="((Role.SUPER_ADMIN | inRole) || (Role.MAINTENANCE | inRole)) && !editable"
+  <button *ngIf="Role.MAINTENANCE | inRole"
           mat-icon-button
-          matTooltip="Edit Email Template"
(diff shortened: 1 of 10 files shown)


prime-angular-frontend/src/app/modules/adjudication/shared/components/notification-email-view/notification-email-view.component.html
```diff
@@ -11,14 +11,14 @@
...
-  <button *ngIf="(Role.SUPER_ADMIN | inRole) && !editable"
+  <button *ngIf="((Role.SUPER_ADMIN | inRole) || (Role.MAINTENANCE | inRole)) && !editable"
           mat-icon-button
           matTooltip="Edit Email Template"
           (click)="toggleEdit(true)">
     <mat-icon>edit</mat-icon>
   </button>
 
-  <button *ngIf="(Role.SUPER_ADMIN | inRole) && editable"
+  <button *ngIf="((Role.SUPER_ADMIN | inRole) || (Role.MAINTENANCE | inRole)) && editable"
```
AMiKaH, line 21: Do we need the check for role here?
This button is inside a dialogue that shows after clicking `Edit` which would not show if the role check does not evaluate to true to display the `Edit` button in the first place (line 14)

mtpultz: For consistency I would make it explicit.  Arguably this should be a single button that changes its configuration based on edit-ability instead of two buttons that create redundant markup in which case it would be applied to "both" as a single element.

prime-angular-frontend/src/app/modules/adjudication/shared/components/notification-email-view/notification-email-view.component.html
```diff
@@ -11,14 +11,14 @@
     {{(editable) ? "Editing" : "Preview"}}
   </ng-container>
 
-  <button *ngIf="(Role.SUPER_ADMIN | inRole) && !editable"
+  <button *ngIf="((Role.SUPER_ADMIN | inRole) || (Role.MAINTENANCE | inRole)) && !editable"
```
mtpultz, line 14: I would combine these buttons into single button.  Also, the `InRolePipe` can take singular or multiple roles: 

```html
  <button *ngIf="[Role.SUPER_ADMIN, Role.MAINTENANCE] | inRole"
          mat-icon-button
          [matTooltip]="(!editable) ? 'Edit Email Template' : 'Cancel Editing'"
          (click)="toggleEdit(!editable)">
    <mat-icon>{{ (!editable) ? 'edit' : 'close' }}</mat-icon>
  </button>
```

prime-dotnet-webapi/Controllers/BannersController.cs
```diff
@@ -49,7 +49,7 @@ public async Task<ActionResult> CreateBanner(BannerViewModel viewModel, [FromQue
         /// <param name="bannerId"></param>
         /// <param name="viewModel"></param>
         [HttpPut("{bannerId}", Name = nameof(UpdateBanner))]
-        [Authorize(Roles = Roles.PrimeSuperAdmin)]
+        [Authorize(Roles = Roles.PrimeMaintenance)]
```
james-hollinger, line 52: Create/delete banner probably could also be maintenance instead of super admin

mtpultz (approved):
LGTM

prime-dotnet-webapi/Controllers/BannersController.cs
AMiKaH: I'll have to double check tomorrow just to be sure. I asked a question about create/delete during scrum and if my memory is working the answer was that maintenance role would be more for editing text.

AMiKaH: Confirmed, will be adding create/delete for banners

james-hollinger (approved):
LGTM

prime-angular-frontend/src/app/modules/adjudication/shared/components/notification-email-view/notification-email-view.component.html
AMiKaH: Changed to toggle buttons and check for one role instead of 2.

By the way, this check `*ngIf="[Role.SUPER_ADMIN, Role.MAINTENANCE] | inRole` would not work if the user only has one of the roles, it evaluates to true if user has both `SUPER_ADMIN` and `MAINTENANCE` but evaluates to false if use has one or the other

Paahn (approved):

PR #1618, opened 2021-09-14
diagnostics

james-hollinger:


prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -56,11 +57,25 @@ public class EnrolleesController : PrimeControllerBase
...
+                System.Console.WriteLine($"---- first db call: {sw.ElapsedMilliseconds}ms");
+
+                sw = Stopwatch.StartNew();
                 var enrollees = await _enrolleeService.GetEnrolleesAsync(searchOptions);
+                sw.Stop();
+                System.Console.WriteLine($"---- second db call: {sw.ElapsedMilliseconds}ms");
+
+                sw = Stopwatch.StartNew();
                 var result = enrollees.Select(e => e.SetNotification(notifiedIds.Contains(e.Id)));
+                sw.Stop();
```
AMiKaH, line 76: If the `notifiedIds` gets too large would the `Contains(e.Id)` also cause a delay which is not be related to DB performance? I noticed that the last `WriteLine` is not writing about DB call round time measurement but just wanted to confirm that's not what we're looking for in the last logged message.

prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -56,11 +57,25 @@ public class EnrolleesController : PrimeControllerBase
         [ProducesResponseType(typeof(ApiResultResponse<IEnumerable<EnrolleeListViewModel>>), StatusCodes.Status200OK)]
         public async Task<ActionResult> GetEnrollees([FromQuery] EnrolleeSearchOptions searchOptions)
         {
+            Console.WriteLine("---- entered controller");
```
AMiKaH, line 60: Adding the controller name can help here.

AMiKaH (commented):
Is the logging supposed to stay uncommented all the time? Or is it just to get to PR environment? If the latter then ignore my other comments :D

james-hollinger:
This was a PR I put up to do some timing on DB calls in openshift due to database connection issues we were seeing yesterday

PR #1637, opened 2021-09-23
PRIME-1821 Cron Job to Re-run Rules on Manual Review Enrollees

neophyte57:


prime-dotnet-webapi/Services/Rules/AutomaticAdjudicationRules.cs
```diff
@@ -86,6 +92,7 @@ public override async Task<bool> ProcessRule(Enrollee enrollee)
 
             bool passed = true;
 
+            _logger.LogDebug($"Going to use {_collegeLicenceClient.GetType().Name} in PharmanetValidationRule");
```
james-hollinger, line 95: Testing code?

prime-dotnet-webapi/Services/SubmissionService.cs
```diff
@@ -304,6 +305,29 @@ private async Task RerunRulesAsync(Enrollee enrollee)
             await _enrolleeService.RemoveNotificationsAsync(enrollee.Id);
         }
 
+        public async Task RerunRulesAsync()
+        {
+            var enrollees = GetBaseQueryForEnrolleeApplicationRules()
+                .Where(e => e.Adjudicator == null)
+                .Where(e => e.CurrentStatus.StatusCode == (int)StatusType.UnderReview)
+                // TODO: Don't hard-code StatusReasonCodes
+                .Where(e => e.CurrentStatus.EnrolmentStatusReasons.Any(esr => esr.StatusReasonCode > 2 && esr.StatusReasonCode < 8))
```
james-hollinger, line 314: I would prefer a list of applicable status reason codes with a `.Contains()` rather than doing a range operation here. Its confusing what the intent of the code is.

james-hollinger: Also, you don't need the magic numbers, we have `StatusReasonType`

prime-dotnet-webapi/Controllers/SubmissionsController.cs
```diff
@@ -215,6 +215,23 @@ public async Task<ActionResult> RerunRules(int enrolleeId)
...
+        /// <summary>
+        /// Re-runs the automatic adjudication rules for all applicable Enrollees.
+        /// </summary>
+        [HttpPost("rerun-rules", Name = nameof(RerunAutomaticAdjudicationRules))]
+        [Authorize(Roles = Roles.PrimeApiServiceAccount)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status200OK)]
```
james-hollinger, line 228: Many of these Swagger annotations are incorrect

prime-dotnet-webapi/Services/Rules/AutomaticAdjudicationRules.cs
neophyte57: Although this isn't exactly telemetry, from https://www.se-radio.net/2020/05/episode-409-joe-kutner-on-the-twelve-factor-app/  "And so in the book beyond the 12-factor app, Kevin makes this analogy to a space probe. Like your application is like this craft that you’re sending into space and it’s not coming back to earth. So if there’s a problem with it, when it’s in orbit, you can only rely on the instrumentation that you installed before it was launched. And so we need to think about our apps that run in the cloud the same way. If there’s a problem in production, you need to have the metrics, the telemetry, the tools that you need to debug that right there with it designed, you know, included with it upfront." ... thoughts?

james-hollinger: Neat article, and I agree logging and telemetry is important. But this isn't logging anything of use; this is logging the name of which college license client is used. It will log "CollegeLicenceClient" every time. That is not useful data.

neophyte57: It can also log `DummyCollegeLicenceClient` depending on the situation (currently `OC_APP` value)

prime-dotnet-webapi/Services/SubmissionService.cs
neophyte57: Apologies, you may have explained before, but is there a reason the enum is named `StatusReasonType` rather than `StatusReasonCode`?

james-hollinger: It was the convention near the start of the app's development. I think calling them codes is a bit nicer in most cases but there are a few times it can be confusing. Also, `StatusCodes` is unfortunately already an enum (for http status codes like 200Ok)

prime-dotnet-webapi/Controllers/SubmissionsController.cs
neophyte57: Should be fixed now

prime-dotnet-webapi/Services/Rules/AutomaticAdjudicationRules.cs
neophyte57: Due to inability to change signature of `Startup.ConfigureServices`, I couldn't log just a single time.  But I moved the log statement to `SubmissionRulesService`

AMiKaH (approved):

prime-dotnet-webapi-tests/Unit Tests/AutomaticAdjudicationRules.Tests.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
```
johnnyl6, line 4: seems to be unused

neophyte57: Fixed  :-)

PR #1642, opened 2021-09-24
[hotfix] use correct image tag

i5okie:
- let imagechange trigger in deployment config automatically update correct image tag for init containers
fixes initcontainers not running latest migrations


mtpultz (approved):

hogmann (approved):
LGTM

PR #1643, opened 2021-09-25
Fix obosites on renewal

mtpultz:
Allow the addition of obosites for new health authority care settings on renewal

NOTE: Applied to test already


james-hollinger (approved):
L AlreadyAppliedtoTest TM

PR #1644, opened 2021-09-25
Refactor existing cron jobs to use jq and ConfigMap 

neophyte57:
This change is to mainly reflect what is actually deployed in `dev`.  The changes that weren't synced to the repo used `jq` (small binary) rather than `python` (large dependency) as well as referencing the services accurately as the naming convention changed.

One new change is to use the ConfigMap `keycloak` rather than the Secret `keycloak`, the latter being deprecated.


james-hollinger (approved):
LGTM

johnnyl6 (approved):

PR #1647, opened 2021-09-27
PRIME-1873 No indication of site status to user on site registration card

johnnyl6:
- set SubmittedDate to null to indicate current status in frontend

**NOTE:** Already applied to TEST


james-hollinger (approved):
L SubmittedDate=null TM

PR #1658, opened 2021-09-29
Feature/prime 199 report absence

hogmann:


hogmann:
Just in the process of adding a dialog for creating an absence, so as to let the user know a past one is being made

prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -926,5 +926,142 @@ public async Task<ActionResult> GetPlrData(int enrolleeId)
...
             return Ok(result);
         }
+
+        // POST: api/Enrollees/5/enrollee-absence
+        /// <summary>
+        /// Creates a new enrollee absence.
+        /// </summary>
+        /// <param name="enrolleeId"></param>
+        /// <param name="createModel"></param>
+        [HttpPost("{enrolleeId}/enrollee-absence", Name = nameof(CreateEnrolleeAbsence))]
```
james-hollinger, line 936: We can simplify this URL to just `{enrolleeId}/absences`

prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -926,5 +926,142 @@ public async Task<ActionResult> GetPlrData(int enrolleeId)
...
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        public async Task<ActionResult> CreateEnrolleeAbsence(int enrolleeId, EnrolleeAbsenceViewModel createModel)
+        {
+            var record = await _enrolleeService.GetPermissionsRecordAsync(enrolleeId);
+            if (record == null)
+            {
+                return NotFound($"Enrollee not found with id {enrolleeId}");
+            }
+            if (!record.MatchesUserIdOf(User))
```
james-hollinger, line 948: if the intention is to exclude adjudicators from calling this endpoint, we should add `[Authorize(Roles  = Roles.PrimeEnrollee)] to this method.

prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -926,5 +926,142 @@ public async Task<ActionResult> GetPlrData(int enrolleeId)
...
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(typeof(ApiResultResponse<EnrolleeAbsenceViewModel>), StatusCodes.Status200OK)]
+        public async Task<ActionResult> GetAbsence(int enrolleeId)
+        {
+            var record = await _enrolleeService.GetPermissionsRecordAsync(enrolleeId);
+            if (record == null)
+            {
+                return NotFound($"Enrollee not found with id {enrolleeId}");
+            }
+            if (!record.MatchesUserIdOf(User))
```
james-hollinger, line 975: As above

prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -954,5 +954,75 @@ public async Task<IEnumerable<string>> GetEnrolleeEmails(BulkEmailType bulkEmail
                 .DecompileAsync()
                 .ToListAsync();
         }
+
+        public async Task<EnrolleeAbsence> CreateEnrolleeAbsenceAsync(int enrolleeId, DateTime startTimestamp, DateTime endTimestamp)
+        {
+            if (startTimestamp >= endTimestamp)
+            {
+                throw new ArgumentException($"endTimestamp is not after the startTimestamp of absence.");
+            }
```
james-hollinger, line 963: This validation can be on the view model, and we can automatically return a 400 rather expecting to throw here. Search the repo for `AbstractValidator` for some examples (we've only added it to newer VMs).

prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -954,5 +954,75 @@ public async Task<IEnumerable<string>> GetEnrolleeEmails(BulkEmailType bulkEmail
...
+
+            var enrollee = await _context.Enrollees
+                .Include(e => e.EnrolleeAbsences)
+                .SingleOrDefaultAsync(e => e.Id == enrolleeId);
+
+            var absence = new EnrolleeAbsence
+            {
+                StartTimestamp = startTimestamp,
```
james-hollinger, line 971: should be calling `.ToUniversalTime()` on these to make sure they're UTC

prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -954,5 +954,75 @@ public async Task<IEnumerable<string>> GetEnrolleeEmails(BulkEmailType bulkEmail
...
         }
+
+        public async Task<EnrolleeAbsence> CreateEnrolleeAbsenceAsync(int enrolleeId, DateTime startTimestamp, DateTime endTimestamp)
+        {
+            if (startTimestamp >= endTimestamp)
+            {
+                throw new ArgumentException($"endTimestamp is not after the startTimestamp of absence.");
+            }
+
+            var enrollee = await _context.Enrollees
```
james-hollinger, line 965: You don't need to pull up the `Enrollee`, you can just add the new absences to `_context.EnrolleeAbsences`

prime-dotnet-webapi/Models/EnrolleeAbsence.cs
```diff
@@ -0,0 +1,23 @@
...
+  [Table("EnrolleeAbsence")]
+  public class EnrolleeAbsence : BaseAuditable, IEnrolleeNavigationProperty
+  {
+    [Key]
+    public int Id { get; set; }
+
+    [JsonIgnore]
+    public int EnrolleeId { get; set; }
+
+    [JsonIgnore]
```
james-hollinger, line 17: Since this is a new, rather isolated model that uses View Models correctly; I feel its worth starting to avoid using `[JsonIgnore]` in our Models. Its a hold-over from the before times when we were exposing our domain models directly. The only caveat: if this is supposed to be part of the enrolleee snapshot we might have to use these headers anyway :(

prime-dotnet-webapi/ViewModels/Enrollee/EnrolleeAbsenceViewModel.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+namespace Prime.ViewModels
+{
+    public class EnrolleeAbsenceViewModel
+    {
+        public int Id { get; set; }
+
+        public int EnrolleeId { get; set; }
```
james-hollinger, line 9: Do we need Id or EnrolleeId on the VM? Is the frontend actually using them?

prime-dotnet-webapi/Services/EnrolleeService.cs
hogmann: whoops, artifact of the first implementation with not using the context.

prime-dotnet-webapi/ViewModels/Enrollee/EnrolleeAbsenceViewModel.cs
hogmann: We need Id, but can infer enrollee Id.

prime-dotnet-webapi/Services/EnrolleeService.cs
hogmann: Looked at one and put one in that should do what i want, but only way to check is using a postman endpoint with token so can test shortly.

PR #1733, opened 2021-10-29
PRIME-1939 Create mock PharmanetTransactionLog data to test performance of queries

neophyte57:


utilities/mock-pharmanettransactionlog-data-creator/loop-upload.ps1
```diff
@@ -0,0 +1,6 @@
+for ($num = 1 ; $num -le 5 ; $num++)
+{   
+	$command = "oc cp .\pnet-logs${num}.csv pr-1721-webapi-7-dzssg:/tmp/pnet-logs${num}.csv"
```
mtpultz, line 3: Should PR-1721 be replaced with something more descriptive?

mtpultz (approved):

utilities/mock-pharmanettransactionlog-data-creator/loop-upload.ps1
neophyte57: These are just helper scripts e.g. in case the person wanting to create mock data needs to upload the files to a pod to load into the database, which isn't necessary; it can be loaded from local with `oc port-forward`.  In the README, I indicate "These require some tweaking for the given situation"

johnnyl6 (approved):

AMiKaH (approved):

PR #1741, opened 2021-11-05
PRIME-1946 Remote Users Link to Enrollees are Dropped in Subsequent Site Submissions

johnnyl6:
For community site's subsequent submission, changed SiteService so that:
- add remote users if not existing in current site
- update remote users for existing one for site
- delete remote users if not existing in update model


prime-dotnet-webapi/Services/CommunitySiteService.cs
```diff
@@ -196,40 +196,41 @@ private void UpdateBusinessHours(Site current, CommunitySiteUpdateModel updated)
...
             }
 
-            if (updated?.RemoteUsers != null && updated?.RemoteUsers.Count() != 0)
+            foreach (var updateRemoteUser in updated.RemoteUsers)
             {
-                foreach (var remoteUser in updated.RemoteUsers)
-                {
-                    remoteUser.SiteId = current.Id;
-                    var remoteUserCertifications = new List<RemoteUserCertification>();
+                var existingRemoteUser = _context.RemoteUsers
```
mtpultz, line 206: Feels like we could do this up front and not call out to the database on each iteration something like:

```c#
var existingRemoteUsers = await _context.RemoteUsers
    .Where(ru => ru.SiteId == current.Id)
    .Include(ru => ru.RemoteUserCertifications)
    .ToListAsync();

foreach (var updatedRemoteUser in updated.RemoteUsers)
{
    var existingRemoteUser = existingRemoteUsers.FirstOrDefault(ru => ru.Id == updatedRemoteUser.Id);
```

james-hollinger, line 206: the current Site already has the existing remote users, you don't need to re-fetch them

prime-dotnet-webapi/Services/CommunitySiteService.cs
```diff
@@ -196,40 +196,41 @@ private void UpdateBusinessHours(Site current, CommunitySiteUpdateModel updated)
 
         private void UpdateRemoteUsers(Site current, CommunitySiteUpdateModel updated)
```
james-hollinger, line 197: its a bit cleaner to pass in the updateModel.RemoteUsers rather than the whole update model here

prime-dotnet-webapi/Services/CommunitySiteService.cs
```diff
@@ -196,40 +196,41 @@ private void UpdateBusinessHours(Site current, CommunitySiteUpdateModel updated)
...
-                {
-                    remoteUser.SiteId = current.Id;
-                    var remoteUserCertifications = new List<RemoteUserCertification>();
+                var existingRemoteUser = _context.RemoteUsers
+                    .FirstOrDefault(u => u.Id == updateRemoteUser.Id);
 
-                    foreach (var certification in remoteUser.RemoteUserCertifications)
+                if (existingRemoteUser == null)
+                {
+                    current.RemoteUsers.Add(updateRemoteUser);
```
james-hollinger, line 211: Unfortunately, we are still using models for the site update endpoint here rather than view models. Because of this, you have to be careful when taking them and using them directly like this because the front-end could have manipulated the ID (or SiteId). In this case, you should set `updatedRemoteUser.Id = 0` so EF knows its a new entity and doesn't either try to update some unrelated record or complain you are trying to add a record that already exists.

prime-dotnet-webapi/Services/CommunitySiteService.cs
```diff
@@ -196,40 +196,41 @@ private void UpdateBusinessHours(Site current, CommunitySiteUpdateModel updated)
...
 
-            if (updated?.RemoteUsers != null && updated?.RemoteUsers.Count() != 0)
+            foreach (var updateRemoteUser in updated.RemoteUsers)
             {
-                foreach (var remoteUser in updated.RemoteUsers)
-                {
-                    remoteUser.SiteId = current.Id;
-                    var remoteUserCertifications = new List<RemoteUserCertification>();
+                var existingRemoteUser = _context.RemoteUsers
+                    .FirstOrDefault(u => u.Id == updateRemoteUser.Id);
```
james-hollinger, line 207: pragmatically should be `SingleOrDefault` since we expect Ids to be unique. Also, should be `SingleOrDefaultAstnc` if youa re hitting the DB.

prime-dotnet-webapi/Services/CommunitySiteService.cs
```diff
@@ -196,40 +196,41 @@ private void UpdateBusinessHours(Site current, CommunitySiteUpdateModel updated)
...
+                            _context.Entry(existingCertification).CurrentValues.SetValues(certification);
+                        }
                     }
-                    remoteUser.RemoteUserCertifications = remoteUserCertifications;
-
-                    _context.Entry(remoteUser).State = EntityState.Added;
                 }
             }
+
+            _context.RemoteUsers.RemoveRange(current.RemoteUsers.Where(u => !updated.RemoteUsers.Contains(u)));
```
james-hollinger, line 233: Calling `.Contains` on a collection of objects isn't going to work the way you hope, unfortunately. 
image

The default equality comparer for objects is reference equality, i.e, are they the same object. This will be false for every member since they are not the same object (one set of objects came from the DB and the other came from the FE).

As an aside, you _could_ implement the `.Equals()` on `RemoteUser` to have a custom comparer, but there are several problems with that. For one, you would also have to implement `.GetHashCode()` and this is kinda polluting the data model at that point. The other issue is that it isn't really the correct way to do it anyway. I would not expect `.Equals` on two objects to only test the Id to say they are the same object. These are not the same object:

```C#
var person1 = new Person
{
    Id = 1,
    Name = "Alice"
};

var person2 = new Person
{
    Id = 1,
    Name = "Bob"
};
```

so I would not expect them to return true if I did `person1.Equals(person2)`.  In reality they may represent the same *database entry*, witch is not the same concept and so is not accurately handled by `.Equals()` or `.Contains()`.

prime-dotnet-webapi/Services/CommunitySiteService.cs
```diff
@@ -196,40 +196,41 @@ private void UpdateBusinessHours(Site current, CommunitySiteUpdateModel updated)
...
+                    .FirstOrDefault(u => u.Id == updateRemoteUser.Id);
 
-                    foreach (var certification in remoteUser.RemoteUserCertifications)
+                if (existingRemoteUser == null)
+                {
+                    current.RemoteUsers.Add(updateRemoteUser);
+                }
+                else
+                {
+                    _context.Entry(existingRemoteUser).CurrentValues.SetValues(updateRemoteUser);
```
james-hollinger, line 215: This line will overwrite the SiteId on the existing `RemoteUser` with whatever SiteId comes from the UpdateModel, which is bad since it could have been set by the FE to any value. We need to set the SiteId on the UpdateModel to the site's Id to prevent it from being manipulated.

prime-dotnet-webapi/Services/CommunitySiteService.cs
johnnyl6: Had to use the _context to check if a remote user exists for current site here. In the foreach loop new remote user is being added to current.RemoteUsers and the collection is changing so cannot use it for query.

prime-dotnet-webapi/Services/CommunitySiteService.cs
johnnyl6: Line 206 checks existing remote user using `Id` field already, so if it does not exist in DB, `updatedRemoteUser.Id` should be 0. If the Id has been manipulated by frontend, say setting to an existing `Id=123`, backend should do its duty to update the existing remote user but not treat it as a new remote user.

prime-dotnet-webapi/Services/CommunitySiteService.cs
johnnyl6: In the foreach loop, new remote users are added by `current.RemoteUsers.Add()`, and existing remote user are updated thru `_context.Entry().SetValues()`. Somehow using `Contains` here does work magically (tested) likely due to EF uses reference in SetValues() maybe?

prime-dotnet-webapi/Services/CommunitySiteService.cs
```diff
@@ -194,42 +194,45 @@ private void UpdateBusinessHours(Site current, CommunitySiteUpdateModel updated)
...
+                }
+                else
+                {
+                    updateRemoteUser.SiteId = current.Id;
+                    _context.Entry(existingRemoteUser).CurrentValues.SetValues(updateRemoteUser);
+                    foreach (var certification in existingRemoteUser.RemoteUserCertifications)
                     {
-                        var newCertification = new RemoteUserCertification
+                        var existingCertification = existingRemoteUser.RemoteUserCertifications
+                            .FirstOrDefault(c => c.Id == certification.Id);
```
james-hollinger, line 221: This line does nothing, it takes the certification and then finds it again?

johnnyl6: Good catch. Should be ` var existingCertification = _context.RemoteUserCertifications...`

prime-dotnet-webapi/Services/CommunitySiteService.cs
johnnyl6: Done some tests, and it seems checking equality by reference is valid here.

prime-dotnet-webapi/Services/CommunitySiteService.cs
james-hollinger: We don't have to update the `RemoteUserCertification`s, we can just drop and add them

prime-dotnet-webapi/Services/CommunitySiteService.cs
james-hollinger: Due to this line of code, existing `RemoteUser`s are being dropped and re-added rather than updated.

james-hollinger: Please see the failing unit test I added

james-hollinger (changes requested):
Code is not working as intended

image
image

image
image

prime-dotnet-webapi/Services/CommunitySiteService.cs
johnnyl6: Thanks for the unit tests! Pushed more fix.

prime-dotnet-webapi/Services/CommunitySiteService.cs
johnnyl6: > Code is not working as intended
> 
> image image
> 
> image image

Saw the unit test. Updated code.

mtpultz (approved):
image

hogmann (approved):
LGTM tested it up and works as expected!

AMiKaH (approved):

PR #1743, opened 2021-11-05
PRIME-1851 - Device provider site reg

hogmann:
To the site registration workflow adding the ability to select Device Provider as a careType. This is behind a feature flag.

Device Provider sites act just like Comm Pharm sites with one addition. They have at least one individual device provider. This page falls right after the hours of operation page.

Currently you cannot view device provider sites in admin as there is no tab for it.


prime-angular-frontend/src/app/modules/site-registration/pages/device-provider-page/device-provider-page-form-state.class.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  public patchValue(providers: IndividualDeviceProvider[]): void {
+    if (!this.formInstance || !Array.isArray(providers)) {
+      return;
+    }
+
+    this.individualDeviceProviders.clear(); // Clear out existing indices
+
+    providers.forEach((p: IndividualDeviceProvider) => this.push(p));
```
mtpultz, line 36: The method name `push` is slightly deceptive since you're creating a group, patching, and then pushing.  Also, it looks like you only use this once so you could drop the method and inline its content within this `patchValue`

prime-angular-frontend/src/app/modules/site-registration/pages/device-provider-page/device-provider-page-form-state.class.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  public at(index: number): FormGroup {
```
mtpultz, line 39: Since it's `public` it might be worth providing a bit of context to the method name like `individualDeviceProvidersAt(index: number)` or `individualDeviceProviderAt(index: number)`

prime-angular-frontend/src/app/modules/site-registration/pages/care-setting-page/care-setting-page.component.ts
```diff
@@ -149,6 +152,14 @@ export class CareSettingPageComponent extends AbstractCommunitySiteRegistrationP
     this.patchForm();
   }
 
+  protected submissionRequest(): NoContent {
+    if (this.formState.json.careSettingCode !== CareSettingEnum.DEVICE_PROVIDER) {
+      this.siteFormStateService.deviceProviderFormState.patchValue([]);
```
mtpultz, line 157: Does this clear the device providers?  Feels like it would need to be `this.siteFormStateService.deviceProviderFormState.patchValue({ individualDeviceProviders: [] });`, or access the getter and invoke clear -`this.siteFormStateService.deviceProviderFormState.individualDeviceProviders.clear()`

prime-angular-frontend/src/app/modules/site-registration/pages/device-provider-page/device-provider-page.component.ts
```diff
@@ -0,0 +1,176 @@
...
+
+@Component({
+  selector: 'app-device-provider-page',
+  templateUrl: './device-provider-page.component.html',
+  styleUrls: ['./device-provider-page.component.scss'],
+  providers: [FormatDatePipe]
+})
+export class DeviceProviderPageComponent extends AbstractCommunitySiteRegistrationPage implements OnInit {
+  public formState: DeviceProviderPageFormState;
+  public form: FormGroup;
```
mtpultz, line 33: `form` is already included in the base class

prime-angular-frontend/src/app/modules/site-registration/pages/device-provider-page/device-provider-page.component.ts
```diff
@@ -0,0 +1,176 @@
...
+      ],
+      action: {
+        title: 'Update Information'
+      }
+    });
+  }
+
+  protected additionalValidityChecks(obj: { individualDeviceProviders: IndividualDeviceProvider[] }): boolean {
+    // At least one provider needs to exist
+    return (obj.individualDeviceProviders.length > 0);
```
mtpultz, line 162: You won't need the parenthesis since length will either be 0 (falsey) or greater than 0 (truthy)

prime-angular-frontend/src/app/modules/site-registration/pages/device-provider-page/device-provider-page.component.ts
```diff
@@ -0,0 +1,176 @@
...
+  public onEdit(index: number): void {
+    this.currentIndex = index;
+    this.form = this.formState.at(index);
+    this.isEditing = true;
+  }
+
+  public onContinue(): void {
+    if (this.formUtilsService.checkValidity(this.form)) {
+      if (this.currentIndex !== -1) {
+        this.formState.individualDeviceProviders.at(this.currentIndex).patchValue(this.form);
```
mtpultz, line 97: Can access the indexed FormGroup using `this.formState.at(index).patchValue(...)` or after renaming `this.formState.individualDeviceProviderAt(index).patchValue(...)`

prime-angular-frontend/src/app/modules/site-registration/pages/hours-operation-page/hours-operation-page.component.ts
```diff
@@ -161,11 +161,23 @@ export class HoursOperationPageComponent extends AbstractCommunitySiteRegistrati
...
   protected afterSubmitIsSuccessful(): void {
     const site = this.siteService.site;
-    let routePath = SiteRoutes.REMOTE_USERS;
-
-    if (site.careSettingCode === CareSettingEnum.COMMUNITY_PHARMACIST || site.careSettingCode === CareSettingEnum.DEVICE_PROVIDER) {
-      routePath = SiteRoutes.ADMINISTRATOR;
-    } else if (this.isCompleted) {
+    let routePath;
+
+    switch (site.careSettingCode) {
```
mtpultz, line 166: Feels like this could be in an `else` since `this.isCompleted` will always override when true.

prime-angular-frontend/src/app/core/resources/site-resource.service.ts
```diff
@@ -55,9 +56,14 @@ export class SiteResource {
 
   public getSiteById(siteId: number, statusCode?: number): Observable<Site> {
     const params = this.apiResourceUtilsService.makeHttpParams({ statusCode });
-    return this.apiResource.get<Site>(`sites/${siteId}`, params)
-      .pipe(
-        map((response: ApiHttpResponse<Site>) => response.result),
+    return forkJoin({
+      site: this.apiResource.get<Site>(`sites/${siteId}`, params)
+        .pipe(map((response: ApiHttpResponse<Site>) => response.result)),
```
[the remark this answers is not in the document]
hogmann: im assuming the next line i can do the same thing?
``` this.apiResource.get<IndividualDeviceProvider[]>(`sites/${siteId}/individual-device-providers`, null, null, true) ```

File: prime-angular-frontend/src/app/modules/site-registration/pages/device-provider-page/device-provider-page-form-state.class.ts (hunk 2 above)
Comment (hogmann) [reply prime-angular-frontend/src/app/modules/site-registration/pages/device-provider-page/device-provider-page-form-state.class.ts:39]: Only use it once in the component so just removed it and call it directly to make it clearer.

File: prime-angular-frontend/src/app/modules/site-registration/pages/care-setting-page/care-setting-page.component.ts (hunk 3 above)
Comment (hogmann) [reply prime-angular-frontend/src/app/modules/site-registration/pages/care-setting-page/care-setting-page.component.ts:157]: It works because the patchValue is the method defined in the form state.
image

File: prime-angular-frontend/src/app/core/resources/site-resource.service.ts (hunk 8 above)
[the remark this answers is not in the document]
Comment (mtpultz) [reply prime-angular-frontend/src/app/core/resources/site-resource.service.ts:61]: Yeah definitely, and in PIP it would be `this.apiResource.get<IndividualDeviceProvider[]>('sites/${siteId}/individual-device-providers')` so we won't need the extra params in the near future

File: prime-angular-frontend/src/app/modules/site-registration/pages/care-setting-page/care-setting-page.component.ts (hunk 3 above)
Comment (mtpultz) [reply prime-angular-frontend/src/app/modules/site-registration/pages/care-setting-page/care-setting-page.component.ts:157]: The intent might be clearer if you used `this.siteFormStateService.deviceProviderFormState.clearIndividualDeviceProviders()` then you can reuse it in the `patchValue` method

File: prime-angular-frontend/src/app/modules/site-registration/pages/device-provider-page/device-provider-page.component.ts (hunk 5 above)
Comment (mtpultz) [reply prime-angular-frontend/src/app/modules/site-registration/pages/device-provider-page/device-provider-page.component.ts]: You'll find `Pick<...>` effective for communicating the params and knowing where they originate and in your case you don't have a `IndividualDeviceProviderForm` model so you could use `Site`

```
protected additionalValidityChecks({ individualDeviceProviders }: Pick<Site, 'individualDeviceProviders'>): boolean {
  return individualDeviceProviders?.length;
}
```

**Review by mtpultz — APPROVED** [review]
image

**Review by Paahn — APPROVED** [review]
🦔

File: prime-dotnet-webapi/Controllers/SitesController.cs [hunk 9]
```diff
@@ -1198,5 +1198,27 @@ public async Task<ActionResult> FlagSite(int siteId, FromBodyData<bool> flagged)
...
+        /// <param name="siteId"></param>
+        [HttpGet("{siteId}/individual-device-providers", Name = nameof(GetIndividualDeviceProviders))]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(typeof(ApiResultResponse<IEnumerable<IndividualDeviceProviderViewModel>>), StatusCodes.Status200OK)]
+        public async Task<ActionResult> GetIndividualDeviceProviders(int siteId)
+        {
+            var site = await _communitySiteService.GetSiteAsync(siteId);
+            if (site == null)
```
james-hollinger, line 1215: this can be `.SiteExists()`

prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -1198,5 +1198,27 @@ public async Task<ActionResult> FlagSite(int siteId, FromBodyData<bool> flagged)
...
+        /// <summary>
+        /// Gets the Individual Device Providers for a Device Provider Site.
+        /// </summary>
+        /// <param name="siteId"></param>
+        [HttpGet("{siteId}/individual-device-providers", Name = nameof(GetIndividualDeviceProviders))]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(typeof(ApiResultResponse<IEnumerable<IndividualDeviceProviderViewModel>>), StatusCodes.Status200OK)]
+        public async Task<ActionResult> GetIndividualDeviceProviders(int siteId)
```
james-hollinger, line 1212: If this is an endpoint that can be hit by the user, we need to check the permissions record. If its an admin only endpoint, this needs `[Authorize(Roles = Roles.ViewSite)]`

prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -1198,5 +1198,27 @@ public async Task<ActionResult> FlagSite(int siteId, FromBodyData<bool> flagged)
...
+        {
+            var site = await _communitySiteService.GetSiteAsync(siteId);
+            if (site == null)
+            {
+                return NotFound($"Site not found with id {siteId}");
+            }
+
+            var providers = await _communitySiteService.GetIndividualDeviceProvidersAsync(siteId);
+            return Ok(providers);
```
james-hollinger, line 1221: don't need this in two lines, it can be `return Ok(await _communitySiteService.GetIndividualDeviceProvidersAsync(siteId));`

prime-dotnet-webapi/Models/SiteRegistration/IndividualDeviceProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Prime.Models
+{
+    [Table("IndividualDeviceProvider")]
+    public class IndividualDeviceProvider : BaseAuditable
+    {
+        [Key]
+        public int Id { get; set; }
+
+        [Required]
+        public int CommunitySiteId { get; set; }
```
james-hollinger, line 15: Dont need `[Required]` on the FK property, it should go on the `CommunitySite` property. I'm pretty sure you actually don't need this at all, since this is an `int` property rather than a `int?` so EF knows that `CommunitySite` is required.

prime-dotnet-webapi/Models/SiteRegistration/IndividualDeviceProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        [JsonIgnore]
```
james-hollinger, line 17: `[JsonIgnore]` on our domain models was a band-aid because we didn't use View Models properly. Lets avoid adding it to new models where possible.

prime-dotnet-webapi/Services/CommunitySiteService.cs
```diff
@@ -99,6 +99,7 @@ public async Task UpdateSiteAsync(int siteId, CommunitySiteUpdateModel updatedSi
             UpdateContacts(currentSite, updatedSite);
             UpdateBusinessHours(currentSite, updatedSite);
             UpdateRemoteUsers(currentSite, updatedSite.RemoteUsers);
+            await UpdateIndividualDeviceProviders(siteId, updatedSite);
```
james-hollinger, line 102: Might as well pass in `updatedSite.IndividualDeviceProviders` rather than the whole update Model.

prime-dotnet-webapi/Services/CommunitySiteService.cs
```diff
@@ -263,6 +264,27 @@ private void UpdateVendors(CommunitySite current, CommunitySiteUpdateModel updat
...
+            var currentProviders = await _context.IndividualDeviceProviders
+                .Where(p => p.CommunitySiteId == siteId)
+                .ToListAsync();
+            _context.IndividualDeviceProviders.RemoveRange(currentProviders);
+
+
+            if (updated.IndividualDeviceProviders == null)
+            {
+                return;
+            }
```
james-hollinger, line 278: The current trend of our updates is if the update's collection is `null`, we don't touch the collection. We empty out the collection if we sent up an empty collection. Does that behavior fit with this feature? If so, this guard clause should move up to line 269.

prime-dotnet-webapi/ViewModels/SiteRegistration/IndividualDeviceProviderCreateModel.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace Prime.ViewModels
+{
+    public class IndividualDeviceProviderChangeModel
+    {
+        public int? CommunitySiteId { get; set; }
```
james-hollinger, line 7: Is this property necessary?

hogmann: was before not now.

james-hollinger (approved):
L "LGMT" TM

PR #1746, opened 2021-11-05
Fix Check Warning

james-hollinger:
Since the fix to this is stuck in the PR for Device Provider enrollment until we can merge it, I've pulled this out to stop the warnings.


AMiKaH (approved):

hogmann (approved):
LGTM

PR #1749, opened 2021-11-08
PRIME-1960 Health Authority Site PEC Validation

AMiKaH:


prime-dotnet-webapi/Controllers/HealthAuthoritySitesController.cs
```diff
@@ -251,5 +251,30 @@ public async Task<ActionResult> HealthAuthoritySiteSubmission(int healthAuthorit
...
+            {
+                return NotFound();
+            }
+            if (await _healthAuthoritySiteService.CheckForDuplicatePecAsync(healthAuthorityId, pec))
+            {
+                return Ok();
+            }
+
+            return NotFound();
+        }
```
james-hollinger, line 278: Why make a new endpoint + service method instead of updating the existing code to determine if a PEC is assignable?

AMiKaH: That could be done. I just tried keep the HA end point in the HA Sites controller. Also, I just looked at the `PecAssignable()` end point and the `PecAssignableAsync()` method. I'm not sure those are easier to edit and use than to just add the new endpoint ands service method. We can chat about that when you have a minute.

prime-angular-frontend/src/app/modules/health-auth-site-reg/pages/site-information-page/site-information-page.component.ts
```diff
@@ -83,4 +85,17 @@ export class SiteInformationPageComponent extends AbstractHealthAuthoritySiteReg
 
     this.routeUtils.routeRelativeTo(nextRoutePath);
   }
+
+  private addAsyncValidator() {
```
hogmann, line 89: Might be best to add this validation to the form state as it seems to be constant. That being said you can then get around and just create a AsyncValidatorFn and attach it when you create the form group. As well as adding a blur so its not firing all the time on change.

This is an example of what you can do
``` 
public buildForm(): void {
    this.formInstance = this.fb.group({
      siteName: ['', [Validators.required]],
      pec: [null, {
        asyncValidators: [this.pecValidator()],
        updateOn: 'blur'
      }],
      securityGroupCode: [null, [Validators.required]]
    });
  }

private pecValidator(): AsyncValidatorFn {
    return (control: AbstractControl): Observable<ValidationErrors | null> => {
      return this.healthAuthorityResource.checkPecExistsInOtherHEalthAuthority(
        this.healthAuthoritySiteService.site.healthAuthorityOrganizationId,
        control.value
      ).pipe(
        map((exists: boolean) => exists ? null : { duplicate: true })
      )
    }
  }
```

hogmann (changes requested):

prime-dotnet-webapi/Services/HealthAuthoritySiteService.cs
```diff
@@ -153,5 +153,16 @@ public async Task SiteSubmissionAsync(int siteId)
 
             await _businessEventService.CreateSiteEventAsync(site.Id, "Health Authority Site Submitted");
         }
+
+        public async Task<bool> CheckForDuplicatePecAsync(int healthAuthorityId, string pec)
+        {
+            var site = await _context.HealthAuthoritySites
+                .AsNoTracking()
+                .Where(s => string.Equals(s.PEC, pec)
```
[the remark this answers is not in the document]
AMiKaH: changed back to `==`

prime-angular-frontend/src/app/core/resources/health-authority-resource.service.ts
```diff
@@ -375,6 +375,28 @@ export class HealthAuthorityResource {
       );
   }
 
+  /**
+   * @description
+   * Checks if the provided PEC is already assigned to another site under another Health Authority
+   */
+  public checkPecExistsInOtherHEalthAuthority(healthAuthCode: number, pec: string): Observable<boolean> {
+    // In order to use this method with the AsyncValidator, we're inverting the results here.
```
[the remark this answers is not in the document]
AMiKaH: I tried doing the same yesterday, I just failed with making a name that makes sense, since the error is only when a duplicate is found (204) and if I call it for example `hasNoDuplicate` and negate that it sounds kind of weird. I will see if I can come up with a better naming

prime-angular-frontend/src/app/core/resources/health-authority-resource.service.ts
```diff
@@ -375,6 +375,28 @@ export class HealthAuthorityResource {
       );
   }
 
+  /**
+   * @description
+   * Checks if the provided PEC is already assigned to another site under another Health Authority
+   */
+  public checkPecExistsInOtherHEalthAuthority(healthAuthCode: number, pec: string): Observable<boolean> {
```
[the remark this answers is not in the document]
AMiKaH: Fixed

prime-angular-frontend/src/app/core/resources/health-authority-resource.service.ts
```diff
@@ -375,6 +375,28 @@ export class HealthAuthorityResource {
...
+    // If we get a 404 meaning no duplicate PECs we return true: (No Error)
+    const params = this.apiResourceUtilsService.makeHttpParams({ pec });
+    return this.apiResource.head<boolean>(`health-authorities/${healthAuthCode}/sites/duplicate-pec`, params)
+      .pipe(
+        map(() => false),
+        catchError((error: any) => {
+          if (error.status === 404) {
+            return of(true);
+          }
+          this.logger.error('[Core] HealthAuthorityResource::healthAuthorityPecCheck error has occurred: ', error);
```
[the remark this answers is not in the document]
AMiKaH: Fixed, good eye

prime-angular-frontend/src/app/modules/health-auth-site-reg/pages/site-information-page/site-information-page.component.ts
AMiKaH: I will do that I just have to import the HA resource service in the form file, I didn't do it initially just so I do not include the import in the form file.

prime-dotnet-webapi/Controllers/HealthAuthoritySitesController.cs
AMiKaH: Now using the existing end point and service method for `pecAssignable`

prime-angular-frontend/src/app/core/resources/health-authority-resource.service.ts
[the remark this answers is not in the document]
AMiKaH: Moved to healtha authority site information form

prime-angular-frontend/src/app/modules/health-auth-site-reg/pages/site-information-page/site-information-page.component.ts
AMiKaH: Updated, now the async validator is assigned at form building

prime-angular-frontend/src/app/core/resources/health-authority-resource.service.ts
```diff
@@ -283,10 +283,10 @@ export class HealthAuthorityResource {
...
       .pipe(
         map(({
-               healthAuthoritySite,
-               businessHours,
-               remoteUsers
-             }: { healthAuthoritySite: HealthAuthoritySite, businessHours: BusinessDay[], remoteUsers: RemoteUser[] }) => {
+          healthAuthoritySite,
+          businessHours,
+          remoteUsers
+        }: { healthAuthoritySite: HealthAuthoritySite, businessHours: BusinessDay[], remoteUsers: RemoteUser[] }) => {
```
AMiKaH, line 289: @hogmann @mtpultz Is this indentation also wrong? It appears to be normal for me as it's one Tab up. Just double checking in case my editor is doing its thing again.

hogmann: This looks to be correct

prime-angular-frontend/src/app/modules/health-auth-site-reg/pages/site-information-page/site-information-page.component.html
```diff
@@ -29,7 +29,9 @@
             <mat-label>Site ID/PEC Code</mat-label>
             <input matInput
                    [formControl]="formState.pec">
-            <mat-error>Required</mat-error>
```
hogmann, line 32: I don't see this in the ticket, but is this field not required anymore?

prime-angular-frontend/src/app/modules/health-auth-site-reg/pages/site-information-page/site-information-form.model.ts
```diff
@@ -1,3 +1,3 @@
 import { HealthAuthoritySite } from '@health-auth/shared/models/health-authority-site.model';
 
-export interface SiteInformationForm extends Pick<HealthAuthoritySite, 'siteName' | 'pec' | 'securityGroupCode'> {}
+export interface SiteInformationForm extends Pick<HealthAuthoritySite, 'siteName' | 'pec' | 'securityGroupCode' | 'id'> { }
```
hogmann, line 3: Why is this id being added?

hogmann: Sorry i guess to explain more, since this is not a field on the form we should not have in on this model. I see later on you use this to get the site id. You might want to go another  route to get this id (since it is accessible in route.snapshot.params.sid)

prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -683,15 +683,11 @@ public async Task<ActionResult> GetSiteAdjudicationDocument(int siteId, int docu
         public async Task<ActionResult> PecAssignable(int siteId, string pec)
         {
             var site = await _communitySiteService.GetSiteAsync(siteId);
-            if (site == null)
```
hogmann, line 686: What is the reason to remove this check, Since if site==null here this endpoint returns nothing.

prime-dotnet-webapi/Services/SiteService.cs
```diff
@@ -42,12 +42,30 @@ public async Task<bool> PecAssignableAsync(int siteId, string pec)
...
-            if (siteDto.CareSettingCode == (int)CareSettingType.HealthAuthority
-                || siteDto.PEC == pec)
+            if (siteDto.PEC == pec)
             {
                 return true;
             }
 
+            if (siteDto.CareSettingCode == (int)CareSettingType.HealthAuthority)
+            {
+                var healthAuthorityDto = await _context.HealthAuthoritySites
```
hogmann, line 52: Since you are jsut returning a single field you don't need to make a DTO you can just return the HealthAuthorityOrganizationId

hogmann (changes requested):

prime-angular-frontend/src/app/modules/health-auth-site-reg/pages/site-information-page/site-information-form.model.ts
AMiKaH: Sure I will try find a different way, it was kind of odd that I could see the `id` key in the object in the debugger but can't access it in code so I added it to the `Pick`, Site was null in that page and so I resorted to this, I'll see what else can provide a site ID.

prime-dotnet-webapi/Controllers/SitesController.cs
AMiKaH: It was not being used

prime-angular-frontend/src/app/modules/health-auth-site-reg/pages/site-information-page/site-information-page.component.html
AMiKaH: It was never used

prime-dotnet-webapi/Controllers/SitesController.cs
AMiKaH: Now using `SiteExists()`

prime-dotnet-webapi/Services/SiteService.cs
AMiKaH: Changed to only return the `HealthAuthorityOriganizationId`

prime-angular-frontend/src/app/modules/health-auth-site-reg/pages/site-information-page/site-information-form.model.ts
AMiKaH: Fixed

prime-dotnet-webapi/Services/interfaces/ISiteService.cs
```diff
@@ -29,5 +29,6 @@ public interface ISiteService
         /// Save the fact that the given <c>notifiedUsers</c> were notified by email.
         /// </summary>
         Task MarkUsersAsNotifiedAsync(IEnumerable<RemoteUser> notifiedUsers);
+        Task<bool> SiteExists(int siteId);
```
james-hollinger, line 32: 👍

prime-dotnet-webapi/Services/SiteService.cs
```diff
@@ -37,17 +37,27 @@ public async Task<bool> PecAssignableAsync(int siteId, string pec)
...
-            if (siteDto.CareSettingCode == (int)CareSettingType.HealthAuthority
-                || siteDto.PEC == pec)
+            if (siteDto.PEC == pec)
             {
                 return true;
             }
 
+            if (siteDto.healthAuthorityId.HasValue)
+            {
+                return !await _context.HealthAuthoritySites
```
hogmann, line 52: As this code is written a health authority site can be made that has the same PEC code as a Community site, but a community site cannot be made if a HA site has the same PEC code.

So one of these ways are wrong, Marty has asked Sarah to clarify but that has to be changed.

hogmann (changes requested):

prime-angular-frontend/src/app/modules/health-auth-site-reg/pages/site-information-page/site-information-page.component.html
```diff
@@ -29,6 +29,9 @@
             <mat-label>Site ID/PEC Code</mat-label>
             <input matInput
                    [formControl]="formState.pec">
+            <mat-error *ngIf="formState.pec.hasError('assignable')">
+              Site ID/PEC Code has already been used by another health authority site
```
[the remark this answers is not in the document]
AMiKaH: Updated error message

prime-dotnet-webapi/Services/SiteService.cs
hogmann: Confirmed with Sarah that you cannot have a health authority site that has the same PEC as a community site, so this logic needs to be updated.

AMiKaH: Fixed, now it checks for any sites in other HAs as well as any community site that has the same PEC

prime-dotnet-webapi/Services/SiteService.cs
```diff
@@ -37,17 +37,33 @@ public async Task<bool> PecAssignableAsync(int siteId, string pec)
...
+
+                var otherHealthAuthoritySites = await _context.HealthAuthoritySites
+                    .AsNoTracking()
+                    .Where(
+                        s => s.PEC == pec
+                        && s.HealthAuthorityOrganizationId != siteDto.healthAuthorityId
+                        )
+                    .AnyAsync();
+
+                return !communitySites && !otherHealthAuthoritySites;
```
james-hollinger, line 64: So this totally works, but it evokes one of my favorite phrases: "there is no trap so deadly as the one you set for yourself".

Do you see the trap?

_What happens if we add a third type of site?_

This breaks because we won't be checking the new type of site. And we don't want to have to remember to come back and update this method of we do add another type of site.
I'm cool with the PR as is, I will probably come back and refactor this

AMiKaH: I see, I can update that to retrieve all non HA sites

mtpultz (approved):

mtpultz (approved):

Paahn (approved):
LGTM

james-hollinger (approved):
LGTM

PR #1764, opened 2021-11-10
PRIME-1987 Fix Paper Enrolment Regulatory Select

mtpultz:
The regulatory selection in Paper Enrolment was not appearing due to no options being in the list


AMiKaH (approved):
This will have a conflict with the device provider ticket I believe

mtpultz:
> This will have a conflict with the device provider ticket I believe

Yes, only in that it is in the same file, but not functionally

mtpultz:
> This will have a conflict with the device provider ticket I believe

Only in that it has changes in the same file, but not in functionality which

james-hollinger (approved):
L 📄 TM

PR #1765, opened 2021-11-10
PRIME-1989 Remove Remote Users from HA Site Registration

mtpultz:


mtpultz:
@james-hollinger pulled back the remote users to the migration, and sounds like we should be performing the last bit and moving remote users up into community site and generate a migration.

hogmann (approved):
LGTM

james-hollinger (approved):
L ✂ TM

PR #1768, opened 2021-11-10
PRIME-1951 Part 1 - Remove Claiming from Normal Workflow

mtpultz:
Initial PR for having two workflows one without claiming and another that is used specifically for claiming.  This PR only removes claiming from the normal workflow.

Already applied to TEST


Paahn (approved):
🦔 🦔

johnnyl6 (approved):
🏄‍♂️

PR #1778, opened 2021-11-12
PRIME-2001 inappropriate appearance of Update Business License button

mtpultz:
Create a community site and submit, update the expiry date in the database to push the site into renewal, and the business licence button will be displayed, but after submission when InReview it will not be displayed.

### DIFF
--- a/prime-angular-frontend/src/app/shared/components/site/overview-container/overview-container.component.html
+++ b/prime-angular-frontend/src/app/shared/components/site/overview-container/overview-container.component.html
@@ -72,7 +72,7 @@
                         [editRoute]="SiteRoutes.BUSINESS_LICENCE"
                         (route)="onRouteRelative($event)">
 
-    <button *ngIf="showEditRedirect && withinRenewalPeriod"
+    <button *ngIf="showEditRedirect && site.status !== SiteStatusType.IN_REVIEW && withinRenewalPeriod"
             mat-flat-button
             color="primary"
             class="mb-4"


AMiKaH:
Tested, **Renew Site** button won't show after submission, but will be displayed after changing the expiry date to, for example, today. 

However, when updating the business license to renew, the document is not sent to backend and therefore I get **Site registration could not be submitted** error

AMiKaH (commented):
Tested, Renew Site button won't show after submission, but will be displayed after changing the expiry date to, for example, today.

However, when updating the business license to renew, the document is not sent to backend and therefore I get **Site registration could not be submitted** error

AMiKaH:
Also, the route to `REMOTE_USER` vs `REMOTE_USERS` is causing an issue in this ticket. I saw that it is fixed in another PR

mtpultz:
> Tested, Renew Site button won't show after submission, but will be displayed after changing the expiry date to, for example, today.
> 
> However, when updating the business license to renew, the document is not sent to backend and therefore I get **Site registration could not be submitted** error

Your mentioning other issues that are covered by other tasks in the sprint.  All this PR does is affect the button.  Does it do that or not?

AMiKaH (approved):

AMiKaH:
It does what it's supposed to do. I just mentioned the issue as I came across it. Too many small PRs, may be better to be in one PR that encompasses all the issues so that we can test one and not multiple PRs? Just a thought

PR #1779, opened 2021-11-12
PRIME-2004 change global renewal constant to 30 days

mtpultz:


hogmann (approved):
LGTM

prime-dotnet-webapi/Services/Rules/MinorUpdateRules.cs
```diff
@@ -47,7 +47,7 @@ public override Task<bool> ProcessRule(Enrollee enrollee)
             }
 
             TimeSpan diff = enrollee.ExpiryDate.Value - DateTimeOffset.Now;
-            return Task.FromResult(diff > TimeSpan.FromDays(90));
+            return Task.FromResult(diff > TimeSpan.FromDays(30));
```
mtpultz, line 50: @james-hollinger request was change renewal periods for enrolments and site registration to 30 from 90 days.  This appears to be the only reference to the 90 renewal period for enrolments on the backend, but wanted to double check.

james-hollinger: Fairly certain this is the only place, though we may need to check emails to see if the 90 day renewal is mentioned anywhere.

AMiKaH (approved):

Paahn (approved):
LGTM

prime-dotnet-webapi/Services/Rules/MinorUpdateRules.cs
mtpultz: Only other 90 day refs are around Site inactivity so looks like we're good

PR #1785, opened 2021-11-15
Remove Double Index

james-hollinger:
We will need to update the index in Dev/Test manually, but this will prevent us from having to add the index to Prod twice which is a time-consuming operation with the volume of data we have.


mtpultz (approved):

neophyte57 (approved):
Thank you for fixing!

PR #1800, opened 2021-11-18
PRIME 1982 - Enable Site link on Admin Enrollee Overview -> Remote User

Paahn:
[JIRA Ticket](https://bcgovmoh.atlassian.net/jira/software/c/projects/PRIME/boards/1?modal=detail&selectedIssue=PRIME-1982)

Testing steps:

1. Create a Community Practice Site and declare a Remote User on it.
2. Approve the Site in Admin.
3. Create an Enrollee with Certs matching the Remote User.
4. Claim the Remote User and submit.
5. In Admin, go to the overview for the Enrollee and scroll down to the header for Remote Access
6. Click the link to the site.


prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html
```diff
@@ -61,7 +61,7 @@ <h3>Enrollee Information</h3>
           <app-enrollee-review *ngIf="enrolment"
                                [enrolment]="enrolment"
                                [admin]="true"
-                               (route)="onRoute($event)">
+                               (route)="onRoute($event, true)">
```
mtpultz, line 64: This means on every use of the route emitter within `EnrolleeReview` in admin you will always be redirecting to site registration.  Is that your intent?

If it is and there is only a single route within `EnrolleeReview` that gets emitted in the admin you can be more explicit and rename `onRoute` to `onWhateverYouNeedToDo()` to emphasize it has a singular use in admin and then you don't need any parameters.  Then you can refactor `onNavigateEnrollee` to use `this.routeUtils` directly instead since there isn't much benefit to it calling `onRoute`.

AMiKaH (approved):

AMiKaH:
Tested, looks good!

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html
Paahn: Great point Marty. After some investigation, I found this breaks the routing for the buttons in `EnrolleeReview` when `showEditRedirect = true`, so this implementation has to change.

Paahn: doesnt look like `showEditRedirect` can be `true` in admin, as admins dont edit enrollees information

Paahn: refactored initially, may want to change the way the url is built

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.ts
```diff
@@ -61,10 +63,15 @@ export class EnrolleeOverviewComponent extends AdjudicationContainerComponent im
     this.hasActions = true;
   }
 
-  public onNavigateEnrollee(enrolleeId: number) {
+  public onNavigateEnrollee(enrolleeId: number): void {
     this.onRoute([enrolleeId, RouteUtils.currentRoutePath(this.router.url)]);
   }
 
+  public onRedirectCommunitySite(routePath: string | (string | number)[]): void {
+    const path = `${AdjudicationRoutes.MODULE_PATH}/${AdjudicationRoutes.SITE_REGISTRATIONS}/${routePath[0]}/${routePath[1]}/${routePath[2]}`;
```
mtpultz, line 71: If you need to convert an array of strings to a URL you'll want to first check it's an array, and then use `join`.  You won't need the intermediate `const`.  This will handle an array when provided and make a string, and otherwise it's a string and you don't need to do anything.

```ts
if(Array.isArray(routePath)) {
  routePath = routePath.join('/');
}

this.router.navigate([routePath]);
```

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.ts
```diff
@@ -36,6 +37,7 @@ export class EnrolleeOverviewComponent extends AdjudicationContainerComponent im
   public showAdjudication: boolean;
   public documents: EnrolleeAdjudicationDocument[];
   public absence: EnrolleeAbsence;
+  public AdjudicationRoutes = AdjudicationRoutes;
```
mtpultz, line 40: Only need this if you are using it in the template

mtpultz (approved):

PR #1804, opened 2021-11-19
PRIME-1900 Paper Enrollee Returnee Admin Updates

johnnyl6:
[JIRA ticket](https://bcgovmoh.atlassian.net/browse/PRIME-1900)

Includes a bug fix for enrollee overview page (not populating status reason correctly)

To test linked enrolment
1. Know one enrollee's birthdate beforehand
2. Create a paper enrolment with the known birthdate
3. Create a new enrolment and choose `I requested access via the offline form`, and provide the GPID of the above paper enrolment

To test possible paper enrolment match
- Repeat step 1 & 2 from the above
- Create a new enrolment without choosing `I requested access via the offline form`


neophyte57:
> [JIRA ticket](https://bcgovmoh.atlassian.net/browse/PRIME-1900)
> 
> Includes a bug fix for enrollee overview page (not populating status reason correctly)
> 
> To test linked enrolment
> 
> 1. Know one enrollee's birthdate beforehand
> 2. Create a paper enrolment with the known birthdate
> 3. Create a new enrolment and choose `I requested access via the offline form`, and provide the GPID of the above paper enrolment
> 
> To test possible paper enrolment match
> 
> * Repeat step 1 & 2 from the above
> * Create a new enrolment without choosing `I requested access via the offline form`

Probably un-related to this PR's changes but when the Admin goes Back to the Paper Enrollment Self-Declaration screen, whichever radio buttons were selected are de-selected.  Is that expected?

prime-dotnet-webapi/Models/BusinessEventType.cs
```diff
@@ -18,6 +18,7 @@ public sealed class BusinessEventType : ILookup<int>
         public const int ADMIN_VIEW_CODE = 7;
         public const int ORGANIZATION_CODE = 8;
         public const int PHARMANET_API_CALL_CODE = 9;
+        public const int PAPER_ENROLMENT_LINK_CODE = 10;
```
james-hollinger, line 21: These have been updated on Dev, they are now `CamelCase` rather than `SCREAMING_SNAKE_CASE`

prime-dotnet-webapi/ViewModels/Enrollee/EnrolleeListViewModel.cs
```diff
@@ -19,6 +19,8 @@ public class EnrolleeListViewModel
 
         public int CurrentStatusCode { get; set; }
 
+        public DateTime DateOfBirth { get; set; }
```
james-hollinger, line 22: is this required on this View Model?

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html
```diff
@@ -16,11 +16,27 @@
...
+           class="mb-4 align-self-center">
           <strong>GPID:</strong> {{ enrolment?.enrollee.gpid }}
       </div>
-
+      <div *ngIf="enrollee?.linkedEnrolleeId">
+        <a [routerLink]=""
+           (click)="onRoute([enrollee?.linkedEnrolleeId, AdjudicationRoutes.ENROLLEE_OVERVIEW])">
+          {{ (enrolment?.enrollee.gpid | contains: paperEnrolleeGpidFilter) ? 'Linked' : 'Paper' }} Enrolment
+        </a>
+      </div>
```
neophyte57, line 39: Is the business OK with this difference, as implemented:

image
 
versus mockup:

image

prime-dotnet-webapi/ViewModels/Enrollee/EnrolleeListViewModel.cs
johnnyl6: 👍unused now. should be removed.

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html
johnnyl6: appended the word `link`

prime-dotnet-webapi/Models/BusinessEventType.cs
johnnyl6: merged code from Dev and fixed

johnnyl6:
> > [JIRA ticket](https://bcgovmoh.atlassian.net/browse/PRIME-1900)
> > Includes a bug fix for enrollee overview page (not populating status reason correctly)
> > To test linked enrolment
> > 
> > 1. Know one enrollee's birthdate beforehand
> > 2. Create a paper enrolment with the known birthdate
> > 3. Create a new enrolment and choose `I requested access via the offline form`, and provide the GPID of the above paper enrolment
> > 
> > To test possible paper enrolment match
> > 
> > * Repeat step 1 & 2 from the above
> > * Create a new enrolment without choosing `I requested access via the offline form`
> 
> Probably un-related to this PR's changes but when the Admin goes Back to the Paper Enrollment Self-Declaration screen, whichever radio buttons were selected are de-selected. Is that expected?

From regular enrolment self declaration page code, the radio button patching has a `defaultValue` determined by `profileComplete` evaluated to either `false` or `null`. 

For paper enrolment, `profileComplete` is not applicable so I think it's fair to reset the radio to de-selected whose value is `null`.

https://github.com/bcgov/moh-prime/blob/82915c3d386cce94324d99c4ee25d455b9399b27/prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts#L563

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html
neophyte57: I mean the lack of the GPID above the hyperlink

johnnyl6: GPID is only shown when available, ,i.e. after enrolment is approved.

neophyte57:
This might be outside of the original scope of this PR but on the "Previous Offline Request for PharmaNet Access" page, I can enter a bogus GPID as long as it is 20 characters and starts with "NOBCSC" ... is this OK?

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html
```diff
@@ -16,11 +16,27 @@
...
+           class="mb-4 align-self-center">
           <strong>GPID:</strong> {{ enrolment?.enrollee.gpid }}
       </div>
-
+      <div *ngIf="enrollee?.linkedEnrolleeId">
+        <a [routerLink]=""
+           (click)="onRoute([enrollee?.linkedEnrolleeId, AdjudicationRoutes.ENROLLEE_OVERVIEW])">
+          {{ (enrolment?.enrollee.gpid | contains: paperEnrolleeGpidFilter) ? 'Linked' : 'Paper' }} Enrolment link
+        </a>
+      </div>
```
neophyte57, line 39: It seems like we should never have a situation like this where the alert is shown and the link has also been established.  They seem to be exclusive of each other

image

neophyte57:
image

This seems visually too tight together

Paahn (approved):
✌️🥳🤳

prime-dotnet-webapi/Configuration/Database/Unseeded Tables/EnrolleeConfiguration.cs
```diff
@@ -17,6 +17,12 @@ public void Configure(EntityTypeBuilder<Enrollee> builder)
             builder
                 .HasIndex(e => e.HPDID)
                 .IsUnique();
+            builder
+                .HasOne(e => e.PaperEnrolment)
+                .WithOne(l => l.Enrollee);
+            builder
+                .HasOne(e => e.LinkedEnrolment)
+                .WithOne(l => l.PaperEnrollee);
```
james-hollinger, line 25: I'm concerned by the name of these navigation properties; they are too ambiguous in my opinion. For one thing, they are not the linked `Enrollee` objects, they are the `EnrolleeLinkedEnrolment` objects. This is kinda confusing when you have to do `Enrollee,PaperEnrolment.PaperEnrollee`. It is also a bit ambiguous what `Enrollee.LinkedEnrolment` means on a non-paper `Enrollee`. I would expect that regular Enrollees with a linked paper enrolment would have `Enrollee.PaperEnrolment != null`, but then what is `Enrollee.LinkedEnrolment`? They have a linked enrolment, so it sounds like these are redundant properties.

prime-dotnet-webapi/Migrations/ApiDbContextModelSnapshot.cs
```diff
@@ -260,7 +260,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
...ve, you must promptly (and in any event before the effective date) cease all access or use of\n      PharmaNet.\n    </p>\n\n    <p>\n      Any written notice to you under (i) above will be in writing and delivered by the Province to you using any of the\n      contact mechanisms identified by you in PRIME, including by mail to a specified postal address, email to a\n      specified email address or text message to a specified cell phone number. You may be required to click a URL link\n      or log into PRIME to receive the contents of any such notice.\n    </p>\n\n  </li>\n  {$lcPlaceholder}\n  <li>\n\n    <p class=\"bold underline\">\n      Governing Law\n    </p>\n\n    <p>\n      These terms will be governed by and will be construed and interpreted in accordance with the laws of British\n      Columbia and the laws of Canada applicable therein.\n    </p>\n\n    <p>\n      Unless otherwise specified, a reference to a statute or regulation by name means the statute or regulation of\n      British Columbia of that name, as amended or replaced from time to time, and includes any enactment made under the\n      authority of that statute or regulation.\n    </p>\n\n  </li>\n</ol>\n",
```
james-hollinger, line 263: If you notice, this is replacing "\r\n" with "\n"
This text is populated from the HTML files in Configuration/Database/Resources.  I am confident you will find that these files have the "wrong" line endings on your local based on your git `autocrlf` setting. Normally, when you commit a file back to the repo it will automatically convert the line ending back. But because the file is being serialized here, your local line endings are being copied into the snapshot.

james-hollinger: You should check in with Nathan, he had the same issue.

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html
AMiKaH: There should never be a case where one enrollee is linked to more than one paper enrolment.

prime-dotnet-webapi/Services/SubmissionRulesService.cs
```diff
@@ -48,7 +48,7 @@ public async Task<bool> QualifiesForAutomaticAdjudicationAsync(Enrollee enrollee
                 new IdentityAssuranceLevelRule(),
                 new IdentityProviderRule(),
                 new NoAssignedAgreementRule(),
-                new IsPotentialPaperEnrolleeReturnee(_enrolleePaperSubmissionService),
+                new IsPotentialPaperEnrolleeReturnee(_enrolleePaperSubmissionService, _businessEventService),
```
AMiKaH, line 51: You're passing the constructor's positional arguments arguments in the wrong order
**Passing:** `new IsPotentialPaperEnrolleeReturnee(_enrolleePaperSubmissionService, _businessEventService)`
Constructor: `public IsPotentialPaperEnrolleeReturnee(IBusinessEventService businessEventService,       IEnrolleePaperSubmissionService enrolleePaperSubmissionService)`

prime-dotnet-webapi/Services/Rules/AutomaticAdjudicationRules.cs
```diff
@@ -243,10 +243,14 @@ public override Task<bool> ProcessRule(Enrollee enrollee)
 
     public class IsPotentialPaperEnrolleeReturnee : AutomaticAdjudicationRule
     {
+        private readonly IBusinessEventService _businessEventService;
         private readonly IEnrolleePaperSubmissionService _enrolleePaperSubmissionService;
 
-        public IsPotentialPaperEnrolleeReturnee(IEnrolleePaperSubmissionService enrolleePaperSubmissionService)
+        public IsPotentialPaperEnrolleeReturnee(
+            IBusinessEventService businessEventService,
+            IEnrolleePaperSubmissionService enrolleePaperSubmissionService)
```
AMiKaH, line 251: Same as other comment:
You're passing the constructor's positional arguments arguments in the wrong order
Passing: new IsPotentialPaperEnrolleeReturnee(_enrolleePaperSubmissionService, _businessEventService)
Constructor: public IsPotentialPaperEnrolleeReturnee(IBusinessEventService businessEventService, IEnrolleePaperSubmissionService enrolleePaperSubmissionService)

AMiKaH (changes requested):
Small change but need to correct the order of positional arguments for the `IsPotentialPaperEnrolleeReturnee` rule

AMiKaH:
I am getting multiple possible matches when there's only one paper enrolment with matching birthdate, but no matching GPID
Enrollee table:
image

EnrolleeLinkedEnrolment table:
image

Error:
image

AMiKaH:
If I have 2 paper enrollees with matching birthdates but one of them is linked, I don't get a warning about a potential match at all.

AMiKaH (commented):
If I have 2 paper enrollees with matching birthdates but one of them is linked, I don't get a warning about a potential match at all.

prime-dotnet-webapi/ViewModels/Enrollee/_EnrolleeMappingProfile.cs
```diff
@@ -23,10 +24,14 @@ public EnrolleeMappingProfile()
...
+                .ForMember(dest => dest.PossiblePaperEnrolmentMatch, opt => opt.MapFrom(src => unlinkedPaperEnrolments.Any(e => e.DateOfBirth.Date == src.DateOfBirth.Date)))
                 .ForMember(dest => dest.Confirmed, opt => opt.MapFrom(src => src.Submissions.OrderByDescending(s => s.CreatedDate).FirstOrDefault().Confirmed == true));
 
             CreateMap<Enrollee, EnrolleeDTO>()
                 .ForMember(dest => dest.Confirmed, opt => opt.MapFrom(src => src.Submissions.OrderByDescending(s => s.CreatedDate).FirstOrDefault().Confirmed == true))
+                .ForMember(dest => dest.LinkedEnrolleeId, opt => opt.MapFrom(src => (src.PaperEnrolment == null) ? src.LinkedEnrolment.EnrolleeId : src.PaperEnrolment.PaperEnrolleeId))
+                .ForMember(dest => dest.PossiblePaperEnrolmentMatch, opt => opt.MapFrom(src => (src.GPID != null && src.GPID.Contains("NOBCSC")) ? false : unlinkedPaperEnrolments.Any(e => e.DateOfBirth.Date == src.DateOfBirth.Date)))
                 .ForMember(dest => dest.HasNewestAgreement, opt => opt.MapFrom(src => newestAgreementIds.Any(id => id == src.CurrentAgreementId)));
 
             CreateMap<EnrolleeDTO, EnrolleeViewModel>();
```
AMiKaH, line 37: I left this comment here just so it shows up as a conversation and allows replies.
If I have 2 paper enrollees with matching birthdates but one of them is linked, I don't get a warning about a potential match at all.

AMiKaH, line 37: I am getting "Multiple possible paper enrolment matches" warning when there's only one paper enrolment with matching birthdate, but no matching GPID

**Enrollee table:**
image

**EnrolleeLinkedEnrolment table:**
image

**Warning:**
image

Maybe this can be solved by wording it differently if one potential match is found. There's another ticket that I'm working on where we display a link to the potential match only, but not to the confirmed links

prime-dotnet-webapi/Configuration/Database/Unseeded Tables/EnrolleeConfiguration.cs
johnnyl6: > I'm concerned by the name of these navigation properties; they are too ambiguous in my opinion. For one thing, they are not the linked `Enrollee` objects, they are the `EnrolleeLinkedEnrolment` objects. This is kinda confusing when you have to do `Enrollee,PaperEnrolment.PaperEnrollee`. It is also a bit ambiguous what `Enrollee.LinkedEnrolment` means on a non-paper `Enrollee`. I would expect that regular Enrollees with a linked paper enrolment would have `Enrollee.PaperEnrolment != null`, but then what is `Enrollee.LinkedEnrolment`? They have a linked enrolment, so it sounds like these are redundant properties.

Ideally a single property `LinkedEnrolment`on the `Enrollee` model is the best, which just points to the *linked* enrolllee. For regular Enrollee, `LinkedEnrolment` means a paper enrolment. For paper enrollee, `LinkedEnrolment` means the linked regular enrollee.

However EF does not seem to support using a single property to configure two different relationship indicated by different foreignKey(`FK_EnrolleeLinkedEnrolment_Enrollee_EnrolleeId` and `FK_EnrolleeLinkedEnrolment_Enrollee_PaperEnrolleeId` in table `EnrolleeLinkedEnrolment`). I had to use two separate properties to configure the relationships

prime-dotnet-webapi/ViewModels/Enrollee/_EnrolleeMappingProfile.cs
johnnyl6: Agree. Maybe the wording should be ` Possible paper enrolment matches`

johnnyl6: Good question. Currently linked indicator and potentialMatch indicator are mutually exclusive so only one is displayed. Will confirm.

AMiKaH: I can test it again after your edits, but when I tried it, if any enrollee's birthdate matches linked paper enrollee, the warning will disappear, even though there's another unlinked potential match.

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.ts
```diff
@@ -36,6 +36,7 @@ export class EnrolleeOverviewComponent extends AdjudicationContainerComponent im
   public showAdjudication: boolean;
   public documents: EnrolleeAdjudicationDocument[];
   public absence: EnrolleeAbsence;
+  public readonly paperEnrolleeGpidFilter = 'NOBCSC';
```
hogmann, line 39: In the front end we have a constant.ts. Probably time to move this into that file and in the places we use it for logic we should use it from there.

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html
```diff
@@ -14,13 +14,29 @@
...
       <div *ngIf="enrolment?.enrollee.gpid"
-            class="mb-4 align-self-center">
+           class="mb-4 align-self-center">
           <strong>GPID:</strong> {{ enrolment?.enrollee.gpid }}
       </div>
-
+      <div *ngIf="enrollee?.linkedEnrolleeId">
+        <a [routerLink]=""
+           (click)="onRoute([enrollee?.linkedEnrolleeId, AdjudicationRoutes.ENROLLEE_OVERVIEW])">
+          {{ (enrolment?.enrollee.gpid | contains: paperEnrolleeGpidFilter) ? 'Linked' : 'Paper' }} Enrolment link
```
hogmann, line 37: The word link here is self explanatory, best to remove it. If then it's not to obvious what clicking the link does you can use a verb in front of it like "Go to linked enrolment" "Go to paper enrolment"

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html
```diff
@@ -183,6 +183,12 @@
                           matTooltip="Submission confirmed"
                           matTooltipPosition="after">verified
                 </mat-icon>
+                <mat-icon *ngIf="row.linkedEnrolleeId || row.possiblePaperEnrolmentMatch"
+                          class="paper-submission-icon"
+                          [class]="row.linkedEnrolleeId ? 'linked' : 'possible-match'"
+                          [matTooltip]="row.linkedEnrolleeId ? 'Paper Submission Linked' : 'Possible Paper Enrolment Matches'"
+                          matTooltipPosition="after">task
```
hogmann, line 190: I'm not really onboard with using the same icon with a different colour and tooltip. Something I don't think we have really done.

I would make the suggestion to instead use a different icon like find_in_page. Has a magnifying glass instead of check mark. and then keep the colouring how it is with the rest of the icons that they change with the status.

prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -22,6 +22,8 @@ namespace Prime.Services
 {
     public class EnrolleeService : BaseService, IEnrolleeService
     {
+        private const string PaperGpidPrefix = "NOBCSC";
```
hogmann, line 25: Same with front end this is now used in what looks to be three places. Is it possible to have this defined as a constant somewhere and import from one location?

prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -149,7 +161,7 @@ public async Task<IEnumerable<EnrolleeListViewModel>> GetEnrolleesAsync(Enrollee
                     .Where(e => e.GPID.StartsWith("NOBCSC")
```
james-hollinger, line 161: a place to use the new constant

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html
johnnyl6: Good advice but this would change the acceptance criteria for PRIME-1900. Could become a future improvement ticket.

prime-dotnet-webapi/Configuration/Database/Unseeded Tables/EnrolleeConfiguration.cs
james-hollinger: how about `EnrolleeToPaperLink` and `PaperToEnrolleeLink` for names?

prime-dotnet-webapi/ViewModels/Enrollee/Constants.cs
```diff
@@ -0,0 +1,7 @@
+namespace Prime.ViewModels
+{
+    public static class Constants
+    {
+        public const string PaperGpidPrefix = "NOBCSC";
+    }
+}
```
james-hollinger, line 7: I don't think creating a new global `Constants` class is the best way to do this; and the ViewModels folder is not the right place to put it.

I could see it being added to `PrimeConfiguration`, but really it's an implementation detail of how we are currently distinguishing regular `Enrollee`s from paper `Enrollee`s. Previously I spent some time thinking about how we could have the `enrollee.GPID.StartsWith("NOBCSC")` predicate on the `PaperEnrolleeService` for use in other services but ultimately decided it wasn't worth it. 

I feel either the `PaperEnrolleeService` or the `Enrollee` model should own this constant. I first thought that the service was the correct place for it but now I'm leaning more toward putting it on the `Enrollee` model. At some point in the future an `enrollee.IsPaper` property could be added or something similar.

prime-dotnet-webapi/ViewModels/Enrollee/_EnrolleeMappingProfile.cs
```diff
@@ -23,10 +24,14 @@ public EnrolleeMappingProfile()
...
                 .ForMember(dest => dest.Confirmed, opt => opt.MapFrom(src => src.Submissions.OrderByDescending(s => s.CreatedDate).FirstOrDefault().Confirmed == true));
 
             CreateMap<Enrollee, EnrolleeDTO>()
                 .ForMember(dest => dest.Confirmed, opt => opt.MapFrom(src => src.Submissions.OrderByDescending(s => s.CreatedDate).FirstOrDefault().Confirmed == true))
+                .ForMember(dest => dest.LinkedEnrolleeId, opt => opt.MapFrom(src => (src.PaperEnrolment == null) ? src.LinkedEnrolment.EnrolleeId : src.PaperEnrolment.PaperEnrolleeId))
+                .ForMember(dest => dest.PossiblePaperEnrolmentMatch, opt => opt.MapFrom(src => (src.GPID != null && src.GPID.Contains(Constants.PaperGpidPrefix)) ? false : unlinkedPaperEnrolments.Any(e => e.DateOfBirth.Date == src.DateOfBirth.Date)))
                 .ForMember(dest => dest.HasNewestAgreement, opt => opt.MapFrom(src => newestAgreementIds.Any(id => id == src.CurrentAgreementId)));
 
             CreateMap<EnrolleeDTO, EnrolleeViewModel>();
```
AMiKaH, line 37: This is not an issue in the PR and wasn't mentioned in the ticket at the time of creation.
image

Our warning is telling the admin to search, but I have a ticket that will display the potential matches on the right side, under Adjudication >> Status Reasons.

Not sure if changing it to "Use the Enrollee search to determine which paper enrolments match, or see the related paper enrolment under the Adjudication card" would be good, probably will need client feed back

@mtpultz, any thoughts?

prime-dotnet-webapi/ViewModels/Enrollee/Constants.cs
johnnyl6: Moving the constant to `Enrollee` model for now since it's shared by both `EnrolleeService` and `ErnolleePaperSumissionService`

prime-dotnet-webapi/ViewModels/Enrollee/_EnrolleeMappingProfile.cs
mtpultz: I would change the wording but keep the alert.  Don't use the word "card" though that's dev speak and means nothing to the average user.  Might be better to leave it as "... see the possible matches linked under Adjudication" then let the client choose what they want displayed.

mtpultz: or even "... see the possible matches linked under the Adjudication section"

AMiKaH (approved):

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html
hogmann: Sorry I fail to see how this solution reflects the acceptance criteria and mockups. The mockups fail to show the 
``` possibleMatch && unlinked ```
state and the only reference in the criteria is as follows
```a. Multiple results unlinked icon to yellow```

So this would lead me to believe we should have an "unlinked" icon. 

Also with our current state of the admin all icons are matched to the colour of the status. This is how currently every icon is done. 
image
So looking at this an admin might assume that the first enrollee is linked and all good since it has the "linked" icon and in the review state. Since the colour of the icon could be somewhat hidden by the presumption of the status being the same colour. 

So if we through all other facets of admin used different icons to display different "states" and colour is inherited from the status, changing those presumptions for this one thing is wrong and an oversite of this ticket.

File: prime-dotnet-webapi/Configuration/Database/Unseeded Tables/EnrolleeConfiguration.cs (hunk 5 above)
Comment (johnnyl6) [reply prime-dotnet-webapi/Configuration/Database/Unseeded Tables/EnrolleeConfiguration.cs:25]: > 

Done

Comment (johnnyl6) [reply prime-dotnet-webapi/Configuration/Database/Unseeded Tables/EnrolleeConfiguration.cs:25]: > 

Done

File: prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html (hunk 4 above)
Comment (johnnyl6) [reply prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html]: Made them mutually exclusive

File: prime-dotnet-webapi/Migrations/ApiDbContextModelSnapshot.cs [hunk 17]
```diff
@@ -11682,8 +11689,12 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
                     b.Navigation("IdentificationDocuments");
 
+                    b.Navigation("LinkedEnrolment");
+
                     b.Navigation("OboSites");
 
+                    b.Navigation("PaperEnrolment");
+
```
james-hollinger, line 11697: Need to re-generate the migration after changing the navigation property names

johnnyl6: > 

Nice catch! Re-generated.

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html
mtpultz: @hogmann I agree, but that's what they wanted and had us write up in the ticket during a refinement.  It wasn't our teams idea to change the colour and at the time we questioned it since it is based on status, but Paper Enrollees are considered "different".  One ticket in the backlog is to change the status message on Paper Enrolments since a Paper Enrollee can never be "Under Review" it is either "Incomplete" (being filled in by Admin) or "Completed" (submitted and finalized) making it unrelated to any status.  The only suggestion of ours that made it in was the alert at the top with a message to flush out the meaning.

prime-angular-frontend/src/app/lib/constants.ts
```diff
@@ -8,3 +8,8 @@ export const MINIMUM_AGE = 18;
  * Renewal period in days.
  */
 export const RENEWAL_PERIOD = 30;
+/**
+ * @description
+ * Paper enrollee GPID filter.
+ */
+export const PAPER_ENROLLEE_GPID_FILTER = 'NOBCSC';
```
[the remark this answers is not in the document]
johnnyl6: Refactored.

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html
hogmann: Sounds good ill close this conversation, but will probably bring it up down the road again with that ticket.

prime-angular-frontend/src/app/modules/adjudication/pages/enrollee-overview/enrollee-overview.component.html
```diff
@@ -14,13 +14,29 @@
                       (navigateEnrollee)="onNavigateEnrollee($event)"
                       (sendBulkEmail)="onSendBulkEmail()"></app-enrollee-table>
 
-  <div class="d-flex justify-content-between align-items-center mt-4">
+  <div class="d-flex justify-content-between align-items-center mt-4 mb-4">
     <span>
+      <app-alert *ngIf="!enrollee?.linkedEnrolleeId && enrollee?.possiblePaperEnrolmentMatch"
```
mtpultz, line 19: We should try and constrain the alert.  Have it be the same width as the Enrolment overview that is contained underneath it so we have separation between it and the TOA Selection.

image

johnnyl6: Fixed

mtpultz: That's actually not how you're supposed to use `col-6` that's a Bootstrap Grid class and you're using it within Flexbox.  The reason it appears to work is due to the grid being built using flexbox, but `col-6` should always be child of `row` and it also applies a gutter.

This will help with reducing the presence of the alert on the page, provide space between elements, and improving readability.  I took out the alert title since it seems a bit large for our needs and replaced it with a block-level strong container with the title, as well as, a small container for the helper text then we don't need to play with widths to have the justify-content-between work for us since this reduces the content size associated space.  I reduced the text since they found the paper enrollee likely through searching so we can simply point out what they might not know about the adjudication section which will be updated to links in the other PR.

```html
<div class="d-flex justify-content-between align-items-center my-4">
  <span>
    <app-alert *ngIf="!enrollee?.linkedEnrolleeId && enrollee?.possiblePaperEnrolmentMatch"
               type="warning">
      <ng-container #alertContent
                    class="alert-content">
        <strong class="d-block">Possible paper enrolment matches found</strong>
        <small>See the possible matches linked under the Adjudication section.</small>
      </ng-container>
    </app-alert>
    ...
  </span>
  ...
</div>
```

mtpultz: image

mtpultz (approved):

PR #1828, opened 2021-11-25
PRIME-1767 Send Nag Emails if TOAs aren't Signed within 5 days

neophyte57:
| Scenario                                                                                                                                                                                 | Expected                                              |
| ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------- |
| 1\. As enrollee, submit enrollment<br>2\. As admin, approve enrollment<br>3\. Use Postman to hit end-point (or wait for cron job to execute)                                             | Checking Mailhog, no reminder emails are sent out yet |
| 1\. Use database client to adjust approval date (related \`EnrolmentStatus\` rows) to more than 5 days in the past<br>2\. Use Postman to hit end-point (or wait for cron job to execute) | Checking Mailhog, reminder emails were sent out       |

**Related OpenShift cron job commands**
`oc get cronjob`
`oc delete cronjob send-unsigned-toa-emails-cronjob`
`oc process -f .\send-unsigned-toa-emails.cron.yml -p ENV_NAME=pr-1828 -p CRON_SCHEDULE="*/15 * * * *" | oc apply -f -`


prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -246,6 +246,39 @@ public async Task SendEnrolleeRenewalEmails()
...
+                .Include(e => e.EnrolmentStatuses)
+                .Select(e => new
+                {
+                    e.FirstName,
+                    e.LastName,
+                    e.Email,
+                    e.ApprovedDate,
+                    e.CurrentStatus.StatusCode
+                })
+                // were approved
```
AMiKaH, line 262: just a minor typo w**h**ere

AMiKaH (approved):

prime-dotnet-webapi/Services/EmailService.cs
hogmann, line 262: These comments are self explanatory by the lines that follow.

hogmann (approved):
LGTM other than un-needed comments

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -246,6 +246,39 @@ public async Task SendEnrolleeRenewalEmails()
...
+                .Where(e => e.ApprovedDate != null)
```
james-hollinger, line 263: I feel `e.ApprovedDate.HasValue` reads better

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -246,6 +246,39 @@ public async Task SendEnrolleeRenewalEmails()
             }
         }
 
+
+        public async Task SendEnrolleeUnsignedToaReminderEmails()
+        {
+            var enrollees = await _context.Enrollees
+                .Include(e => e.EnrolmentStatuses)
+                .Select(e => new
```
james-hollinger, line 254: doing the `.Select` statement first and then the `.Where` afterwards causes extra SQL to be emitted by EF, since the `CurrentStatus` needs to be both SELECTed and WHEREd. Compare the generated SQL:
```sql
SELECT e1."FirstName", e1."LastName", e1."Email", (
    SELECT e."StatusDate"
    FROM "EnrolmentStatus" AS e
    WHERE (e1."Id" = e."EnrolleeId") AND (e."StatusCode" = 3)
    ORDER BY e."StatusDate" DESC, e."Id" DESC
    LIMIT 1) AS "ApprovedDate", (
    SELECT e0."StatusCode"
    FROM "EnrolmentStatus" AS e0
    WHERE e1."Id" = e0."EnrolleeId"
    ORDER BY e0."StatusDate" DESC, e0."Id" DESC
    LIMIT 1) AS "StatusCode"
FROM "Enrollee" AS e1
WHERE ((
    SELECT e2."StatusDate"
    FROM "EnrolmentStatus" AS e2
    WHERE (e1."Id" = e2."EnrolleeId") AND (e2."StatusCode" = 3)
    ORDER BY e2."StatusDate" DESC, e2."Id" DESC
    LIMIT 1) IS NOT NULL) AND ((
    SELECT e3."StatusCode"
    FROM "EnrolmentStatus" AS e3
    WHERE e1."Id" = e3."EnrolleeId"
    ORDER BY e3."StatusDate" DESC, e3."Id" DESC
    LIMIT 1) = 3)
```
to this SQL:
```sql
SELECT e0."FirstName", e0."LastName", e0."Email", (
    SELECT e."StatusDate"
    FROM "EnrolmentStatus" AS e
    WHERE (e0."Id" = e."EnrolleeId") AND (e."StatusCode" = 3)
    ORDER BY e."StatusDate" DESC, e."Id" DESC
    LIMIT 1) AS "ApprovedDate"
FROM "Enrollee" AS e0
WHERE (
    SELECT TRUE
    FROM "EnrolmentStatus" AS e1
    WHERE (e0."Id" = e1."EnrolleeId") AND (e1."StatusCode" = 3)
    ORDER BY e1."StatusDate" DESC, e1."Id" DESC
    LIMIT 1) AND ((
    SELECT e2."StatusCode"
    FROM "EnrolmentStatus" AS e2
    WHERE e0."Id" = e2."EnrolleeId"
    ORDER BY e2."StatusDate" DESC, e2."Id" DESC
    LIMIT 1) = 3)
```
generated when the `.Where` clause is moved up:
```C#
var enrollees = await _context.Enrollees
    .Where(e => e.ApprovedDate.HasValue
        && e.CurrentStatus.StatusCode == (int)StatusType.RequiresToa)
    .Select(e => new
    {
        e.FirstName,
        e.LastName,
        e.Email,
        e.ApprovedDate
    })
    .DecompileAsync()
    .ToListAsync();
```

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -246,6 +246,39 @@ public async Task SendEnrolleeRenewalEmails()
...
+                .Where(e => e.ApprovedDate != null)
+                // enrolees that are in “Requires TOA” Status
+                .Where(e => e.StatusCode == (int)StatusType.RequiresToa)
+                .DecompileAsync()
+                .ToListAsync();
+
+            foreach (var enrollee in enrollees)
+            {
+                // were approved more than 5 days ago
+                if ((DateTime.Now - (enrollee.ApprovedDate ?? default)).TotalDays > 5)
```
james-hollinger, line 272: You are filtering by `ApprovedDate != null` in the query, so the null coalescing operator here isn't necessary. You can just do `ApprovedDate.Value`.

Also, this should probably be `DateTimeOffset.Now`

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -246,6 +246,39 @@ public async Task SendEnrolleeRenewalEmails()
             }
         }
 
+
+        public async Task SendEnrolleeUnsignedToaReminderEmails()
+        {
+            var enrollees = await _context.Enrollees
+                .Include(e => e.EnrolmentStatuses)
```
james-hollinger, line 253: Generally, you don't need to do `.Include` and `.Select` in the same query. `.Select` does a projection, which tells EF to join the tables it needs to get the data. It will automatically join to the `EnrolmentStatus` table to compute `ApprovedDate`, so this `.Include`is ignored.

prime-dotnet-webapi/Services/EmailService.cs
james-hollinger: Logic wise, I think it's more clear to remove the concept of `ApprovedDate` from this method entirely. Conceptually I think it makes more sense to base this check off the date of the Enrollee's RequiresTOA status rather than the `ApprovedDate`. Under the covers, the `ApprovedDate` _is_ the date of the Enrollee's RequiresTOA status so the logic ends up being the same. Here's what I'm thinking:
```C#
var enrollees = await _context.Enrollees
    .Where(e => e.CurrentStatus.StatusCode == (int)StatusType.RequiresToa)
    .Select(e => new
    {
        e.FirstName,
        e.LastName,
        e.Email,
        e.CurrentStatus.StatusDate
    })
    .DecompileAsync()
    .ToListAsync();
```

james-hollinger: My reasoning behind this is that attaching this to `ApprovedDate` rather than the date of the RequiresTOA status increases the odds that this method needs to change in the future. Hypothetically, if the business decided that the `ApprovedDate` should not be the date that the Enrollee was _first_ approved rather than most recently approved, that would break this feature and it would have to be re-written.

james-hollinger (approved):
LGTM

mtpultz (approved):

PR #1830, opened 2021-11-26
PRIME -2025 Error during Enrollment if go Back from Care Setting screen without choosing any care setting

AMiKaH:
[PRIME-2025](https://bcgovmoh.atlassian.net/jira/software/c/projects/PRIME/boards/1?modal=detail&selectedIssue=PRIME-2025)

### How to test:

1. Login as a brand new enrollee
2. Enter phone# and email
3. Save and continue
4. do not select any care settings
5. click back (to bcsc-demographics)
6. click Save and Continue

you should go back to Care Settings page without issues

You can try adding multiple care settings without making a selection, or mix and match between selections then repeat steps 6 and 7

### DIFF
--- a/prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
+++ b/prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
@@ -228,7 +228,6 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
       const control = this.careSettings.controls[index];
 
       if (!value.careSettingCode && control?.invalid) {
-        const index = this.careSettings.value.findIndex((value) => value.careSettingCode === null);
         this.removeCareSetting(index, value.careSettingCode);
       }
     })
--- a/prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
+++ b/prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
@@ -224,7 +224,7 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
   private removeIncompleteCareSettings(allowEmptyCareSettings: boolean = false) {
     this.enrolmentFormStateService.json.careSettings.forEach((careSetting: CareSetting) => {
       const value = careSetting.careSettingCode;
-      const index = this.careSettings.value.findIndex((value) => value.careSettingCode === null);
+      const index = this.careSettings.value.findIndex((value) => !value.careSettingCode);
       const control = this.careSettings.controls[index];
 
       if (!value && control?.invalid) {
--- a/prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
+++ b/prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
@@ -162,7 +162,7 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
   }
 
   public ngOnDestroy() {
-    this.removeIncompleteCareSettings(true);
+    this.removeIncompleteCareSettings();
   }
 
   protected createFormInstance() {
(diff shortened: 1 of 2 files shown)


prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
```diff
@@ -233,16 +221,20 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
     super.nextRouteAfterSubmit(nextRoutePath);
   }
 
-  private removeIncompleteCareSettings() {
-    this.careSettings.controls
-      .forEach((control: FormGroup, index: number) => {
-        const value = control.get('careSettingCode').value;
+  private removeIncompleteCareSettings(allowEmptyCareSettings: boolean = false) {
+    this.enrolmentFormStateService.json.careSettings.forEach((careSetting: CareSetting) => {
```
mtpultz, line 225: You've got a solution, but does it make for clean code?  You're looping over a list and then within the loop using a method that loops over a list to find an index to be removed based on a predicate.

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
```diff
@@ -233,16 +221,20 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
...
+      const value = careSetting;
+      const index = this.careSettings.value.findIndex((value) => !value.careSettingCode);
```
mtpultz, line 227: Using `value` in the outer scope and inner scope is called variable shadowing.  It should be avoided.

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
```diff
@@ -233,16 +221,20 @@ export class CareSettingComponent extends BaseEnrolmentProfilePage implements On
     super.nextRouteAfterSubmit(nextRoutePath);
   }
 
-  private removeIncompleteCareSettings() {
-    this.careSettings.controls
-      .forEach((control: FormGroup, index: number) => {
-        const value = control.get('careSettingCode').value;
+  private removeIncompleteCareSettings(allowEmptyCareSettings: boolean = false) {
```
mtpultz, line 224: As I was mentioning you can choose whatever you want for this so you don't have to pass a value yourself and the default works for your use case.  What I provided was just an example and it means you have to pass true, but I was trying to explain you can pass nothing when invoking and have the default work for you.

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
mtpultz: You're using the JSON, but remember when we were talking in Slack I indicated you just need `this.careSetting.value` not invoking the `.json` getter.

mtpultz: Also, I mentioned that you would use `forEach` but more needs to happen before you use it.  Also, that you were on the right track with filter, but didn't take it far enough.

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
AMiKaH: The reason I chose to pass a value is so it doesn't affect existing code that doesn't provide value, and the default value in my first pass was to run through the code as before so as not to cause any regression issues. Seeing that this method is only called once, I can revert to passing nothing and have the default value skip the code that adds one site, or we can remove that code as discussed. I am not sure if it will be needed in the future, but will keep it there just in case and have the default skip the code block at line 241.

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
AMiKaH: fixed,

prime-angular-frontend/src/app/modules/enrolment/pages/care-setting/care-setting.component.ts
AMiKaH: Refactored as discussed, made minor changes to cover certain cases

mtpultz (approved):

Paahn (approved):

PR #1832, opened 2021-11-29
Bug - fix error caused by undefined property

johnnyl6:
Bug was introduced by PRIME-1805. When enrollee table is used with single row data e.g. enrollee overview page, sort property is undefined.


AMiKaH (approved):

Paahn (approved):

hogmann (approved):
LGTM

PR #1841, opened 2021-12-01
Fix - Read PLR extract as csv file

neophyte57:
| Scenario                                                            | Expected                                                                                                                    |
| ------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------- |
| Something about row is unparseable (e.g.  a date value)             | Saving row fails and log indicates which row by IPC                                                                         |
| Input file has header row and invalid rows                          | Correct number of records, i.e. excluding header row and invalid rows but including all others                              |
| Row has unusual but valid date (e.g. "1899-12-31")                  | Row is saved, including unusual date                                                                                        |
| Data has commas within fields                                       | All data in field is saved to the correct column                                                                            |
| Data has fields (Expertise, Credentials) with pipe-delimited values | Data is saved in database as JSON array                                                                                     |
| Data has date values                                                | Blank date cells are treated as NULL, 9999-12-30 is treated as NULL, and parseable dates are saved properly in the database |
| All fields for a row are populated                                  | Are values are saved properly                                                                                               |

### DIFF
--- a/utilities/PlrIntakeUtility/PlrIntakeUtility.csproj
+++ b/utilities/PlrIntakeUtility/PlrIntakeUtility.csproj
@@ -7,7 +7,7 @@
   <ItemGroup>
     <PackageReference Include="Serilog.Sinks.File" Version="4.1.0" />
     <PackageReference Include="System.Text.Encoding.CodePages" Version="4.5.1" />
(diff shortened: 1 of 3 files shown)


itchison (approved):
lgtm

mtpultz (approved):

PR #1845, opened 2021-12-03
PRIME-1972 site registration (admin) - search on signing authority

AMiKaH:
[PRIMET-1972](https://bcgovmoh.atlassian.net/jira/software/c/projects/PRIME/boards/1?modal=detail&selectedIssue=PRIME-1972)

How to test:

1. Create a few sites with different names and signing authorities
2. Login as admin
3. Got to Site Registration
4. In the search box search using the signing authority name (test with partial names, complete first and last names, full name i.e. "firstname lastname" You will probably need to change the first names of our test cards in DB since they will all match the first name as they all have PRIMET as their first name)


prime-dotnet-webapi/Infrastructure/TextSearchQuery.cs
```diff
@@ -34,11 +34,24 @@ public TextSearchQuery<T> SearchCollections(params Expression<Func<T, IEnumerabl
         public IQueryable<T> Containing(string term)
         {
             term = term.ToLower();
+            var searchWords = term.Split(' ');
             var predicate = PredicateBuilder.New<T>();
 
             foreach (var selector in _propertySelectors)
             {
-                predicate.Or(x => selector.Invoke(x).ToLower().Contains(term));
+                // This block allows searching and matching for partial names for example Jo Do when searching for Jone Doe
```
hogmann, line 42: So for this user story it is a signing authority phoning in and giving prime support their name to find their admin entry. I think for this case we can assume correct spelling of the name and just use the contains logic that already exists. Since this looks like you would get some weird results. Since you don't differentiate the first and second search term to be first and last name.

So when a admin types "Jo Do" it would return 

John Smith
Don Jeremy
Jone Doe
Nathan Johnson

prime-dotnet-webapi/Services/OrganizationService.cs
```diff
@@ -57,7 +57,9 @@ public async Task<IEnumerable<OrganizationSearchViewModel>> GetOrganizationsAsyn
                 .If(!string.IsNullOrWhiteSpace(searchOptions.TextSearch), q => q
                     .Search(
                         o => o.Name,
-                        o => o.DisplayId.ToString())
+                        o => o.DisplayId.ToString(),
+                        o => o.SigningAuthority.FirstName,
+                        o => o.SigningAuthority.LastName)
```
hogmann, line 62: Cas also add 
``` o => o.SigningAuthority.FirstName + " " + o.SigningAuthority.LastName) ```
To allow for more refined search on whole name so that they will most certainly filter to what they want to see.

Comment (AMiKaH) [reply prime-dotnet-webapi/Services/OrganizationService.cs:62]: Hmm let me try that, I tried the otherway with "${o.SigningAuthority.FirstName} ${o.SigningAuthority.FirstName}" and code didn't like that

File: prime-dotnet-webapi/Infrastructure/TextSearchQuery.cs (hunk 1 above)
Comment (AMiKaH) [reply prime-dotnet-webapi/Infrastructure/TextSearchQuery.cs:42]: If we use the existing contain, the check won't match anything way more often. Since the two fields are searched independently, writing Jone D won't even find the proper Jone Doe if it already exists

Comment (AMiKaH) [reply prime-dotnet-webapi/Infrastructure/TextSearchQuery.cs:42]: Hmm this won't allow for partial searches

**Review by johnnyl6 — APPROVED** [review]
LGTM and works🎉

File: prime-dotnet-webapi/Services/OrganizationService.cs [hunk 3]
```diff
@@ -57,7 +57,8 @@ public async Task<IEnumerable<OrganizationSearchViewModel>> GetOrganizationsAsyn
                 .If(!string.IsNullOrWhiteSpace(searchOptions.TextSearch), q => q
                     .Search(
                         o => o.Name,
-                        o => o.DisplayId.ToString())
+                        o => o.DisplayId.ToString(),
+                        o => o.SigningAuthority.FirstName + " " + o.SigningAuthority.LastName)
```
james-hollinger, line 61: Need to also update `OrganizationListViewModel.MatchedOn()`

AMiKaH: Updated, please check, since we're checking both fields combined I added `SigningAuthority.FirstName` and `SigningAuthority.LastName` together using the same logic

prime-dotnet-webapi/ViewModels/SiteRegistration/OrganizationListViewModel.cs
```diff
@@ -41,6 +41,12 @@ public IEnumerable<string> MatchedOn(string textSearch)
                 matchedOn.Add(nameof(Name));
             }
 
+            if (SigningAuthority != null && (SigningAuthority.FirstName + " " + SigningAuthority.LastName).Contains(textSearch))
+            {
+                matchedOn.Add(SigningAuthority.FirstName);
+                matchedOn.Add(SigningAuthority.LastName);
```
james-hollinger, line 47: This is just the name of the field, it doesn't make sense for the object to come back and say
`Matched on: James`
This can just be `matchedOn.Add("SigningAuthorityName");` or something like that

prime-dotnet-webapi/ViewModels/SiteRegistration/OrganizationListViewModel.cs
```diff
@@ -41,6 +41,12 @@ public IEnumerable<string> MatchedOn(string textSearch)
                 matchedOn.Add(nameof(Name));
             }
 
+            if (SigningAuthority != null && (SigningAuthority.FirstName + " " + SigningAuthority.LastName).Contains(textSearch))
```
james-hollinger, line 44: without converting the Signing Authority's name to lower case this will likely never match

AMiKaH: Converted to lower

prime-dotnet-webapi/ViewModels/SiteRegistration/OrganizationListViewModel.cs
AMiKaH: Changed, did it as the other ones with `matchedOn.Add(nameof(SigningAuthority))`

AMiKaH: Changed

Paahn (approved):
LGTM

hogmann (approved):
LGTM

PR #1848, opened 2021-12-06
PRIME-2015 Make. Email. Fast.

johnnyl6:
Some areas to test speed of sending emails: (where emails are sent after)
- Submit a paper enrolment
- Submit a health authority site
- Approve an organization claim in admin
- Share enrolment approval via email on summary page of enrolment
- Submit community site registration
- Business licence is uploaded
- Remote user is updated
- Site is reviewed by an assigned adjudicator
- Site is active before registration
- Site is approved

### DIFF
--- a/prime-dotnet-webapi/Consumer/SendEnrolleeEmailConsumer.cs
+++ b/prime-dotnet-webapi/Consumer/SendEnrolleeEmailConsumer.cs
@@ -2,7 +2,7 @@ using System.Threading.Tasks;
 
 using MassTransit;
 using MassTransit.Definition;
-
+using Microsoft.Extensions.Logging;
 using Prime.Contracts;
 using Prime.Services;
 
@@ -10,11 +10,16 @@ namespace Prime.Consumer
 {
     public class SendEnrolleeEmailConsumer : SendEmailConsumerBase, IConsumer<SendEnrolleeEmail>
     {
-        public SendEnrolleeEmailConsumer(IEmailService emailService) : base(emailService)
-        { }
+        protected readonly ILogger _logger;
+
+        public SendEnrolleeEmailConsumer(IEmailService emailService, ILogger<SendEnrolleeEmailConsumer> logger) : base(emailService)
+        {
+            this._logger = logger;
+        }
 
         public async Task Consume(ConsumeContext<SendEnrolleeEmail> context)
         {
+            _logger.LogDebug("SendEnrolleeEmailConsumer.Consume called ...");
             switch (context.Message.EmailType)
             {
                 case EnrolleeEmailType.Reminder:
@@ -32,6 +37,7 @@ namespace Prime.Consumer
                 default:
                     break;
             }
+            _logger.LogDebug("SendEnrolleeEmailConsumer.Consume completed");
         }
     }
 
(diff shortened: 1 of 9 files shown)


itchison (changes requested):
First off - I think this PR does a good job at showing the value that a queue can add for our long running processes.

What I dislike is how aware of the bus the consumers of email need to be. We should be able to add this bus mechanism only into the emailservice.cs and queueing only the emailrender and the subsequent send.

The recreation of the models seems like a smell. I feel we'll have to keep these in sync going forward? I'm not sure I love it.

I'm also very curious about about this approach vs a background worker. I think masstransit is a better building block but since we're only implementing in-memory a background thread might also work.

And finally - What happens if an exception is thrown within a masstransit thread? Does it have access to our loggers and subsequently the console?

johnnyl6:
> First off - I think this PR does a good job at showing the value that a queue can add for our long running processes.
> 
> What I dislike is how aware of the bus the consumers of email need to be. We should be able to add this bus mechanism only into the emailservice.cs and queueing only the emailrender and the subsequent send.
> 
Got similar feedback from @james-hollinger ServiceBus is injected at controller level. I have created a IEmailDispatchService to remove the ServiceBus dependency now. IEmailDispatchService is responsible for dispatch Emails to ServiceBus queue.

EmailRender and generating the Attachment are procedures returning email string & binary file respectively. The result are then fed to the SMTP client for sending. It's simpler to treat them as one unit of work as one message into the queue.

> The recreation of the models seems like a smell. I feel we'll have to keep these in sync going forward? I'm not sure I love it.
> 
> I'm also very curious about about this approach vs a background worker. I think masstransit is a better building block but since we're only implementing in-memory a background thread might also work.

Currently all message consumers are in the same process of ASP.NET. A console app to include the consumers in a separate pod should also work, with some configurations in place.

> 
> And finally - What happens if an exception is thrown within a masstransit thread? Does it have access to our loggers and subsequently the console?

There is a ConsumeObserver which has ILogger injected. The observer have `PreConsume` `PostConsume` and `ConsumeFault` to monitor/log the message consuming process for before, after and exception respectively.

johnnyl6:
> First off - I think this PR does a good job at showing the value that a queue can add for our long running processes.
> 
> What I dislike is how aware of the bus the consumers of email need to be. We should be able to add this bus mechanism only into the emailservice.cs and queueing only the emailrender and the subsequent send.
> 
> The recreation of the models seems like a smell. I feel we'll have to keep these in sync going forward? I'm not sure I love it.
> 
> I'm also very curious about about this approach vs a background worker. I think masstransit is a better building block but since we're only implementing in-memory a background thread might also work.
I am not sure how background worker could fit into MassTransit. The official documentation supports 3 types of configuration

> 
> And finally - What happens if an exception is thrown within a masstransit thread? Does it have access to our loggers and subsequently the console?



> First off - I think this PR does a good job at showing the value that a queue can add for our long running processes.
> 
> What I dislike is how aware of the bus the consumers of email need to be. We should be able to add this bus mechanism only into the emailservice.cs and queueing only the emailrender and the subsequent send.
> 
> The recreation of the models seems like a smell. I feel we'll have to keep these in sync going forward? I'm not sure I love it.
> 
> I'm also very curious about about this approach vs a background worker. I think masstransit is a better building block but since we're only implementing in-memory a background thread might also work.
> 
> And finally - What happens if an exception is thrown within a masstransit thread? Does it have access to our loggers and subsequently the console?

@itchison could you advise what the next steps would be for this PR? Since in-memory queue is not recommended for production system, we could either open new tickets for PR feedback work and RabbitMQ implementation, Or close this PR now and keep the branch for future reference.

PR #1850, opened 2021-12-06
PRIME-1682 Health Authority Organization Registration: Care Type/Vendor Deletions

AMiKaH:
[PRIME-1682](https://bcgovmoh.atlassian.net/jira/software/c/projects/PRIME/boards/1?modal=detail&selectedIssue=PRIME-1682)

### How to test:
1. Login as admin
2. Add the information needed for one of the health authorities (make sure you add more than one care type and vendor)
3. Create a health authority site (take note of the site's vendor and care type) (This may require that you create an authorized user and approve them before you can create/submit the site)
4. Submit site
5. Log in as admin again, go to **Site Registrations** >> **Health Authorities**
6. Click the kebab menu of the health authority you used in step one and click "Add Organization Information"
7. Edit **Health Authority Care Types:** 
-- Try to remove the care type that the site from step 4 is using (you should get a toast that says "Health authority care type could not be deleted")
-- Try to delete a care type that isn't used by any site (it should be deleted)
-- Click **Save and Continue** the updated care types list should be displayed in the **Organization Information**
-- Existing care types should not be editable
8. Edit the **Health Authority Vendors:**
-- Try to remove the care type that the site from step 4 is using (you should get a toast that says "Health authority vendor could not be deleted")
-- Try to delete a care type that isn't used by any site (it should be deleted)
-- Click **Save and Continue** the updated vendors list should be displayed in the **Organization Information**
-- Existing vendors should not be editable


prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
```diff
@@ -240,5 +240,62 @@ public async Task<ActionResult> UpdatePharmanetAdministratorContacts(int healthA
...
             return NoContent();
         }
+
+        // HEAD: api/health-authorities/5/vendor/5
+        /// <summary>
+        /// Checks if the specified vendor is in use by a Health Authority site
+        /// </summary>
+        /// <param name="healthAuthorityId"></param>
+        /// <param name="vendorCode"></param>
+        [HttpHead("{healthAuthorityId}/vendor/{vendorCode}", Name = nameof(IsVendorInUse))]
```
james-hollinger, line 250: I commend you for trying out new Http verbs where they may be better suited than just using plain GET/POST etc. 

However, this is not quite how HEAD works. A HEAD request is semantically the same as a GET without a body; returning only the HEADers that would be returned if the equivalent GET request was performed. So what would you expect a GET to `api/health-authorities/5/vendor/5` to return? I would be very surprised to get back a 404 response code when the resource _exists and is in use by a site_. That does not make sense in the context of a REST API. (Also, the URL should be `api/health-authorities/5/vendor**s**/5`)

Instead, how about `HEAD api/health-authorities/5/vendor/5/sites`? this makes more semantic sense, returning 200OK if there are HA sites using the code, or 404 if there are none using it? 404 is still kinda suspect since you would really expect an empty array rather than 404 if none exist but at least its more in line with expectation.

In fact, I think it would make even more sense if we just did `GET api/health-authorities/5/vendor/5/sites` and it returns a list of site Ids or very basic view models of the sites that are using the vendor, that way we could theoretically message to the user who is using the vendor.

prime-dotnet-webapi/Services/HealthAuthorityService.cs
```diff
@@ -148,17 +154,23 @@ public async Task UpdatePrivacyOfficeAsync(int healthAuthorityId, PrivacyOfficeV
...
-            });
+            _context.HealthAuthorityVendors.RemoveRange(deletedVendors);
+
+            var newVendors = vendorCodes
+                .Except(existingVendors.Select(ev => ev.VendorCode))
+                .Select(code => new HealthAuthorityVendor
+                {
+                    HealthAuthorityOrganizationId = healthAuthorityId,
+                    VendorCode = code
+                });
```
james-hollinger, line 173: I will show you how to use the `EntityMatcher` to greatly simplify this code

prime-dotnet-webapi/Services/HealthAuthorityService.cs
```diff
@@ -148,17 +154,23 @@ public async Task UpdatePrivacyOfficeAsync(int healthAuthorityId, PrivacyOfficeV
...
+            var deletedVendors = existingVendors
+                .Except(existingVendors
+                    .Where(ev => vendorCodes.Contains(ev.VendorCode)));
 
-            var newVendors = vendorCodes.Select(code => new HealthAuthorityVendor
-            {
-                HealthAuthorityOrganizationId = healthAuthorityId,
-                VendorCode = code
-            });
+            _context.HealthAuthorityVendors.RemoveRange(deletedVendors);
```
james-hollinger, line 165: You did not guard the backend to prevent deletion of vendor codes/care types that are in use

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
AMiKaH: I was in fact contemplating the last point of returning the sites that are using the vendor/care type. I only didn't go through with it since the call is barely to prevent a deletion while using that page. I thought it may be better to have a search feature by vendor/care type. But lets have a discussion about that.

prime-dotnet-webapi/Services/HealthAuthorityService.cs
AMiKaH: Sounds good, that code existed from before though.

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
james-hollinger: in the case of this feature, you only would be checking that any Sites/SiteId were returned and blocking the action if any exist.

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
```diff
@@ -240,5 +240,62 @@ public async Task<ActionResult> UpdatePharmanetAdministratorContacts(int healthA
...
+        [Authorize(Roles = Roles.EditSite)]
```
james-hollinger, line 251: Probably can be `Roles.ViewSite`

prime-dotnet-webapi/Services/HealthAuthorityService.cs
```diff
@@ -214,5 +226,33 @@ public async Task<bool> VendorExistsOnHealthAuthorityAsync(int healthAuthorityId
                 .AnyAsync(vendor => vendor.Id == healthAuthorityVendorId
                     && vendor.HealthAuthorityOrganizationId == healthAuthorityId);
         }
+
+        public async Task<bool> IsVendorInUseAsync(int healthAuthorityId, int healthAuthorityVendorCode)
```
james-hollinger, line 230: Names are important because often they can be misleading if not carefully chosen.
This method is called by hitting the API at `api/health-authorities/1/vendors/5`. That "5" isn't a `VendorCode`, its a _`HealthAuthorityVendorId`_. Some confusion may have been caused due to the equivalent PUT request to `api/health-authorities/1/vendors` taking a list of `VendorCodes` to create the `HealthAuthorityVEndor`s.

So if this parameter is renamed to `HealthAuthorityVendorId` its clear the method can be simplified:

```C#
return await _context.HealthAuthoritySites
    .AsNoTracking()
    .AnyAsync(site => site.HealthAuthorityOrganizationId == healthAuthorityId
        && site.HealthAuthorityVendorId == healthAuthorityVendorId);
```

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts
```diff
@@ -65,7 +65,11 @@ export class HealthAuthCareTypesPageComponent implements OnInit {
   }
 
   public removeCareType(index: number) {
-    this.careTypes.removeAt(index);
+    const careType = this.careTypes.value[index].careType;
```
hogmann, line 68: If we are doing a check to see in the front end if you can remove a vendor/caretype you will want to give the user feedback as to why their action didn't work.

We have a couple routes we could take.
One option you could just not let the user have the option to do the action so instead of checking in this call you could check once on load of the page for each care type and make the "X" disabled (maybe add a tooltip as to why when they hover over).
The other option is the one which path you are on, you just need to relay the message to the user that the vendor/caretype is in use either through a toast or other device.

prime-dotnet-webapi/Services/HealthAuthorityService.cs
AMiKaH: Added guard

prime-dotnet-webapi/Services/HealthAuthorityService.cs
AMiKaH: Refactored

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
AMiKaH: Changed, but would that still allow for a deletion of a care type/vendor?

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
AMiKaH: So you mean we still send a list of sites or site ids to front end and block based on whether or not it's empty?

prime-dotnet-webapi/Services/HealthAuthorityService.cs
AMiKaH: Hmm, on the front end, I only have access to the Vendor's code, careSettingCode and name. That's why I am sending the code and getting the ID on the backend. Attached is a screenshot of the info I have for a vendor on the vendor page. 
image

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts
AMiKaH: @hogmann on that note, should we disable the dropdown too so that an admin cannot make a change if a vendor/care-setting is in use?

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
AMiKaH: Now returning an array of `siteIds`, not currently used outside the logger.

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts
hogmann: Yeah understandable, the way the ticket was worded it made sense it led to the solution in the PR. Just wanted to put this comment as an alternative to that style of approach. Since the majority of times we want to guide the user to perform the actions they are allowed to rather than slap their wrist when they do an action they are not allowed to do.

hogmann: Feel free to close this conversation at any time

prime-dotnet-webapi/Services/HealthAuthorityService.cs
AMiKaH: Found a way to get the `vendorIds` on that page. Working on getting those ids

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts
```diff
@@ -103,7 +111,14 @@ export class VendorsPageComponent implements OnInit {
       .subscribe(({ vendors }: HealthAuthority) =>
         (vendors?.length)
           ? this.configService.vendors
-            .filter(v => vendors.some(({ vendorCode }) => v.code === vendorCode))
+            .reduce((vendorConfigs: VendorConfig[], v: VendorConfig) => {
```
AMiKaH, line 114: this block can be simplified by looping through `vendors` only, but will require changes to `VendorConfig` which in turn can have a regression impact. It may be easier to create a `HealthAuthorityVendorConfig`

prime-dotnet-webapi/Services/HealthAuthorityService.cs
AMiKaH: Refactored and now we have the ids of care-types and vendors on their respective pages.

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
```diff
@@ -240,5 +246,56 @@ public async Task<ActionResult> UpdatePharmanetAdministratorContacts(int healthA
...
+        /// </summary>
+        /// <param name="healthAuthorityId"></param>
+        /// <param name="careTypeId"></param>
+        [HttpGet("{healthAuthorityId}/care-types/{healthAuthorityCareTypeId}/sites", Name = nameof(GetSitesByCareType))]
+        [Authorize(Roles = Roles.ViewSite)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        public async Task<ActionResult> GetSitesByCareType(int healthAuthorityId, int careTypeId)
```
hogmann, line 289: you have a miss match here, since you use "healthAuthorityCareTypeId" above.

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
```diff
@@ -240,5 +246,56 @@ public async Task<ActionResult> UpdatePharmanetAdministratorContacts(int healthA
...
+        /// </summary>
+        /// <param name="healthAuthorityId"></param>
+        /// <param name="vendorId"></param>
+        [HttpGet("{healthAuthorityId}/vendors/{healthAuthorityVendorId}/sites", Name = nameof(GetSitesByVendor))]
+        [Authorize(Roles = Roles.ViewSite)]
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        public async Task<ActionResult> GetSitesByVendor(int healthAuthorityId, int vendorId)
```
hogmann, line 263: you have a miss match here, since you use "healthAuthorityVendorId" above.

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
```diff
@@ -240,5 +246,56 @@ public async Task<ActionResult> UpdatePharmanetAdministratorContacts(int healthA
...
+
+            var siteIds = await _healthAuthorityService.GetSitesByVendorAsync(healthAuthorityId, vendorId);
+
+            return Ok(siteIds);
+        }
+
+
+        // GET: api/health-authorities/5/care-types/5/sites
+        /// <summary>
+        /// Checks if the specified care type is in use by a Health Authority site
```
hogmann, line 278: Could probably simplify this to
``` returns a list of site ids of a given care type and health authority ```
Since as written it is a little miss leading.

File: prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs [hunk 11]
```diff
@@ -240,5 +246,56 @@ public async Task<ActionResult> UpdatePharmanetAdministratorContacts(int healthA
             await _healthAuthorityService.UpdateContactsAsync<HealthAuthorityPharmanetAdministrator>(healthAuthorityId, contacts);
             return NoContent();
         }
+
+        // GET: api/health-authorities/5/vendors/5/sites
+        /// <summary>
+        /// Checks if the specified vendor is in use by a Health Authority site
```
hogmann, line 252: Could probably simplify this to
```returns a list of site ids of a given vendor and health authority```
Since as written it is a little miss leading.

File: prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts [hunk 12]
```diff
@@ -58,14 +61,20 @@ export class HealthAuthCareTypesPageComponent implements OnInit {
...
   }
 
   public removeCareType(index: number) {
-    this.careTypes.removeAt(index);
+    const careTypeId = this.careTypes.value[index].careType.id;
+    this.healthAuthResource.getHealthAuthorityCareTypeSiteIds(this.route.snapshot.params.haid, careTypeId)
+      .subscribe((healthAuthoritySites) => {
+        (!healthAuthoritySites.length)
+          ? this.careTypes.removeAt(index)
+          : this.toastService.openErrorToast('Health authority care type could not be deleted. One or more sites are using it');
```
hogmann, line 76: To reduce the amount of text the user has to read in a toast could simplify to
``` Care type could not be removed, one or more sites are using it```

File: prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts [hunk 13]
```diff
@@ -72,7 +74,13 @@ export class VendorsPageComponent implements OnInit {
   }
 
   public removeVendor(index: number) {
-    this.vendors.removeAt(index);
+    const vendorId = this.vendors.value[index].vendor.id;
+    this.healthAuthResource.getHealthAuthorityVendorSiteIds(this.route.snapshot.params.haid, vendorId)
+      .subscribe((healthAuthoritySites) => {
+        (!healthAuthoritySites.length)
+          ? this.vendors.removeAt(index)
+          : this.toastService.openErrorToast('Health authority vendor could not be deleted. One or more sites are using it');
```
hogmann, line 82: same as above
``` Vendor could not be removed, one or more sites are using it```

File: prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts [hunk 14]
```diff
@@ -95,7 +104,11 @@ export class HealthAuthCareTypesPageComponent implements OnInit {
     this.healthAuthResource.getHealthAuthorityById(this.route.snapshot.params.haid)
       .subscribe(({ careTypes }: HealthAuthority) =>
         (careTypes?.length)
-          ? careTypes.map(ct => this.addCareType(ct.careType))
+          ? careTypes.map(ct => {
+            const id = ct.id;
+            const careTypeName = ct.careType;
+            this.addCareType({ id, careTypeName });
```
hogmann, line 110: Could reduce this to a line
``` this.addCareType({ ct.id, careTypeName: ct.careType })```

File: prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.html [hunk 15]
```diff
@@ -28,6 +28,7 @@
                   <app-options-form [form]="careType"
                                     fieldLabel="Care Type"
                                     controlName="careType"
+                                    optionLabel="careTypeName"
```
hogmann, line 31: On the initial load of the page you are not adapting the initial care types and they will not have this option label so nothing appears in the drop down.

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts
hogmann: Yeah should be making another model and then when you initialize the filteredVendors subject you convert the vendorConfigs into HealthAuthorityVendors and then only use HealthAuthorityVendors in this component.
Same goes with HealthAuthorityCareTypes

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
AMiKaH: I wrote `healthAuthorityCareTypeId` at first then it felt redundant since the Controller is that of a health authority. I can change it back.

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
AMiKaH: I wrote `healthAuthorityVendorId` at first then it felt redundant since the Controller is that of a health authority. I can change it back.

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
AMiKaH: Simplified

prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs
AMiKaH: Simplified

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts
AMiKaH: Simplified

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts
AMiKaH: Simplified

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts
AMiKaH: Reduced, you meant `this.addCareType({ id: ct.id, careTypeName: ct.careType })` right?

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts
AMiKaH: After fiddling with this, our `vendors` do not have a **vendor name**, therefore we need to make changes to the **VendorVM** and **Model** and add a migration, we'd also have to change the endpoints to take a list of **VendorVMs** rather than just **vendorCodes** in order to have the vendor name in the DB when we call the HA vendor update mehtod. Or we could keep the way that it is mapping the name to the rest of the object?

mtpultz: Like we were discussing a private member variable of `private vendorIdCodeMap: { id: number, code: number }` and then using it remove would de-complicate this quite a bit.

AMiKaH: Fixed

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts
hogmann: can't recall but i thought its intuitive and since if you just pass obj.id it infers it as a field with the same key "id"

AMiKaH: Sadly not when the id is a property of another object

prime-angular-frontend/src/app/core/resources/health-authority-resource.service.ts
```diff
@@ -107,6 +106,30 @@ export class HealthAuthorityResource {
       );
   }
 
+  public getHealthAuthorityVendorSiteIds(healthAuthorityId: number, vendorId: number): Observable<number[]> {
+    return this.apiResource.get<number[]>(`health-authorities/${healthAuthorityId}/vendors/${vendorId}/sites`)
+      .pipe(
+        map((response: ApiHttpResponse<number[]>) => response.result),
+        tap((healthAuthoritySiteIds: number[]) => this.logger.info('HEALTH_AUTHORITY_SITES', healthAuthoritySiteIds)),
+        catchError((error: any) => {
+          this.logger.error('[Core] HealthAuthorityResource::isHealthAuthorityVendorInUse error has occurred: ', error);
```
hogmann, line 115: wrong method in error "isHealthAuthorityVendorInUse"

prime-angular-frontend/src/app/core/resources/health-authority-resource.service.ts
```diff
@@ -107,6 +106,30 @@ export class HealthAuthorityResource {
...
+      );
+  }
+
+  public getHealthAuthorityCareTypeSiteIds(healthAuthorityId: number, careTypeId: number): Observable<number[]> {
+    return this.apiResource.get<number[]>(`health-authorities/${healthAuthorityId}/care-types/${careTypeId}/sites`)
+      .pipe(
+        map((response: ApiHttpResponse<number[]>) => response.result),
+        tap((healthAuthoritySiteIds: number[]) => this.logger.info('HEALTH_AUTHORITY_SITES', healthAuthoritySiteIds)),
+        catchError((error: any) => {
+          this.logger.error('[Core] HealthAuthorityResource::isHealthAuthorityCareTypeInUse error has occurred: ', error);
```
hogmann, line 127: Same as above

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts
```diff
@@ -43,7 +51,7 @@ export class HealthAuthCareTypesPageComponent implements OnInit {
       AdjudicationRoutes.HEALTH_AUTHORITIES,
       this.route.snapshot.params.haid
     ]);
-    this.filteredCareTypes = new BehaviorSubject<string[]>(this.configService.careTypes.map(ct => ct.name));
+    this.filteredCareTypes = new BehaviorSubject<HealthAuthorityCareTypeMap[]>(this.configService.careTypes.map(ct => { return { name: ct.name, code: ct.code } }));
```
hogmann, line 54: First off when wanting to return an object from an arrow function you can by just wrapping it in ( )
``` ct => ({ name: ct.name, code: ct.code }) ```

As well since CareTypeConfig is a subset of the fields in HealthAuthorityCareTypeMap
```this.configService.careTypes as HealthAuthorityCareTypeMap[]```

so we can just cast it, Also to make things more clear you could always define HealthAuthorityCareTypeMap like so
``` interface HealthAuthorityCareTypeMap extends CareTypeConfig ```

File: prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts [hunk 19]
```diff
@@ -52,20 +60,33 @@ export class HealthAuthCareTypesPageComponent implements OnInit {
...
+      this.healthAuthResource.getHealthAuthorityCareTypeSiteIds(this.route.snapshot.params.haid, careTypeId)
+        .subscribe((healthAuthoritySites) => {
+          (!healthAuthoritySites.length)
+            ? this.careTypes.removeAt(index)
+            : this.toastService.openErrorToast('Care type could not be removed, one or more sites are using it');
+        });
+    } else {
+      // when careTypeId is undefined that means we're deleting a care type after adding it and before hitting Save and Continue
+      // i.e. before it was given an Id
+      this.careTypes.removeAt(index);
```
hogmann, line 88: In cases like this when we have an if else and one side is a lot simpler of logic we can do something like this
```
public removeCareType(index: number) {
   const careTypeId = this.careTypes.value[index]?.careType?.id;
   if (!careTypeId) {
      this.careTypes.removeAt(index);
      return;
   }
   ... rest
}
```

This helps in the long run to keep indentation as clean as possible since the api call is no longer wrapped in an if/else.

Also i don't think this comment adds to the code.

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts
```diff
@@ -59,20 +68,32 @@ export class VendorsPageComponent implements OnInit {
...
+      this.healthAuthResource.getHealthAuthorityVendorSiteIds(this.route.snapshot.params.haid, vendorId)
+        .subscribe((healthAuthoritySites) => {
+          (!healthAuthoritySites.length)
+            ? this.vendors.removeAt(index)
+            : this.toastService.openErrorToast('Vendor could not be removed, one or more sites are using it');
+        });
+    } else {
+      // when vendorId is undefined that means we're deleting a vendor after adding it and before hitting Save and Continue
+      // i.e. before it was given an Id
+      this.vendors.removeAt(index);
```
hogmann, line 95: Same as mentioned in care types

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts
```diff
@@ -92,21 +113,23 @@ export class VendorsPageComponent implements OnInit {
...
 
     this.healthAuthResource.getHealthAuthorityById(this.route.snapshot.params.haid)
-      .subscribe(({ vendors }: HealthAuthority) =>
-        (vendors?.length)
-          ? this.configService.vendors
-            .filter(v => vendors.some(({ vendorCode }) => v.code === vendorCode))
-            .map(v => this.addVendor(v))
-          : this.addVendor()
-      );
+      .subscribe(({ vendors }: HealthAuthority) => {
```
hogmann, line 124: I don't think you need this opening brace.

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts
```diff
@@ -92,21 +113,23 @@ export class VendorsPageComponent implements OnInit {
...
+        vendors
```
hogmann, line 125: So like you have in care setting you will need the check on length, since this currently doesn't work when you are filling out a new HA

prime-angular-frontend/src/app/core/resources/health-authority-resource.service.ts
AMiKaH: changed to match new method name

prime-angular-frontend/src/app/core/resources/health-authority-resource.service.ts
AMiKaH: changed to match new method name

prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts
```diff
@@ -52,20 +60,33 @@ export class HealthAuthCareTypesPageComponent implements OnInit {
 
   public onSubmit() {
     if (this.formUtilsService.checkValidity(this.form)) {
-      const careTypes = [...new Set(this.careTypes.value.map(({ careType }) => careType.trim()) as string[])];
+      const careTypes = [...new Set(this.careTypes.getRawValue()
```
hogmann, line 63: Currently this doesn't work when you add a new custom care type, you will need some logic in here to handle the custom care types that are just a string
``` careType.name ? careType.name.trim() : careType.trim() ```

File: prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts (hunk 18 above)
Comment (AMiKaH) [reply prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts:54]: Done, I created the CareTypeConfig in the config.model.ts file for now since we don't have that type config.

File: prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts (hunk 23 above)
Comment (AMiKaH) [reply prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts:63]: Fixed

File: prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts (hunk 19 above)
Comment (AMiKaH) [reply prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/health-auth-care-types-page/health-auth-care-types-page.component.ts:88]: Fixed for both careTypes and vendors

File: prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts (hunk 20 above)
Comment (AMiKaH) [reply prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts:95]: fixed

File: prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts (hunk 21 above)
Comment (AMiKaH) [reply prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts:124]: Removed

File: prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts (hunk 22 above)
Comment (AMiKaH) [reply prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts:125]: Ah ok, that was the case where I was wondering if we'd ever have no vendors.

File: prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts (hunk 21 above)
Comment (AMiKaH) [reply prime-angular-frontend/src/app/modules/adjudication/pages/health-authorities/vendors-page/vendors-page.component.ts:124]: removed

File: prime-dotnet-webapi/Controllers/HealthAuthoritiesController.cs [hunk 24]
```diff
@@ -139,7 +139,10 @@ public async Task<ActionResult> UpdateCareTypes(int healthAuthorityId, IEnumerab
                 return NotFound($"Health Authority not found with id {healthAuthorityId}");
             }
 
-            await _healthAuthorityService.UpdateCareTypesAsync(healthAuthorityId, careTypes);
+            if (!await _healthAuthorityService.UpdateCareTypesAsync(healthAuthorityId, careTypes))
+            {
+                return BadRequest("Unable to update care types. One or more health authority care types may be in use");
```
james-hollinger, line 144: I think this ready better with just "are in use"

prime-dotnet-webapi/Services/HealthAuthorityService.cs
```diff
@@ -76,23 +77,37 @@ public async Task<bool> AuthorizedUserExistsOnHealthAuthorityAsync(int healthAut
                 .AnyAsync();
         }
 
-        public async Task UpdateCareTypesAsync(int healthAuthorityId, IEnumerable<string> careTypes)
+        public async Task<bool> UpdateCareTypesAsync(int healthAuthorityId, IEnumerable<string> careTypes)
```
james-hollinger, line 80: We should add a check either in this method or in the controller to prevent adding two identical care types, since it breaks the Entity Matcher

AMiKaH: The same check should be done on a vendor as well right?

AMiKaH: For vendors I created a method ```AnyVendorExistsOnHealthAuthorityAsync``` that takes in a health authority id and  an array of vendor codes and checks if they already exist within the health authority. Alternatively we could do a `foreach` and use the existing ```VendorExistsOnHealthAuthorityAsync``` that takes in one vendor code instead of a list.

AMiKaH: Done

james-hollinger (approved):
LGTM

johnnyl6 (approved):
🚀

hogmann (approved):
LGTM

PR #1851, opened 2021-12-06
PRIME-1983 Show signing authority name in event log

Paahn:
[Jira ticket ](https://bcgovmoh.atlassian.net/jira/software/c/projects/PRIME/boards/1?modal=detail&selectedIssue=PRIME-1983)
Steps to test:
1. Create and submit a site
2. Go to admin and select the site, approve/reject it
3. Go to event log and view party name tag for each business event


james-hollinger (changes requested):
I think there was a misunderstanding of what the JIRA ticket was asking for. We already have the Signing Authority's Name, it's on the `Party` related to the business event; we do not need to add a "PartyName" field to the Business Event model / table. The ticket was to surface the Signing Authority's name in the list of business events

prime-dotnet-webapi/Services/BusinessEventService.cs
```diff
@@ -225,6 +225,15 @@ private async Task<BusinessEvent> CreateSiteBusinessEvent(int BusinessEventTypeC
...
             int? adminId = admin?.Id;
+            var firstName = await _context.Parties
+                .Where(p => p.Id == partyId)
+                .Select(p => p.FirstName)
+                .SingleOrDefaultAsync();
+            var lastName = await _context.Parties
+                .Where(p => p.Id == partyId)
+                .Select(p => p.LastName)
+                .SingleOrDefaultAsync();
+            var partyName = String.Format("{0} {1}", firstName, lastName);
```
james-hollinger, line 236: You don't need to go to the DB twice here, you could just do 
```C#
var partyName = await _context.Parties
    .Where(p => p.Id == partyId)
    .Select(p => $"p.FirstName p.LastName")
    .SingleOrDefaultAsync();
```

prime-dotnet-webapi/Models/BusinessEvent.cs
```diff
@@ -50,5 +50,7 @@ public string AdminIDIR
         public string Description { get; set; }
 
         public DateTimeOffset? EventDate { get; set; }
+
+        public string PartyName { get; set; }
```
hogmann, line 54: On business event we already have the party as a relation, so this could either be a computed column on this model, or when we map the view model for these events we can map party name when we project.

We do not want to store this name on this event.

hogmann: This may be a great time to make view models for business events since we only want party name on non enrollee business events. As well we can get rid of all the [JsonIgnores].

hogmann (changes requested):

prime-dotnet-webapi/Models/BusinessEvent.cs
hogmann: I did not see @james-hollinger floating comment about this same topic, but yes I concur haha

prime-dotnet-webapi/Services/BusinessEventService.cs
Paahn: moved the PartyName building logic to the new ViewModel mapping

prime-dotnet-webapi/Models/BusinessEvent.cs
Paahn: SiteBusinessEventVM created. I didn't delete the JsonIgnores as I think they are still being used on the enrollee business events

prime-angular-frontend/src/app/modules/adjudication/shared/components/dated-content-table/dated-content-table.component.html
```diff
@@ -11,11 +11,15 @@
...
           </small>
-          <mat-chip-list *ngIf="item?.name"
-                         [selectable]="false">
-            <mat-chip selected
+          <mat-chip-list [selectable]="false">
+            <mat-chip *ngIf="item?.name"
+                      selected
                       color="primary"
                       [disableRipple]="true">{{ item.name | replace: '@idir': '' }}</mat-chip>
+            <mat-chip *ngIf="item?.partyName"
```
hogmann, line 19: So first off, when we add something like this to the template that references party name and don't reference it in the controller and the default DatedContent it kind of becomes a hidden feature. We probably want to avoid hidden features especially on generic components like this.

This leads into my second comment about i'm not sure if this is how we want to display this data. This is because of the following:

- It removes from the importance of the adjudicator being labelled on the event
- It will appear on every event and clutters the screen
- I don't think it is that important of information since on any site that hasn't been claimed it will all be the same

What we could do is add a footer to the Dated Content Class and then render the footer below the content and be a little more discreet with the font. Just smaller and maybe in a grey from the palette. That way the footer content is exposed for this component and can be used for other uses.

prime-dotnet-webapi/ViewModels/Business Events/SiteBusinessEventViewModel.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class SiteBusinessEventViewModel
+    {
+        public int Id { get; set; }
+
+        public string AdminIDIR
+        {
+            get => Admin?.IDIR;
+        }
+
+        public Admin Admin { get; set; }
```
james-hollinger, line 15: Do we need the `Admin` object here? I assume that we only need the `Admin`'s IDIR, meaning you could do a mapping just to get the IDIR in the same way you did the mapping to get the Signing Authority's name

prime-dotnet-webapi/ViewModels/Business Events/_BusinessEventMappingProfile.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Prime.ViewModels.Profiles
+{
+    public class BusinessEventMappingProfile : Profile
+    {
+        public BusinessEventMappingProfile()
+        {
+            CreateMap<BusinessEvent, SiteBusinessEventViewModel>()
+                .ForMember(dest => dest.PartyName, opt => opt.MapFrom(src => src.Party.FirstName + ' ' + src.Party.LastName))
+                .ForMember(dest => dest.AdminIDIR, opt => opt.MapFrom(src => src.AdminIDIR));
```
james-hollinger, line 12: `AdminIDIR` is not a `[Cumputed]` property on the `BusinessEvent` model; so while this mapping may (or may not!) succeed, the generated SQL will probably be rather weird looking. Likely, it will still fetch the entire `Admin` just to discard everything but the IDIR. you could map the property as `[Computed]` and ass `.DecompileAsync()`, or just do
```C#
CreateMap<BusinessEvent, SiteBusinessEventViewModel>()
    .ForMember(dest => dest.PartyName, opt => opt.MapFrom(src => src.Party.FirstName + ' ' + src.Party.LastName))
    .ForMember(dest => dest.AdminIDIR, opt => opt.MapFrom(src => src.Admin.IDIR));
```

james-hollinger: As an aside, AutoMapper would do this mapping for us automatically if we *didn't* have `AdminIDIR` on the `BusinessEvent` model. The first thing checks for is matching names, but AutoMapper will _also_ use naming conventions and will map from `Source.Child.Property` to `Destination.ChildProperty`

prime-angular-frontend/src/app/modules/adjudication/shared/components/dated-content-table/dated-content-table.component.html
```diff
@@ -18,7 +18,10 @@
                       [disableRipple]="true">{{ item.name | replace: '@idir': '' }}</mat-chip>
           </mat-chip-list>
         </th>
-        <td>{{ item.content }}</td>
+        <td class="flex">
```
hogmann, line 21: So we use bootstrap in the app, so we can simplify this class by using it and remove code from scss file
``` class="d-flex justify-content-between" ```

File: prime-angular-frontend/src/app/modules/adjudication/shared/components/dated-content-table/dated-content-table.component.scss [hunk 7]
```diff
@@ -1,3 +1,15 @@
...
   width: 1%;
 }
+
+.flex {
+  display: flex;
+  justify-content: space-between;
+}
+
+div.margin-right {
+  font-size: 13px;
```
hogmann, line 13: We probably want to stay away from defining font sizes in px and lean towards using relative font sizing.
``` font-size: smaller ```

Also with fonts when using an exact sizing you want to lean towards using em over px.

AMiKaH (approved):

james-hollinger (approved):
LGTM

hogmann (approved):

PR #1853, opened 2021-12-07
PRIME-1760 - Remote practitioner changes

neophyte57:
| Scenario                                                                                        | Expected                                                                                                  |
| ----------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------- |
| 1\. New site registration                                                                       | |
| 2\. Add new Remote User                                                                         | See confirmation alert; Toggle, which is set to on, is disabled |                                                                             
| 1\. With confirmation alert displaying, remove a Remote User                                    | Confirmation alert should disappear                                                                       |
| 1\. Remove existing Remote User                                                                 | |
| 2\. Add new Remote User                                                                         | See confirmation alert                                                                                    
| 1\. From Remote User page, Go Back (i.e. without "saving")                                      | Do not see confirmation alert                                                                             |
| 1\. With other existing Remote Users, add more Remote Users until there are at least 3 in total | After adding each Remote User, do see confirmation alert on returning to Remote Users screen              |
| 1\. Remove all Remote Users                                                                     | 1. Confirmation alert is not displayed as Remote Users are removed and toggle is enabled only once all Remote Users are removed |
| 2\. Add another Remote User                                                                     | 2. Confirmation alert is displayed after adding Remote User.  Toggle is disabled


prime-angular-frontend/src/app/modules/site-registration/pages/remote-users-page/remote-users-page.component.html
```diff
@@ -24,6 +24,21 @@
 
   </form>
 
+  <div class="mb-4"
+       *ngIf="addedFirstRemoteUser">
```
hogmann, line 28: So this banner you just want to show when atleast 1 remote user has been added. I don't think you need any of this query param passing, you can just observe the formState. Since in this component you are already rendering cards based on remote users.

So here just check
``` *ngIf="formState.remoteUsers.length > 0" ```

Or something along those lines.

File: prime-angular-frontend/src/app/modules/site-registration/pages/remote-user-page/remote-user-page.component.html [hunk 2]
```diff
@@ -71,15 +71,6 @@
                                             (remove)="removeCertification($event)"></app-college-certification-form>
 
           </ng-container>
-
-          <button mat-button
```
hogmann, line 75: So this leads to a discussion about when we are asked to remove features (in this case only one certification for a remote user) should we remove the supporting code and refactor.

Since the next person coming into the component may not realize the form state and controller have a lot of code that no longer reflects the state of the app to the user. 

I think in this sprint we have time to spend some of it to refactor this so that the page only supports one certification.

prime-angular-frontend/src/app/modules/site-registration/pages/remote-users-page/remote-users-page.component.html
```diff
@@ -24,6 +24,21 @@
...
+    <app-alert type="success"
+               icon="check_circle_outline">
+      <ng-container #alertTitle
+                    class="alert-title">
+        Remote Practitioner added
```
AMiKaH, line 33: I can add a user and remove then untoggle and the banner keeps saying that a remote practitioner was added
Is this expected?
image

prime-angular-frontend/src/app/modules/site-registration/pages/remote-users-page/remote-users-page.component.html
neophyte57: Asking Pilar in the Jira ticket re: expected behavior

neophyte57: Maybe I'm misunderstanding you @hogmann but we can't display the banner if the user hasn't actually just added a Remote Practitioner.  I think for what you're suggesting, the banner would display as long as there are any Remote Practitioners associated with the site, correct?

hogmann: So my understanding from 
``` once a remote practitioner is added show green banner “remote practitioner added, continue adding remote practitioners if this site has more than one”. ```

Is we show this banner once we have added a remote user. So we already have the code that only adds the query param "fromRemoteUser" when you have successfully submitted the form on the previous page. So with that param and the amount of remote users in the list we know if we have added/updated a remote user. So we can show the banner in that case.

Then since there is that one corner case that if the user updates a user and submits the banner will show to. So in that case it might not be unreasonable just to say on the banner remote user added/updated

Comment (hogmann) [reply prime-angular-frontend/src/app/modules/site-registration/pages/remote-users-page/remote-users-page.component.html:28]: Then to handle Amir's bug with if you remove them from that page the banner is still there you can just clear whatever flag you set when you see the "fromRemoteUser" on removal of a user.

Comment (neophyte57) [reply prime-angular-frontend/src/app/modules/site-registration/pages/remote-users-page/remote-users-page.component.html]: Should be addressed now

**Review by johnnyl6 — APPROVED** [review]
LGTM

File: prime-angular-frontend/src/app/modules/site-registration/pages/remote-user-page/remote-user-page.component.html (hunk 2 above)
Comment (hogmann) [reply prime-angular-frontend/src/app/modules/site-registration/pages/remote-user-page/remote-user-page.component.html:75]: Is this being addressed?

Comment (neophyte57) [reply prime-angular-frontend/src/app/modules/site-registration/pages/remote-user-page/remote-user-page.component.html:75]: Apologies I didn't join Zoom chat to talk about this.  I tried addressing but haven't yet completed.  Instead I found a couple of other defects that exist in dev.pharmanetenrolment.gov.bc.ca; see PRIME-2050

**Comment by neophyte57** [comment]
> When I have more than one Remote practitioner and I remove at least one, the `This site has remote practitioners ` toggle becomes disabled, and becomes active again by removing all Remote practitioners. Should we disable the toggle after adding a Remote practitioner, rather than when we remove from the list of existing practitioners? toggle disabled

Good point, should be addressed now

**Review by AMiKaH — APPROVED** [review]

**Review by Paahn — APPROVED** [review]
LGTM

File: prime-dotnet-webapi/ViewModels/SiteRegistration/SiteRemoteUserUpdateModel.cs [hunk 4]
```diff
@@ -0,0 +1,15 @@
...
+namespace Prime.ViewModels
+{
+    public class SiteRemoteUserUpdateModel
+    {
+        public int Id { get; set; }
+        public int SiteId { get; set; }
+        public string FirstName { get; set; }
+        public string LastName { get; set; }
+        public string Email { get; set; }
+        public bool Notified { get; set; }
```
james-hollinger, line 12: I don't think `Notified` should be on this model

james-hollinger (approved):
LGTM

PR #1862, opened 2021-12-07
Feature/prime 1925 enrolment status history

hogmann:
Steps to Replicate

1. Create an enrolment and answer Yes to one or more of the self declaration questions, and submit
2. As an adjudicator mark the enrolment as editable and optionally add a note for the reason why they are being put back to Editable
3. As the enrollee edit the enrolment and answer No to all questions, and submit
4. Then go to the "/adjudication/enrollees/{enrolleeId}/review" and view the list of status.

Was discussed that it will just show the status reason since parsing the self declaration answers from the submissions table is way more work than is needed for a tiny fraction of use cases. This way you can still see the reason for past self declaration questions answered.


AMiKaH (approved):

james-hollinger (approved):
LGTM

PR #1863, opened 2021-12-08
Remove noisy code climate comments regarding similar code

AMiKaH:
disabled checks for similar code, keeping the checks for identical code


hogmann (approved):
lgtm

Paahn (approved):

PR #1866, opened 2021-12-13
PRIME-1926 Display previously uploaded self-declaration documents

neophyte57:
| Scenario                                                                            | Expected                                                                        |
| ----------------------------------------------------------------------------------- | ------------------------------------------------------------------------------- |
| 1\. Answer "Yes" to one or more Self-Declaration questions                          | |
| 2\. Upload one or more documents for the Self-Declaration question                  |
| 3\. Save and Continue                                                               |
| 4\. On Overview page, click pencil icon to return to Self-Declaration page          | Multiple documents are grouped underneath the correct Self-Declaration question 
| 1. On Self-Declaration page, click attachment to download previously uploaded document | Correct document is downloaded                                                  |
| | Navigation does NOT leave page


prime-angular-frontend/src/app/modules/enrolment/pages/self-declaration/self-declaration.component.html
```diff
@@ -86,6 +86,12 @@
                                  (remove)="onRemove('hasConvictionDocumentGuids', $event)">
             </app-document-upload>
 
+            <app-enrollee-property>
+              <app-document-attachment [documents]=getSelfDeclarationDocuments(SelfDeclarationTypeEnum.HAS_CONVICTION)
+                                       (download)="downloadSelfDeclarationDocument($event)">
+              </app-document-attachment>
+            </app-enrollee-property>
```
AMiKaH, line 93: Part of the comment already exists in our current code but doesn't seem to be an issue.

If I answer yes and upload a document, the document will show in the **overview** page in the right place, and if I go back to Self-Declaration page, the documents show up under the right section as well, however; if I change the answer to that question to **No**, the document and description in the Self-Declaration page will disappear properly, but the document still shows up in the overview page even with an answer of **No** and no description. Also, if I go back to the Self-Declaration page and change the answer back to yes, the documents show up again.

I am not sure if we should persist the uploaded documents if the answer is changed to **No**.

Otherwise all is good.

AMiKaH (approved):

prime-angular-frontend/src/app/modules/enrolment/pages/self-declaration/self-declaration.component.ts
```diff
@@ -100,9 +102,15 @@ export class SelfDeclarationComponent extends BaseEnrolmentProfilePage implement
...
-    super.onSubmit(hasBeenThroughTheWizard);
+    if (!this.clickedDownloadDoc) {
+      const hasBeenThroughTheWizard = true;
+      this.hasAttemptedFormSubmission = true;
+      super.onSubmit(hasBeenThroughTheWizard);
+    } else {
+      // If downloading previously uploaded doc, remain on page.
+      // Reset flag.
+      this.clickedDownloadDoc = false;
+    }
```
johnnyl6, line 113: when the enrolment is editable and enrollee tries to edit the self-declaration on overview page by the pencil icon, uploading a different file would not change the file button name.
e.g. during initial enrolment, upload a `file1.png` for one of the self-declaration.
after making the enrolment editable, upload a `file2.png' to replace the existing `file1.png'. The download button remains `file1.png' in overview page.

prime-angular-frontend/src/app/modules/enrolment/pages/self-declaration/self-declaration.component.ts
```diff
@@ -100,9 +102,15 @@ export class SelfDeclarationComponent extends BaseEnrolmentProfilePage implement
   }
 
   public onSubmit() {
-    const hasBeenThroughTheWizard = true;
-    this.hasAttemptedFormSubmission = true;
-    super.onSubmit(hasBeenThroughTheWizard);
+    if (!this.clickedDownloadDoc) {
```
hogmann, line 105: What is this code doing? 
It looks to be that I am required to hit the submit button twice if i decided to download a document?
Is this attempting to stop the user from submitting when an action is pending?

hogmann: So played around testing it and I see why this is there. So in this case we need to update the app-document-attachment and make sure that button is updated so it isn't assumed to be a submit button. This has probably never came us since this component isn't usually used in a form.

"Just add type='button' to it. A button without any defined type in a form acts like a submit button for the form."

prime-angular-frontend/src/app/modules/enrolment/pages/self-declaration/self-declaration.component.ts
neophyte57: @johnnyl6  Not sure I understand what you expect:  In `dev` and this PR, uploading another self-declaration document **adds** the document to the ones already associated with that self-declaration question.

prime-angular-frontend/src/app/modules/enrolment/pages/self-declaration/self-declaration.component.html
neophyte57: Will try to address in this PR, i.e. we should **not** persist the uploaded documents if the answer is changed to No.

prime-angular-frontend/src/app/modules/enrolment/pages/self-declaration/self-declaration.component.ts
johnnyl6: Actually might be 2 issues:
1. Changed attachment to `File2.png' but the download button still says the old one `Capture.PNG`
image

2. Changed selection to `No` but download button is still visible.
image

prime-angular-frontend/src/app/modules/enrolment/pages/self-declaration/self-declaration.component.ts
hogmann: Perfect!

hogmann (approved):
LGTM 🎉

prime-angular-frontend/src/app/modules/enrolment/pages/self-declaration/self-declaration.component.html
hogmann: Just a reminder too, that if something is outside the scope of the ticket it's ok to take these thoughts and put them in another ticket. Since I think that's a little bigger task that what you have laid out here and we are already sudo rolling this ticket into another sprint. (this ticket is a 0.5)

I say this because how documents are handled/finalized is an area that needs attention, but it is a task that needs some planning and is outside the scope of this ticket.

I think this and @johnnyl6 comment below are about the same topic so this refers to both*

neophyte57: Alright, created PRIME-2054 to capture this

PR #1879, opened 2021-12-16
Fix add email template types to FE enum

hogmann:
Check maintenance email templates page will load with no errors

Edit: Updated to remove template from the front end and just pass down the template name so we don't have to maintain two enums


AMiKaH (approved):

Paahn (approved):
✌️

Paahn (approved):
🧙‍♀️ 🧙

james-hollinger (approved):
L (?:^|\W)good(?:$|\W) TM

AMiKaH (approved):

PR #1883, opened 2021-12-20
PRIME-2053 do not send renewal emails to absent enrollees

AMiKaH:
[PRIME-2053](https://bcgovmoh.atlassian.net/jira/software/c/projects/PRIME/boards/1?modal=detail&selectedIssue=PRIME-2053)

This is to be merged after [PR 1884](https://github.com/bcgov/moh-prime/pull/1884)

Pre test:

1. Add a new migration.
3. In `management/enrollees/renewal` endpoint add [AllowAnnonymous] and comment out out the authorize role (found on line 45 of the EmailsController.cs)

How to test:

1. Create 2 enrollees and approve them
2. Add an absence to one of the enrollees
3. Hit this endpoint from postman http://localhost:5000/api/Emails/management/enrollees/renewal
4. Check your mailhog (only one enrollee should get an email)


Paahn (approved):
LGTM🤩

prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -1107,5 +1107,15 @@ public async Task UpdateDateOfBirthAsync(int enrolleeId, DateTime dateOfBirth)
             enrollee.DateOfBirth = dateOfBirth;
             await _context.SaveChangesAsync();
         }
+
+        public async Task<bool> IsEnrolleeAbsentAsync(int enrolleeId)
+        {
+            var now = DateTime.Now;
```
johnnyl6, line 1113: should be `DateTime.UtcNow` if comparing with the `StartTimestamp` and `EndTimestamp` which are stored as UTC time without timezone in database.

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -233,7 +237,13 @@ public async Task SendEnrolleeRenewalEmails()
 
             foreach (var enrollee in enrollees)
             {
+                if (await _enrolleeService.IsEnrolleeAbsentAsync(enrollee.Id))
```
johnnyl6, line 240: is it possible to update the query  in line `225` using a join with the EnrolleeAbsense table to get the enrollees need renewal rather than calling `IsEnrolleeAbsentAsync` in the loop to skip absent enrollees. This could save a lot unnecessary database query in the loop.

james-hollinger (changes requested):
As written, the code is going to hit the database once for each Enrollee in the database. So potentially 100,000+ times. That is not good.

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -27,19 +27,22 @@ private static class SendType
...
             IEmailDocumentsService emailDocumentService,
             IEmailRenderingService emailRenderingService,
+            IEnrolleeService enrolleeService,
             ISmtpEmailClient smtpEmailClient)
             : base(context, logger)
         {
             _chesClient = chesClient;
             _emailDocumentService = emailDocumentService;
             _emailRenderingService = emailRenderingService;
+            _enrolleeService = enrolleeService;
```
johnnyl6, line 45: IMO two options here:
1. `EnrolleeService` is no longer needed and removed. Minimal changes needed.
2. Keep the injected `EnrolleeService` and move the entire`getRenewalEnrollees` business logic into `EnrolleeService`. `getRenewalEnrollees` returns a list of `EnrolleeRenewalEmailViewModel` with an added Email field.
Advantage of `#2` is separation of different business.

AMiKaH: I was actually using `GetNonAbsentEnrollees` in `EnrolleeService` which returns a list of enrollee ids of those who are not absent (similar to the query in the renewal method) then I was calling the `!Contains` on that list, I forgot to remove the `EnrolleeService` injection afterwards. The reason I removed it from EnrolleeService was because we don't have a use for that list outside the renewal emails, I'm open to using EnrolleeService again, or we can save that for the future in case we ever need to use that list somewhere else.  For now I just removed the service.

prime-dotnet-webapi/Models/EnrolleeAbsence.cs
```diff
@@ -14,7 +14,6 @@ public class EnrolleeAbsence : BaseAuditable, IEnrolleeNavigationProperty
 
         public Enrollee Enrollee { get; set; }
         public DateTime StartTimestamp { get; set; }
-
-        public DateTime EndTimestamp { get; set; }
+        public DateTime? EndTimestamp { get; set; }
```
johnnyl6, line 17: Might need to generate a db migration since current `EnrolleeAbsence` table column `EndTimestamp` is still `not null`

johnnyl6 (approved):
except the possible db migration otherwise LGTM

prime-dotnet-webapi/Models/EnrolleeAbsence.cs
AMiKaH: That migration will be done in https://github.com/bcgov/moh-prime/pull/1884. I will wait for Nathan's PR to be merged before merging mine.

Paahn (approved):
lgtm

james-hollinger (approved):
LGTM

PR #1884, opened 2021-12-20
PRIME-1991 Enrollee Absence Indefinite + email

hogmann:
To test:

1. Get an enrollee past the point they have accepted a TOA
2. Log in on that enrollee and click the absence management tab on the dashboard
3. Only select one date for the absence
4. Publish
5. Type in an email and send.
6. Rinse and repeat with current/future/past .... etc


prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -622,5 +623,46 @@ public async Task<ActionResult> DeleteFutureEnrolleeAbsence(int enrolleeId, int
...
+            }
+
+            var absences = await _enrolleeService.GetEnrolleeAbsencesAsync(enrolleeId, false);
+            if (!absences.Any())
+            {
+                return NotFound($"No open absences found for enrollee with id: {enrolleeId}");
+            }
+
+            await _emailService.SendEnrolleeAbsenceNotificationEmailAsync(enrolleeId, absences.First(), email);
+            await _businessEventService.CreateEmailEventAsync(enrolleeId, $"Absence notification email sent to: {email}");
```
Paahn, line 663: It seems the email is not a string here, as the message in admin is currently `showing Absence notification email sent to: Prime.Models.Api.FromBodyText`

prime-angular-frontend/src/app/modules/enrolment/pages/absence-management-page/absence-management-page.component.ts
```diff
@@ -43,17 +43,33 @@ export class AbsenceManagementPageComponent extends AbstractEnrolmentPage implem
...
-      .subscribe(() => this.absence = null);
+      .subscribe(() => {
+        this.absence = null;
+        this.formState.email.removeValidators(Validators.required);
+      });
+  }
+
+  public sendEmail(): void {
+    this.formState.email.markAsTouched();
+    if (!this.formState.email.valid) {
```
AMiKaH, line 67: For readability we can use `if(this.formState.email.invalid)`

AMiKaH (approved):

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -342,6 +343,23 @@ public async Task SendPaperEnrolmentSubmissionEmailAsync(int enrolleeId)
...
+            var enrolleeDto = await _context.Enrollees
+                .Where(e => e.Id == enrolleeId)
+                .Select(e => new
+                {
+                    e.FirstName,
+                    e.LastName
+                })
+                .SingleOrDefaultAsync();
+
+            var viewModel = new EnrolleeAbsenceNotificationEmailViewModel(enrolleeDto.FirstName, enrolleeDto.LastName, absence.StartTimestamp, absence.EndTimestamp);
```
johnnyl6, line 357: It seems the intermediate `enrolleeDto` is not necessary as the `viewModel` could be constructed directly in the LINQ query. 

Further, in theory `enrolleeDto` could be null by calling `SingleOrDefaultAsync` hence causing null exception in `enrolleeDto.FirstName` and `enrolleeDto.LastName` later.
```C#
var viewModel = await _context.Enrollees
    .Where(e => e.Id == enrolleeId)
    .Select(e => new EnrolleeAbsenceNotificationEmailViewModel(
        e.FirstName, e.LastName, absence.StartTimestamp, absence.EndTimestamp)
    )
    .SingleOrDefaultAsync();
```

prime-angular-frontend/src/app/modules/enrolment/pages/absence-management-page/absence-management-page.component.ts
hogmann: FormControl's can be in 4 states
https://angular.io/api/forms/FormControlStatus

So when checking if something is !valid it is actually different than checking if it is invalid. This is why we want to fail out if the form control is !valid

johnnyl6 (approved):
LGTM

prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -622,5 +623,46 @@ public async Task<ActionResult> DeleteFutureEnrolleeAbsence(int enrolleeId, int
...
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        public async Task<ActionResult> SendEnrolleeAbsenceEmail(int enrolleeId, FromBodyText email)
+        {
+            if (!Email.IsValidEmail(email))
+            {
+                return BadRequest("The email provided is not valid.");
+            }
+
+            var enrollee = await _enrolleeService.GetEnrolleeAsync(enrolleeId);
```
james-hollinger, line 646: rather than getting the whole enrollee, you can use `EnrolleeService.GetPermissionsRecordAsync()`

prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -622,5 +623,46 @@ public async Task<ActionResult> DeleteFutureEnrolleeAbsence(int enrolleeId, int
...
+            }
+
+            var absences = await _enrolleeService.GetEnrolleeAbsencesAsync(enrolleeId, false);
+            if (!absences.Any())
+            {
+                return NotFound($"No open absences found for enrollee with id: {enrolleeId}");
+            }
+
+            await _emailService.SendEnrolleeAbsenceNotificationEmailAsync(enrolleeId, absences.First(), email);
+            await _businessEventService.CreateEmailEventAsync(enrolleeId, $"Absence notification email sent to: {email.Data}");
```
james-hollinger, line 663: `FromBodyText` can be used as if it was a string since it has an `Implicit` conversion to string. You do not need to do `email.Data` here

Paahn (approved):
LGTM

prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -1046,7 +1046,7 @@ public async Task<IEnumerable<EnrolleeAbsenceViewModel>> GetEnrolleeAbsencesAsyn
             return await _context.EnrolleeAbsences
                 .Where(ea => ea.EnrolleeId == enrolleeId)
                 .If(!includesPast,
-                    absences => absences.Where(ea => rightNow <= ea.EndTimestamp)
+                    absences => absences.Where(ea => rightNow <= ea.EndTimestamp || !ea.EndTimestamp.HasValue)
```
james-hollinger, line 1049: `!...HasValue` can just be `== null`

prime-dotnet-webapi/ViewModels/Enrollee/_EnrolleeMappingProfile.cs
```diff
@@ -37,7 +38,10 @@ public EnrolleeMappingProfile()
             CreateMap<EnrolleeDTO, EnrolleeViewModel>();
 
             CreateMap<EnrolleeNote, EnrolleeNoteViewModel>();
-            CreateMap<EnrolleeAbsence, EnrolleeAbsenceViewModel>();
+
+            CreateMap<EnrolleeAbsence, EnrolleeAbsenceViewModel>()
+                .ForMember(dest => dest.StartTimestamp, opt => opt.MapFrom(src => src.StartTimestamp.ToLocalTime()))
+                .ForMember(dest => dest.EndTimestamp, opt => opt.MapFrom(src => src.EndTimestamp.HasValue ? (DateTime?)src.EndTimestamp.Value.ToLocalTime() : null));
```
james-hollinger, line 44: I think the slightly better way to do this is using a precondition, like we do for agreements:
image

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -344,6 +345,17 @@ public async Task SendPaperEnrolmentSubmissionEmailAsync(int enrolleeId)
             await Send(email);
         }
 
+        public async Task SendEnrolleeAbsenceNotificationEmailAsync(int enrolleeId, EnrolleeAbsenceViewModel absence, string email)
+        {
+            var viewModel = await _context.Enrollees
+                .Where(e => e.Id == enrolleeId)
+                .Select(e => new EnrolleeAbsenceNotificationEmailViewModel(e.FirstName, e.LastName, absence.StartTimestamp, absence.EndTimestamp))
```
james-hollinger, line 352: Could you test what SQL gets generated by this? This is the first time in the app we are using the _constructor_ of a view model rather than assigning the properties using object initialization syntax a la
```c#
.Select(e => new EnrolleeAbsenceNotificationEmailViewModel
{
  e.FirstName, 
  e.LastName, 
  absence.StartTimestamp, 
  absence.EndTimestamp  
})
```
EF might be smart enough here but what may happen is that the entire `Enrollee` model is retrieved from the DB and then just `FirstName` and `LastName` get mapped into the VM object.

hogmann: It appears to be smart enough to not grab it all 👍 
```
SELECT e."FirstName", e."LastName"
FROM "Enrollee" AS e
WHERE e."Id" = @__enrolleeId_0
LIMIT 2
```

james-hollinger (approved):
LGTM

PR #1894, opened 2021-12-21
PRIME-2041 Provision SAT Enrolments in the PRIME Keycloak with a role

AMiKaH:
[PRIMET-2041](https://bcgovmoh.atlassian.net/jira/software/c/projects/PRIME/boards/1?modal=detail&selectedIssue=PRIME-2041)

### DIFF
--- a/documentation/Keycloak settings/realm-export-dev.json
+++ b/documentation/Keycloak settings/realm-export-dev.json
@@ -198,6 +198,14 @@
         "attributes": {}
       },
       {
+        "id": "266c24f3-beab-43fe-8e33-3111b50093f8",
+        "name": "phsa_eforms_sat",
+        "composite": false,
+        "clientRole": false,
+        "containerId": "v4mbqqas",
+        "attributes": {}
+      },
+      {
         "id": "301349dc-e58b-4666-90c0-f6d9782af655",
         "name": "prime_user",
         "description": "BCSC authenticated primary application user",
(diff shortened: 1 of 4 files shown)


prime-dotnet-webapi/Controllers/SatEnrolmentController.cs
```diff
@@ -195,6 +199,7 @@ public async Task<ActionResult> SubmitSatEnrollee(int satId)
 
             var existsInPlr = await _plrProviderService.PartyExistsInPlrWithCollegeIdAndNameAsync(satId);
             var submission = await _partyService.CreateSubmissionAsync(satId, SubmissionType.SatEnrollee, existsInPlr);
+            await _keycloakClient.AssignRealmRole(User.GetPrimeUserId(), Roles.PhsaEformsSat);
```
AMiKaH, line 202: We could also make a wrapper for this method in `PhsaService` if we don't want to call the KeyCloak service in the controller

james-hollinger, line 202: I would not use `User.GetPrimeUserId()` here.  Using `User` ties this to request state unnecessarily (making it harder to test) and could theoretically introduce a bug if we ever made this endpoint callable by admins (in that case, this would give the role to the _admin_, rather than the user).  In general, you should write your code to be defensive against possible bugs when it doesn't sacrifice anything.

I would use `satEnrollee.UserId`

johnnyl6 (approved):
LGTM

james-hollinger (approved):
LGTM

Paahn (approved):
lgtm

PR #1895, opened 2021-12-21
PRIME-1927 Update Email Status doesnt priority non-statuses

johnnyl6:
- Added `UpdateCount` column to track number of update from CHES
- Added `OrderBy UpdateCount` in query to lower priority of email with more `UpdateCount`


prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -310,18 +310,24 @@ public async Task<int> UpdateEmailLogStatuses(int limit)
 
             var emailLogs = await _context.EmailLogs
                 .Where(predicate)
+                .Where(e => e.LatestStatus != ChesStatus.Pending)
```
james-hollinger, line 313: so this will never update email logs that are pending?

james-hollinger (changes requested):
There was a discussion about this ticket when it was first intook a sprint or two ago and it looks like those details weren’t passed on and they weren’t outlined in the ticket appropriately. We should have a quick talk about it since at the time we assumed someone else was going to make these updates with the knowledge of those discussions.

prime-dotnet-webapi/Models/Email/EmailLog.cs
```diff
@@ -29,6 +29,8 @@ public class EmailLog : BaseAuditable
 
         public string StatusMessage { get; set; }
 
+        public int? UpdateCount { get; set; }
```
james-hollinger, line 32: `UpdateCount` should be an `int`, which will initialize at 0

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -310,6 +310,7 @@ public async Task<int> UpdateEmailLogStatuses(int limit)
 
             var emailLogs = await _context.EmailLogs
                 .Where(predicate)
+                .OrderBy(e => e.UpdateCount)
                 .OrderBy(e => e.UpdatedTimeStamp)
```
james-hollinger, line 314: you need to use `.ThenBy(e => e.UpdatedTimeStamp)` here

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -321,6 +322,8 @@ public async Task<int> UpdateEmailLogStatuses(int limit)
                 {
                     email.LatestStatus = status;
                 }
+                email.UpdateCount = email.UpdateCount.HasValue ? email.UpdateCount.Value + 1 : 1;
```
james-hollinger, line 325: with `UpdateCount` being an `int`, this can just be `email.UpdateCount++`

AMiKaH (approved):

james-hollinger (approved):
LGTM

PR #1903, opened 2021-12-28
PRIME-2060 Fix Code Climate

AMiKaH:
Code climate was not being triggered when PRs were merged to develop and therefore coverage percentage wasn't being displayed in the code climate dashboard.

Still investigating the `diff-coverage`. But in order to make sure current changes are working, we need to merge this branch. It is possible that the `diff-coverage` will also be fixed once the PR is merged based on Code Climate's docs

Tested the work by creating a branch off of develop and another branch off of that one to mock our PR process.


prime-angular-frontend/src/app/modules/Test.ts
```diff
@@ -0,0 +1,58 @@
+export class TestingStuff {
```
hogmann, line 1: What is all this?

AMiKaH: This will be deleted as in the commit message. I just wanted to add useless code without adding any unit test to see why the diff coverage is always 100%

.github/workflows/test-pr.yml
```diff
@@ -1,17 +1,17 @@
 name: CodeClimate-Checks
 on:
   push:
-    branches-ignore: 
+    branches:
       - 'develop'
-      - 'test'
-      - 'master'
+      - '!test'
```
hogmann, line 6: this syntax seems off

hogmann: Since what is the point of listing develop here if you have to implicitly list the !branches. If this by default includes all branches just stick with doing ignore.

AMiKaH: We need to use `!` to exclude branches if we are using `branches` (which we are, to target develop) we can't use `branches-ignore` under the same event.

AMiKaH: If we only do branches ignore and add `master` and `test`, then the runs will be triggered twice for each push to a PR, one for `push` and one for `pull_request`

hogmann: I fail to see how functionally these two are different.
```
branches:
   - 'develop'
   - '!test'
   - '!master'
```
```
branches-ignore:
   - 'test'
   - 'master'
```

hogmann: Since if you are saying you need to target  develop wouldn't just this suffice
```
branches:
   - 'develop'
```

AMiKaH: To reply to the first comment:

```
branches:
   - 'develop'
   - '!test'
   - '!master'
```
   
```
branches-ignore:
   - 'test'
   - 'master'
```

the second block will trigger the event twice every time the PR is updated.

```
branches:
   - 'develop'
```
   
With this one, when we merge develop to test and master, the checks will run.

johnnyl6 (approved):
🚀

hogmann (approved):
LGTM

PR #1910, opened 2021-12-31
PRIME-2054 Do not persist SD documents when answer changes from Yes to No

AMiKaH:
[PRIME-2054](https://bcgovmoh.atlassian.net/jira/software/c/projects/PRIME/boards/1?modal=detail&selectedIssue=PRIME-2054)

### How to test:

1. Create an enrolment and go through the steps until you reach the Self Declaration page
2. Answer Yes to any of the questions, add description and upload a document.
3. Go to overview (You should see the document under the Self Declaration section under the question that has a **Yes** answer)
4. Click the edit button on the Self Declaration section in overview (You should be taken back to Self Declaration page and see the uploaded document there)
5. Change the answer from **Yes** to **No**
6. Go to overview (The document should be displayed anymore)
7. Go back to Self Declaration (The document should be displayed)
8. Change the answers to Yes to all questions (No document should be displayed)
9. Login as admin, and view the enrolment (You should see the document under the Self **Declaration** section in the **Enrolment** card)


prime-dotnet-webapi/Models/Documents/SelfDeclarationDocument.cs
```diff
@@ -15,5 +15,6 @@ public class SelfDeclarationDocument : BaseDocumentUpload, IEnrolleeNavigationPr
 
         [JsonIgnore]
         public SelfDeclarationType SelfDeclarationType { get; set; }
+        public bool IsHidden { get; set; }
```
james-hollinger, line 18: "Hidden" already sounds like a boolean, don't really need to call it `IsHidden`.
Also, please observe the spacing of the file.

AMiKaH:
Just found a bug in adjudication. Fixing it.

prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -495,6 +497,19 @@ private async Task CreateSelfDeclarationDocuments(int enrolleeId, ICollection<Se
                     });
                 }
             }
+
+            foreach (var document in currentSelfDeclarationDocuments)
+            {
+                if (!newDeclarations.Select(nd => nd.SelfDeclarationTypeCode).Contains(document.SelfDeclarationTypeCode))
```
james-hollinger, line 503: rather than doing `.Select.Contains` you can just do ` if (!newDeclarations.Any(newDocument => newDocument.SelfDeclarationTypeCode == oldDocument.SelfDeclarationTypeCode))`

prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -495,6 +497,19 @@ private async Task CreateSelfDeclarationDocuments(int enrolleeId, ICollection<Se
...
+                {
+                    var documentToUpdate = await _context.SelfDeclarationDocuments
+                        .Where(sdd => sdd.EnrolleeId == enrolleeId && sdd.DocumentGuid == document.DocumentGuid)
+                        .SingleOrDefaultAsync();
```
james-hollinger, line 507: you don't need this call, right? you already have the document

prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -495,6 +497,19 @@ private async Task CreateSelfDeclarationDocuments(int enrolleeId, ICollection<Se
...
+
+                    documentToUpdate.Hidden = true;
+                    await _context.SaveChangesAsync();
```
james-hollinger, line 510: You don't need to save on every document; in fact you should probably not be saving here at all in the middle of updating the Enrollee.

prime-dotnet-webapi/ViewModels/Enrollee/SelfDeclarationViewModel.cs
```diff
@@ -11,5 +11,6 @@ public class SelfDeclarationViewModel
         public int SelfDeclarationTypeCode { get; set; }
         public string SelfDeclarationDetails { get; set; }
         public IEnumerable<Guid> DocumentGuids { get; set; }
+        public bool IsHidden { get; set; }
```
james-hollinger, line 14: Do you need this on the View Model? It looks like you don't. Also, if you _do_ need it, it should also be `Hidden`.

AMiKaH: No, I swear I remember deleting it from VM lol

prime-dotnet-webapi/Services/EnrolleeService.cs
AMiKaH: `currentSelfDeclarationDocuments` is a list of `SelfDeclarationDocumentViewModel`. When I tried making changes to it those changes weren't reflected in the model itself. I may be missing something.

james-hollinger: I didn't realize that returned View Models. Yes, if you update a view model it wont change the model/entity in the DB. But then why get the view models and then each model individually when you could just fetch the models once?

james-hollinger: Also, if you did want to do it this way, why retrieve the models by doing `.Where(sdd => sdd.EnrolleeId == enrolleeId && sdd.DocumentGuid == currentDocument.DocumentGuid)` instead of just `.Where(sdd => sdd.Id == currentDocument.Id)`?

AMiKaH: The reason I got the view models was because I wanted to reuse the existing `GetSelfDeclarationDocumentsAsync` method and then just flag the documents that we want to hide, We could write another method that gets the models, change the current method to return models, or just do the query in place. 

As to the second question, I guess that was a much better solution that I derped out on lol.

AMiKaH: I am leaning towards changing the current `GetSelfDeclarationDocumentsAsync ` method, and instead of projecting the result inside it to `SelfDeclarationDocumentViewModel` we can instead do the projection where the view model is needed, for now that would be in the controller.

AMiKaH: @james-hollinger please take a look at the changes. I used automapper in the controller and removed the projection from the service method.

prime-dotnet-webapi/Services/EnrolleeService.cs
```diff
@@ -475,6 +475,8 @@ private async Task CreateSelfDeclarationDocuments(int enrolleeId, ICollection<Se
                 return;
             }
 
+            var currentSelfDeclarationDocuments = await GetSelfDeclarationDocumentsAsync(enrolleeId) as IEnumerable<SelfDeclarationDocument>;
```
james-hollinger, line 478: I think you may have miss-understood my earlier comment, I wasn't advocating for making this method return Models instead of View Models. Since this method is being used to return data to the front end this should be returning view models.

Rather, why do a service call here when you can just do
```c#
var currentSelfDeclarationDocuments = await _context.SelfDeclarationDocuments
    .Where(document => document.EnrolleeId == enrolleeId)
    .ToListAsync();
```

_or_ in the `UpdateEnrolleeAsync` method, add the `.Include` for the self declaration documents then just pass them into this method.

james-hollinger: (also, why the cast to `IEnumerable<SelfDeclarationDocument>`? isn't it already returning that type?

AMiKaH: Yeah making the query in place is one of the 3 options I was looking at. The cast is again forgotten there, was testing something.

AMiKaH: The reason I did not want to go with the second option was to avoid code duplication since that query, in your comment, is identical to the query inside `GetSelfDeclarationDocumentsAsync` minus the `ProjectTo` part. That's why I thought we can do the projection outside the service method, for example if we're sending data to frontend, then do the projection right before sending the data to the frontend.

prime-dotnet-webapi/Controllers/EnrolleesController.cs
```diff
@@ -28,20 +29,23 @@ public class EnrolleesController : PrimeControllerBase
...
             IDocumentService documentService)
         {
-            _enrolleeService = enrolleeService;
             _adminService = adminService;
             _businessEventService = businessEventService;
-            _emailService = emailService;
             _documentService = documentService;
+            _emailService = emailService;
+            _enrolleeService = enrolleeService;
+            _mapper = mapper;
```
johnnyl6: `_mapper` does not seem to be used

johnnyl6 (approved):
LGTM

Paahn (approved):
lgtm

Paahn:
in `enrolment-resource.service.ts` you could delete the `from` and `mergeMap` from the declarations

PR #1911, opened 2021-12-31
PRIME-2057 use openshift Sentry

AMiKaH:
Tested on Cloud Sentry.

Will only send reports when a DSN is provided, so we can provide DSNs in Openshift as environment variable only for the environments that we want.

Need to test it with our Openshift Sentry i.e. point to Openshift Sentry Dsn when it is up and running.


Paahn (approved):

johnnyl6 (approved):
LGTM

PR #1930, opened 2022-01-13
PRIME-2021 Paper Enrolment Status Should Not Be Under Review

AMiKaH:
[PRIME-2021](https://bcgovmoh.atlassian.net/jira/software/c/projects/PRIME/boards/1?modal=detail&selectedIssue=PRIME-2021)

### How to test:

1. Login as admin
2. Start a paper enrolment by going to `/paper-enrolment/0/demographic`
3. Go through a few pages in the enrolment then leave without completing it
4. Go to admin panel (You should see the paper enrolment with an **Incomplete** status in Orange)
5. Continue the paper enrolment by going to  `/paper-enrolment/<enrolment-id>/demographic`
6. Finish the enrolment
7. Go back to admin panel (this time you should see the paper enrolment status changed to **Complete** and in Blue)

You can also change the status of the paper enrolment to Locked, Declined ...etc. and the status should change accordingly and displaying the correct colours.

Do a regular enrolment stop in the middle and check (Status should be Editable) the go back and complete it (Status should change to Under Review) showing that regular enrolments are unaffected.


Paahn (approved):
LGTM

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.ts
```diff
@@ -60,6 +60,10 @@ export class EnrolleeTableComponent implements OnInit, OnChanges {
   public EnrolmentStatus = EnrolmentStatusEnum;
   public Role = Role;
   public readonly PAPER_ENROLLEE_GPID_PREFIX = PAPER_ENROLLEE_GPID_PREFIX;
+  public readonly paperEnrolmentStatusMap = new Map<number, string>([
```
hogmann, line 63: Doesn't really reflect this line of code, but the chosen solution for this ticket is as follows,

"Add new status types of IncompletePaperEnrolment and CompletePaperEnrolment, and update the UI to show “Incomplete” and “Complete”."

Not sure if there was other conversations to change this but that requires adding those types in the back end... etc
That is why this ticket was scored at 3 pts, just doing it in the front end is 0.5 of work.

hogmann (changes requested):

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.ts
AMiKaH: Yes I had conversations with James, Martin, and Jason about that. To add the status we would have to make the change in all the places where Editable and Under Review were used and that would be very risky. So we decided to just map the front end.

I forgot to update the ticket on Jira

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.ts
```diff
@@ -105,6 +109,12 @@ export class EnrolleeTableComponent implements OnInit, OnChanges {
     );
   }
 
+  // Check to see if we want to filter/map the status
+  // Conditions: enrollee is a paper enrollee and the status is one of [Under Review, Editable] mapped to [Incomplete, Complete] respectively
+  public filterPaperEnrolmentStatus(enrollee: EnrolleeListViewModel): boolean {
```
hogmann, line 114: We probably want to find a better name for this method, when you see filter our mind first jumps to a method that takes in a list and returns a filtered down list.

With proper naming we can get rid of the comments above. Since comments are just code we have to maintain.

hogmann: to be honest you can just move the logic from the template into a public method
``` getEnrolmentStatus(row: EnrolleeListViewModel): string ```

File: prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html [hunk 3]
```diff
@@ -163,9 +163,13 @@
               [class]="row.currentStatusCode | configCode: 'statuses' | case: 'space' : 'kebab'">
 
             <div class="d-flex align-items-center">
-              <span class="mr-1">{{ row.currentStatusCode | configCode: 'statuses' | default }}</span>
+              <span class="mr-1 status"
+                    [class.paper-enrolment-complete-incomplete]="filterPaperEnrolmentStatus(row)">
```
hogmann, line 167: Is the thought that a paper enrolment (incomplete or complete) will be Blue in colour?

If that is the goal, you should just do something like this
``` [class.paper-enrolment]="row.gpid | contains: PAPER_ENROLLEE_GPID_PREFIX" ``` 
Since we do that elsewhere, but looking at that you may want to change since technically NOBCSC can be found in another part of the gpid. So you can use startsWith instead and with a pipe or not.

File: prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html [hunk 4]
```diff
@@ -163,9 +163,13 @@
...
+                    {{ (filterPaperEnrolmentStatus(row) ? paperEnrolmentStatusMap.get(row.currentStatusCode) : row.currentStatusCode | configCode: 'statuses') | default }}</span>
               <ng-container *ngIf="row.gpid | contains: PAPER_ENROLLEE_GPID_PREFIX; else nonPaperEnrolleeIcons">
-                <mat-icon matTooltip="Paper Submission"
+                <mat-icon class="status"
+                          [class.paper-enrolment-complete-incomplete]="filterPaperEnrolmentStatus(row)"
```
hogmann, line 171: As far as i can tell to be in this *ngif you have to be a paper enrolment and those are the only two status so this should just be true.

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.ts
hogmann, line 63: can make this a map between EnrolmentStatusEnum and string, then remove the comments on the following lines.

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.ts
hogmann: Then use that method to localize the logic and when to use the paper enrolment map

hogmann (changes requested):

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html
AMiKaH: I saw that we can also set the status to locked, declined... etc for paper enrolment in the Adjudicator's Actions

prime-angular-frontend/src/app/modules/adjudication/shared/components/enrollee-table/enrollee-table.component.html
AMiKaH: If I do that, then a declined or locked paper enrolment status will also be in blue which is not correct (not sure if a paper enrolment should be locked or declined, but it's something an admin can do) hence I used the name `paper-enrolment-complete-incomplete`  to separate it from the other coloured statuses. `filterPaperEnrolmentStatus()` already uses `startsWith` .

johnnyl6 (approved):
LGTM

hogmann (approved):
LGTM

PR #1931, opened 2022-01-13
Feature/prime 2055 effective license lookup

hogmann:
This PR cannot be merged until the refactor is done.

To test:

- launch app and look at the lookup api call.
- using a database tool adjust the effective date of a license's detail to the future
- refresh the page and investigate that call
- see that it is no longer shown.


AMiKaH (approved):
LGTM.

Not sure if this applies to backend, but I recall a discussion about using the British spelling for **Licence** instead of **License**

Paahn (approved):
LGTM

james-hollinger (approved):
LGTM

prime-dotnet-webapi-tests/Unit Tests/LookupService.Tests.cs
```diff
@@ -57,5 +58,48 @@ public async void TestGetLookups()
...
+
+            var newResults = await service.GetLookupsAsync();
+
+            Assert.NotNull(newResults);
+            Assert.NotEmpty(newResults.Licenses);
+            Assert.Equal(new LicenseConfiguration().SeedData.Count() + 1, newResults.Licenses.Count);
+            Assert.Empty(newResults.Licenses.Where(l => l.Code == future.Code));
+            Assert.NotEmpty(newResults.Licenses.Where(l => l.Code == past.Code));
+
+            // Clean Up Test Data
```
james-hollinger, line 96: You should not have to do this at the end of your test like this. I'm pretty sure the records wont be there in the next test as is. If records _are_ being left behind, we need a global solution, not having to remove the records at the end of each test like this.

hogmann: ahh true, i though i tried it and got an error but must have been something else.

neophyte57 (changes requested):
Is there a reason the Jira ticket says "Effective October 1 2019" whereas `LicenseDetailConfiguration` uses 2019-09-16?  Does it matter?  @itchison

itchison:
> Is there a reason the Jira ticket says "Effective October 1 2019" whereas `LicenseDetailConfiguration` uses 2019-09-16? Does it matter? @itchison

I dont think this matters - Just happy that a date in the past was used. Although, I am curious where this date came from.

hogmann:
> > Is there a reason the Jira ticket says "Effective October 1 2019" whereas `LicenseDetailConfiguration` uses 2019-09-16? Does it matter? @itchison
> 
> I dont think this matters - Just happy that a date in the past was used. Although, I am curious where this date came from.

It is the seeding date we use for any of the configuration files, was what was decided on way back when and stored as a constant for configurations

neophyte57 (approved):
LGTM

james-hollinger (approved):
LGTM

PR #1932, opened 2022-01-13
Fix PharmaNet dummy client

AMiKaH:
Licence numbers come with leading 0s, so for development, no enrollee will end up automatically adjudicated. trimming the start helps in making enrollees 1-11 automatically adjudicated with the proper licence number.


prime-dotnet-webapi/HttpClients/College Licence API/DummyCollegeLicenceClient.cs
```diff
@@ -17,7 +17,9 @@ public Task<PharmanetCollegeRecord> GetCollegeRecordAsync(string licencePrefix,
                 throw new PharmanetCollegeApiException();
             }
 
-            var info = licenceNumber switch
+            var trimmedLicenceNumber = licenceNumber.TrimStart(new char[] { '0' });
```
james-hollinger, line 20: `string.TrimStart()` has both a `TrimStart(char trimChar)` and a ` TrimStart(params char[]? trimChars)` overload, there is no reason to create an array here. Just use `licenceNumber.TrimStart('0')`.

james-hollinger: Also, you don't need `trimmedLicenceNumber`, so making a new variable here is not nessisary. just do:
```c#
var info = licenceNumber.TrimStart('0') switch
{
  // etc.
}
```

james-hollinger (approved):
LGTM

hogmann (approved):
LGTM

PR #1936, opened 2022-01-15
Fix Email Performance

james-hollinger:
Properly cache email templates rather than compiling and caching every time.


johnnyl6 (approved):
LGTM

AMiKaH (approved):

neophyte57 (changes requested):
I setup some data and currently running send-renewal-emails cron job against this PR.  If web-api pod restarts a lot less than other PR's pod did, by EOD, then things look good

prime-dotnet-webapi/Services/EmailRenderingService.cs
```diff
@@ -15,14 +15,10 @@ public class EmailRenderingService : IEmailRenderingService
         private const string MohEmail = "HLTH.HnetConnection@gov.bc.ca";
         private const string ProviderEnrolmentTeamEmail = "Lori.Haggstrom@gov.bc.ca";
```
neophyte57, line 16: While this isn't part of the PR changes, shouldn't this value be in a ConfigMap, that is, environment variable?  What happens when Lori Haggstrom retires or leaves?

neophyte57:
> I setup some data and currently running send-renewal-emails cron job against this PR. If web-api pod restarts a lot less than other PR's pod did, by EOD, then things look good

The web-api pod is doing very well.  No restarts despite sending out emails every 15 minutes.  Other PR was restarting about once per hour.  Ready to test in `dev` environment.

neophyte57 (approved):
Looks good except that comment about hard-coded email address

hogmann (approved):
L G T M

prime-dotnet-webapi/Services/EmailRenderingService.cs
james-hollinger: Could be a good idea, but having those emails be configurable has never been a requirement

neophyte57: Worthwhile for Product Owner to consider?

itchison: @neophyte57 Can you create a ticket for this comment and then resolve this conversation.

itchison (approved):
SHIP IT

PR #1940, opened 2022-01-17
Fix/downgrade sentry to be compatible with oc sentry

AMiKaH:
Downgraded Sentry to version 1.2 according to Sentry Self Hosted v 9.1.2 docs, this is a compatible version.

### To test:

1. Create a local environment variable called Sentry__Dsn, use the dsn shared on slack.
2. Let me know so I can add you to the project to see the dashboard and the exceptions
3. You need to throw an exception and make sure it shows on the Sentry dashboard

# OpenShift-based method of testing

1.  Add environment variable (i.e. `Sentry__Dsn`) to ReplicationController, delete web-api pod so environment variable takes effect in new pod 
2.  Request https://pr-1940.pharmanetenrolment.gov.bc.ca/api/v1/Lookups to make sure new web-api pod is working 
3.  Rename a lookup table, e.g. ProvinceLookup
4.  Request again, should get HTTP Status 500 error, exception should be reported to Sentry dashboard, web-api pod should continue to run


prime-dotnet-webapi/Sentry/SentryErrorReporter.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+using System.Threading.Tasks;
+
+using Prime;
+using SharpRaven;
```
johnnyl6, line 5: Is `SharpRaven` being used?

AMiKaH: Yes, for creating the `_client` instance and for initializing it in lines 11 and 22 respectively.

johnnyl6 (approved):
LGTM

i5okie (approved):

prime-dotnet-webapi/Sentry/ISentryErrorReporter.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+using System.Threading.Tasks;
+
+namespace SentryCustomReporter
+{
```
neophyte57, line 5: Although capturing these exceptions is definitely useful, not sure if this warrants its own folder, i.e. `Sentry`, and namespace?

AMiKaH: Good point. I only did that because I wasn't sure which namespace to use for it and which directory. Originally I put it in **prime-dotnet-webapi** root, but then it seemed like a lot of files in root so I moved it to its own directory. Will try to find a better home for it

AMiKaH: Now Sentry is part of the Prime.Helpers name space and mopved the files to /Helpers directory

prime-dotnet-webapi/Helpers/SentryErrorReporter.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+using System.Threading.Tasks;
+
+using Prime;
+using SharpRaven;
+
```
neophyte57, line 6: Reorder `using` statements?

prime-dotnet-webapi/Helpers/SentryErrorReporter.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        /// <param name="exception">The exception.</param>
+        /// <returns></returns>
+        /// <exception cref="ArgumentNullException">exception</exception>
+        public async Task CaptureAsync(Exception exception)
+        {
+            if (exception == null)
+                throw new ArgumentNullException(nameof(exception));
+
+            await _client.CaptureAsync(new SharpRaven.Data.SentryEvent(exception));
```
neophyte57, line 37: Use `using` statement rather than fully-qualified classname?

neophyte57 (approved):
🎇 Looks good other than those comments re: `using` statements

prime-dotnet-webapi/Helpers/SentryErrorReporter.cs
AMiKaH: Actually `using Prime` isn't needed anymore after moving to namespace `Prime.Helpers`

PR #1941, opened 2022-01-17
Fix When Selecting HA for enrollee overview validation for OBO site

hogmann:
In the enrollee workflow 

- select care setting Health Authority
- select no certs
- add obo site info
- proceed to overview and try to submit

without fix validation fails


AMiKaH (approved):

Paahn (approved):
lgtm

PR #1958, opened 2022-01-24
PRIME-2099 Cannot submit Device Provider

AMiKaH:
### *** awaiting workflow confirmation from client

[Jira Ticket](https://bcgovmoh.atlassian.net/jira/software/c/projects/PRIME/boards/1?modal=detail&selectedIssue=PRIME-2099)

### How to test:

Try to submit different user enrolments using the following scenarios:

1. Select Device Provider in Care Settings and provide both Device Provider Identifier  and College Certificate in Regulatory page
2. Select Device Provider in Care Settings and provide Device Provider Identifier only in Regulatory Page (you will be forced to provide OBO sites)
3. Select Device Provider in Care Settings and provide College Certificate only in Regulatory Page (you will be forced to provide OBO sites)
4. Select Device Provider in Care Settings and provide nothing in Regulatory Page (you will be forced to provide OBO sites)
5. Select any Care Setting type other than Device Provider

You should be able to complete the submission in all of the above cases.


johnnyl6 (approved):
LGTM

AMiKaH:
Closed until workflow updates are received.

prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts
```diff
@@ -682,15 +682,14 @@ export class EnrolmentFormStateService extends AbstractFormStateService<Enrolmen
...
+   * then they should provide either Device Provider ID or Obo Site
    */
   private hasDeviceProviderOrOboSite(): boolean {
     const { careSettings, certifications, deviceProviderIdentifier, oboSites } = this.json;
     const careSettingPredicate = (cs) => cs.careSettingCode === CareSettingEnum.DEVICE_PROVIDER;
 
     return (careSettings.some(careSettingPredicate))
-      ? ((deviceProviderIdentifier && !!certifications.length)
-        || (!deviceProviderIdentifier && !!oboSites.filter(careSettingPredicate).length))
+      ? (!!deviceProviderIdentifier || !!oboSites.filter(careSettingPredicate).length)
```
hogmann, line 692: So to clear this up, we wanted to add the path in the ticket to accepted paths. Not make it the **Only** excepted path.

hogmann (changes requested):

prime-angular-frontend/src/app/modules/enrolment/shared/services/enrolment-form-state.service.ts
AMiKaH: I could be missing something here, how does the code above make the path in the ticket the only accepted path? 

The tests in the PR description above cover all different paths a user can take in terms of **Care Settings** and **Certification/Device Provider Identifier** and they all passed. I also just added another test case.

hogmann: So this is the old logic
```
((deviceProviderIdentifier && !!certifications.length) 
|| (!deviceProviderIdentifier && !!oboSites.filter(careSettingPredicate).length))
```
We are adding the case
```
deviceProviderIdentifier && !!oboSites.filter(careSettingPredicate).length
```

This does not lead us to
```
(!!deviceProviderIdentifier || !!oboSites.filter(careSettingPredicate).length)
```

hogmann: Since by reading the logic in this function, just having a device provider id would seem like a reasonable path. But it does not work. On initial walkthrough it will take you to obo form and require it. And if from overview page you enter and go back to overview it fails the other validation that is cert || obo

hogmann (approved):

PR #1959, opened 2022-01-25
Feature/prime 2062 minor update im reg

hogmann:
To test this will require a little bit of adjusting in the database, what i will describe in the easiest i found.

- Create an enrolment with care setting **Community Pharmacy** (PRIMET 01 - 11)
- Add a certification
- Select **College of Pharmacists**
- Select any licence type but remember what you selected (for me I used **Full Pharmacist**)
- Use licence number that reflects test card # (1-11)
- Complete the enrolment and get auto approved
- Accept TOA
- Without making changes verify submitting will not cause you to sign a new TOA
- Go into your db management software and find the licence detail for the licence you used (25 if you used full pharm)
- Adjust **NamedInImReg = false** (if 25)
- Resubmit Enrolment
- It should present you with the **OBO TOA** if you used the listed licence classes

**REMEMBER TO ADJUST YOUR LICENCE DETAIL BACK TO ORIGINAL STATE SO YOU DON'T SET A TRAP FOR YOURSELF!!**

Currently added tests for the agreement engine that cover most of the logic of the new rule, will add some tests to cover the rule layer shortly


AMiKaH (approved):
LGTM

prime-dotnet-webapi/Services/Rules/MinorUpdateRules.cs
```diff
@@ -23,14 +24,57 @@ public abstract class MinorUpdateRule : IEnrolleeRule
...
+        }
+    }
+
+    /// <summary>
+    /// Enrollee has a signed a TOA that is of the type the agreementEngine would assign
+    /// at this point in time. In the case the agreementEngine returns indetermanate, return false.
+    /// </summary>
+    public class CorrectToaRule : MinorUpdateRule
+    {
+        public override Task<bool> ProcessRule(Enrollee enrollee)
```
johnnyl6, line 50: Looking at the calling stack upward, it's called by using `await`. should it have modifier `async`?

hogmann: Short answer no.

Long answer is you only require async on a method when there is a await being used inside the method. that will then wrap the method in Task. You will get a warning when adding the async to these methods. Also you **cannot** return Tasks inside an async method since it does the wrapping itself. 

From what I gather is since all these rules are done sync this is kind of the way to wrap the sync methods to return a Task. In the case a rule would have to return something from an async call you would need to make the method async and not wrap that. But the plan was to reduce the amount of dependancies in the Rule files from what i gathered.

james-hollinger (approved):
LGTM

mtpultz (approved):

PR #1963, opened 2022-01-26
PRIME-2080 User needs to sign a new TOA

neophyte57:
image


prime-dotnet-webapi/Services/EnrolleeAgreementService.cs
```diff
@@ -171,5 +178,30 @@ private async Task RenderHtml(params Agreement[] agreements)
...
+                    .ThenInclude(c => c.License)
+                .Include(e => e.EnrolleeCareSettings)
+                .SingleOrDefaultAsync(e => e.Id == enrolleeId);
+            foreach (var cert in enrollee.Certifications)
+            {
+                cert.License.LicenseDetails = await _context.Set<LicenseDetail>()
+                    .OrderByDescending(ld => ld.EffectiveDate)
+                    .Where(ld => ld.LicenseCode == cert.License.Code)
+                    .Where(ld => ld.EffectiveDate <= DateTime.UtcNow)
+                    .ToListAsync();
```
james-hollinger, line 199: I don't see why this `foreach` is necessary, why cant you just `.Include` the license details?

prime-dotnet-webapi/Services/EnrolleeAgreementService.cs
```diff
@@ -120,6 +126,7 @@ public async Task CreateEnrolleeAgreementAsync(int enrolleeId)
         public async Task<Agreement> GetCurrentAgreementAsync(int enrolleeId)
         {
             return await _context.Agreements
+                .Include(a => a.AgreementVersion)
```
james-hollinger, line 129: There is a small issue here: the AgreementVersion has the text of the TOA on it; loading the text for every call to this method is kinda wasteful, especially considering no caller of this method is using the TOA text.

prime-dotnet-webapi/Controllers/EnrolleeAgreementsController.cs
```diff
@@ -243,5 +243,32 @@ public async Task<ActionResult> GetAccessTermSignable(int enrolleeId, int agreem
...
+        [ProducesResponseType(typeof(ApiMessageResponse), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(typeof(ApiResultResponse<bool>), StatusCodes.Status200OK)]
+        public async Task<ActionResult> IsOboToRuAgreementTypeChange(int enrolleeId)
+        {
+            var record = await _enrolleeService.GetPermissionsRecordAsync(enrolleeId);
+            if (record == null)
+            {
+                return NotFound($"Enrollee not found with id {enrolleeId}");
+            }
+            if (!record.AccessableBy(User))
```
hogmann, line 264: Since this is only being used in the enrollee side of things, we can avoid doing this check and add
``` [Authorize(Roles = Roles.PrimeEnrollee)] ```
to the method

Comment (neophyte57) [reply prime-dotnet-webapi/Controllers/EnrolleeAgreementsController.cs:264]: I think this `if (!record.AccessableBy(User))` check is still necessary so that **another** Enrollee cannot check the OBO to RU state of an Enrollee's agreement, however benign that may be.

File: prime-angular-frontend/src/app/modules/enrolment/pages/overview/overview.component.html [hunk 4]
```diff
@@ -85,14 +104,15 @@ <h6 class="mb-0">
...
         {{
         (isInitialEnrolment)
-          ? 'Enrolment Review'
-          : (withinDaysOfRenewal)
-            ? 'Renew Enrolment'
-            : 'Enrolment Information'
+        ? 'Enrolment Review'
+        : (withinDaysOfRenewal)
+        ? 'Renew Enrolment'
+        : 'Enrolment Information'
```
AMiKaH, line 115: Looks like your having some indentation issues here again. Maybe check your IDE settings, or edit file on github.

prime-dotnet-webapi/Services/EnrolleeAgreementService.cs
```diff
@@ -17,16 +20,19 @@ public class EnrolleeAgreementService : BaseService, IEnrolleeAgreementService
...
         private readonly IRazorConverterService _razorConverterService;
+        private readonly IMapper _mapper;
 
         public EnrolleeAgreementService(
             ApiDbContext context,
             ILogger<EnrolleeAgreementService> logger,
             IAgreementService agreementService,
-            IRazorConverterService razorConverterService)
+            IRazorConverterService razorConverterService,
+            IMapper mapper)
```
AMiKaH, line 30: I think we're doing those in alphabetical order (except for context and logger they're always on top).

prime-dotnet-webapi/Controllers/EnrolleeAgreementsController.cs
hogmann: Yeah, that is my bad. Was on the wrong tangent!

prime-dotnet-webapi/Services/EnrolleeAgreementService.cs
```diff
@@ -124,6 +130,16 @@ public async Task<Agreement> GetCurrentAgreementAsync(int enrolleeId)
                 .FirstAsync(at => at.EnrolleeId == enrolleeId);
         }
 
+        public async Task<AgreementType> GetCurrentAgreementTypeAsync(int enrolleeId)
+        {
+            return await _context.Agreements
+                .Include(a => a.AgreementVersion)
```
james-hollinger, line 136: You shouldn't need to `.Include()` the versions if you later do a `.Select()` to get them

prime-dotnet-webapi/Services/EnrolleeAgreementService.cs
```diff
@@ -124,6 +130,16 @@ public async Task<Agreement> GetCurrentAgreementAsync(int enrolleeId)
...
+                .OrderByDescending(a => a.CreatedDate)
+                .Where(a => a.EnrolleeId == enrolleeId)
+                .Select(a => a.AgreementVersion.AgreementType)
+                .FirstAsync();
```
james-hollinger, line 140: This will throw if the given user has no `Agreements`, i.e. every user their first time through the app. This should be `.FirstOrDefault`

prime-dotnet-webapi/Services/EnrolleeAgreementService.cs
```diff
@@ -124,6 +130,16 @@ public async Task<Agreement> GetCurrentAgreementAsync(int enrolleeId)
                 .FirstAsync(at => at.EnrolleeId == enrolleeId);
         }
 
+        public async Task<AgreementType> GetCurrentAgreementTypeAsync(int enrolleeId)
```
james-hollinger, line 133: This needs to be `AgreementType?` since users without `Agreements` have no types

prime-dotnet-webapi/Models/Agreements/AgreementType.cs
```diff
@@ -20,7 +20,7 @@ public enum AgreementType
         [AgreementGroups(AgreementGroup.Organization)]
         CommunityPharmacyOrgAgreement = 5,
 
-        [AgreementGroups(AgreementGroup.Enrollee)]
+        [AgreementGroups(AgreementGroup.Enrollee, AgreementGroup.OnBehalfOf)]
```
james-hollinger, line 23: Good use of the groups 👍

neophyte57: Thanks for introducing this @james-hollinger ... meta-data for enums!

prime-dotnet-webapi/Extensions/EnumExtensions.cs
```diff
@@ -15,5 +15,29 @@ public static class EnumExtensions
                 .OfType<TAttribute>()
                 .SingleOrDefault();
         }
+
+        // To avoid "The annotation for nullable reference types should only be used in code within a '#nullable' annotations context." warning
+#nullable enable
+        public static bool IsValid(this Enum? value)
+#nullable disable
```
james-hollinger, line 22: Rather than using `#pragmas` to hide warnings, it is better to find out why the warnings are happening and fix them.  In this case, its because the base class `Enum` is handled slightly differently than the enums we define. 

`Enum` is an abstract _class_:
image
which means that `Enum` can already be null. Adding the `?` annotation on a class is the nullable reference types feature that is available in newer versions of C#, hence the warning.

When we define our own enums, such as `AgreementType`, they are concrete implementations of the `Enum` abstract class. In C#, concrete implementations of the Enum class (as declared e.g `public enum AgreementType`) are _structs_ which means they are not nullable by default. So when using a concrete enum that you want nullable, that is when you have to use the `AgreementType?` syntax. 99% of the time you are using concrete enums and not manipulating the `Enum` class directly.

The fix here is to remove the question mark.

prime-dotnet-webapi/Services/EnrolleeAgreementService.cs
```diff
@@ -171,5 +187,24 @@ private async Task RenderHtml(params Agreement[] agreements)
...
+                .AsNoTracking()
+                .Include(e => e.Certifications)
+                    .ThenInclude(c => c.License)
+                        .ThenInclude(l => l.LicenseDetails)
+                .Include(e => e.EnrolleeCareSettings)
+                .SingleOrDefaultAsync(e => e.Id == enrolleeId);
+            var agreementDto = _mapper.Map<AgreementEngineDto>(enrollee);
+
+            var expectedAgreementType = AgreementEngine.DetermineAgreementType(agreementDto);
+            return expectedAgreementType.IsValid() && currentAgreementType.IsValid() &&
```
james-hollinger, line 206: Why do `.IsValid()` to check if the enum produced by our system is defined when a simple null check is sufficient?

prime-dotnet-webapi/Services/EnrolleeAgreementService.cs
```diff
@@ -124,6 +131,15 @@ public async Task<Agreement> GetCurrentAgreementAsync(int enrolleeId)
                 .FirstAsync(at => at.EnrolleeId == enrolleeId);
         }
 
+        public async Task<AgreementType?> GetCurrentAgreementTypeAsync(int enrolleeId)
+        {
+            return await _context.Agreements
+                .OrderByDescending(a => a.CreatedDate)
```
hogmann, line 137: We probably want to add a check to make sure the agreement is accepted
``` AcceptedDate != null ```

File: prime-dotnet-webapi/Services/EnrolleeAgreementService.cs (hunk 11 above)
Comment (neophyte57) [reply prime-dotnet-webapi/Services/EnrolleeAgreementService.cs:206]: `0` is returned by `GetCurrentAgreementTypeAsync` if no agreement associated with enrollee

File: prime-dotnet-webapi/Extensions/EnumExtensions.cs (hunk 10 above)
Comment (neophyte57) [reply prime-dotnet-webapi/Extensions/EnumExtensions.cs:22]: Does this comment https://github.com/bcgov/moh-prime/pull/1963/files#r794957970 address?

Comment (james-hollinger) [reply prime-dotnet-webapi/Extensions/EnumExtensions.cs:22]: image
I do not see how that comment address this. The issue is that you are using ` #nullable enable` and ` #nullable disable` to suppress warnings that are informing you that you are miss-using the nullable types feature.

File: prime-dotnet-webapi/Services/EnrolleeAgreementService.cs (hunk 11 above)
Comment (hogmann) [reply prime-dotnet-webapi/Services/EnrolleeAgreementService.cs:207]: Since you are doing a select to an enum, it will default to a 0 value. Since structs can't be null. So in that case we know if it is 0 it was the default.

So we just need to have
``` expectedAgreementType != null && currentAgreementType != 0 & ...etc ```

And we can remove the IsValid method.

As well public methods should be above private and the GetCurrentAgreementTypeAsync should be a private method as it is only called within this method. you could even just move the method to within this method and only have the one method.

Comment (james-hollinger) [reply prime-dotnet-webapi/Services/EnrolleeAgreementService.cs:207]: Looking at the code in `GetCurrentAgreementTypeAsync`:
```C#
return await _context.Agreements
    .OrderByDescending(a => a.CreatedDate)
    .Where(a => a.EnrolleeId == enrolleeId)
    .Where(a => a.AcceptedDate != null)
    .Select(a => a.AgreementVersion.AgreementType)
   .FirstOrDefaultAsync();
```
Examining IntelleSense on the `.Select()`:
image

The enumerable returned is of type `AgreementType`, since that is what `a.AgreementVersion.AgreementType` is.
So when you do `.FirstOrDefault` on it, it returns the default value of `AgreementType`. Since enums are structs, that can't be null. The default value for enums is whatever corresponds to value `0`. But in this case, we actually _want_ null if there are no agreements that match the user. 

There are a few ways of doing this, but the simplest way to to cast the `AgreementType` to an `AgreementType?`:
`.Select(a => (AgreementType?)a.AgreementVersion.AgreementType)`. Now, the `.Select` is of type `AgreementType?` so `.FirstOrDefault` will return `null` instead of `0`.

prime-dotnet-webapi/Services/SubmissionService.cs
```diff
@@ -92,8 +92,9 @@ public async Task SubmitApplicationAsync(int enrolleeId, EnrolleeUpdateModel upd
 
             var minorUpdate = await _submissionRulesService.QualifiesAsMinorUpdateAsync(enrollee, updatedProfile, newestAgreementVersionIds);
             await _enrolleeService.UpdateEnrolleeAsync(enrolleeId, updatedProfile);
+            var hasOboToRuAgreementTypeChange = await _enrolleeAgreementService.IsOboToRuAgreementTypeChangeAsync(enrolleeId);
 
-            if (minorUpdate)
+            if (minorUpdate && !hasOboToRuAgreementTypeChange)
```
james-hollinger, line 97: Please do not do this. We are already checking that the user has the correct TOA type in the minor update rules, we do not need to check again.

prime-dotnet-webapi/Extensions/EnumExtensions.cs
neophyte57: Never mind, I thought the compiler was making me use a nullable type at one point

james-hollinger (approved):
LGTM

hogmann (approved):
LGTM

PR #1964, opened 2022-01-26
fix remote user navigation and form state management

hogmann:
To test navigate every which way from the remote users to remote user page, and try and do actions on a dirty form and cancel.

Only one state that it doesn't handle perfectly is if a user 
- on a fresh page toggles to add remote user
- adds a remote user
- removes said user
- hits the back button

they are then prompted about unsaved changes

### DIFF
--- a/prime-angular-frontend/src/app/modules/site-registration/pages/remote-user-page/remote-user-page.component.ts
+++ b/prime-angular-frontend/src/app/modules/site-registration/pages/remote-user-page/remote-user-page.component.ts
@@ -3,18 +3,19 @@ import { ActivatedRoute, Router } from '@angular/router';
 import { FormGroup, FormArray } from '@angular/forms';
 import { MatDialog } from '@angular/material/dialog';
 
-import { noop, of } from 'rxjs';
+import { noop, Observable, of } from 'rxjs';
+import { map, tap } from 'rxjs/operators';
 
 import { CollegeConfig, LicenseConfig } from '@config/config.model';
 import { ConfigService } from '@config/config.service';
 import { AddressLine } from '@lib/models/address.model';
 import { RouteUtils } from '@lib/utils/route-utils.class';
 import { RemoteUser } from '@lib/models/remote-user.model';
-import { RemoteUserCertification } from '@lib/models/remote-user-certification.model';
 import { AbstractEnrolmentPage } from '@lib/classes/abstract-enrolment-page.class';
 import { FormUtilsService } from '@core/services/form-utils.service';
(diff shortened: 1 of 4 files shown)


AMiKaH (approved):
LGTM

prime-angular-frontend/src/app/modules/site-registration/pages/remote-user-page/remote-user-page.component.ts
```diff
@@ -96,6 +95,41 @@ export class RemoteUserPageComponent extends AbstractEnrolmentPage implements On
...
+  public onSubmit(): void {
+    this.hasAttemptedSubmission = true;
+
+    if (this.checkValidity(this.form)) {
+      this.onSubmitFormIsValid();
+      this.busy = this.performSubmission()
+        .pipe(tap((_) => this.form.markAsPristine()))
+        .subscribe(() => this.afterSubmitIsSuccessful());
+    } else {
+      this.onSubmitFormIsInvalid();
```
johnnyl6, line 129: `onSubmitFormIsValid` and `onSubmitFormIsInvalid` do not seem to be implemented anywhere, wonder  what they are used for?

prime-angular-frontend/src/app/modules/site-registration/pages/remote-users-page/remote-users-page.component.html
```diff
@@ -39,6 +40,21 @@
     </app-alert>
   </div>
 
+  <div class="mb-4"
+       *ngIf="formState.form.dirty && !formState.remoteUsers.controls.length">
```
johnnyl6, line 44: I would use a boolean like `remoteUserRemoved` for this *ngIf explicitly since there is other conditions triggering `formState.remoteUsers.controls.length` to be 0, e.g. clicking back button and choosing discard changes clears the remote users hence length becomes 0. It causes the alert flash briefly before navigating back to previous page.

prime-angular-frontend/src/app/modules/site-registration/pages/remote-user-page/remote-user-page.component.ts
hogmann: So i'm overwriting the functionality from the AbstractEnrolmentPage. This is because the form state in this component points to the parents form state and not the singular remote user.

Those methods are hooks that you can overwrite to add functionality, i chose to leave them in since if later someone wants to implement those hooks they don't have to remember where to add them in the handler.

johnnyl6 (approved):
LGTM🚀

PR #1970, opened 2022-01-28
PRIME-2103 Add loading message to user enrolment submission

AMiKaH:
[Jira Ticket PRIME-2103](https://bcgovmoh.atlassian.net/browse/PRIME-2103)

Using same PR as the changes were made for user enrolment only then a new ticket created to add the loading indicator to all submissions.

### How to test:

Go through a user enrolment. At the end a spinner should show and text will be displayed as well to tell the user that _Wait patiently, submission could take up to a minute_

PR environment may be slow enough to see the spinners, but if you want you can test locally and add a Thread.Sleep in the submission end point.


prime-angular-frontend/src/app/lib/modules/ngx-busy/busy-overlay-message/busy-overlay-message.component.html
```diff
@@ -1,8 +1,11 @@
...
+    </div>
 
-  <!--
-  // NOTE: Already obvious to user that load is occurring this seems
-  // annoying, but left in case it becomes required
-  -->
-
+    <div class="center delay">
+      <h3>{{message}}</h3>
+    </div>
```
johnnyl6, line 9: The message `Wait patiently, the submission can take up to a minute` seems to be on top of existing text in the page and might be causing readability issue if overlapping.

image

prime-angular-frontend/src/app/modules/enrolment/pages/overview/overview.component.html
```diff
@@ -1,4 +1,4 @@
-<app-enrollee-page [busy]="busy">
+<app-enrollee-page [busy]="{busy, message: NG_BUSY_SUBMISSION_MESSAGE}">
```
mtpultz, line 1: So the reason I asked you not to do this ticket for all submissions until the first ticket was reviewed was to allow feedback and avoid making changes across the application that might be unnecessary, as well as, tickets being scored, prioritized top down, and above the ready line regardless of Kanban or Sprint when pulled from the Backlog.

There may be no need to pass the message in through the configuration.  You could likely create a service within NgxBusyModule with show/hide methods that you can pass in the message from your observable using a side-effect.  You might not even need a hide method depending on the components lifecycle.  You'll need to address the opacity in a way that doesn't affect the whole app adversely, but improves readability for this use case.

AMiKaH: At the time when I picked up this ticket, it was already on top of the other tickets above the **Ready** line and was scored. Please check the ticket history on Jira.

The reason I brought the work for all submissions together in this PR was because it was the same work for all _overview_ components. I will revert to add this only for user enrolment and bring the same work over to the other ticket. In any way, I don't see the ng-busy causing a break in the app since it doesn't interfere with any part of it other than displaying the spinner asynchronously.

mtpultz: After refinement it wasn't put at the bottom and that's good feedback.  

No it doesn't break the app but you didn't need to make these changes if you used a service which is why I wanted the review on it.

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy-overlay-message/busy-overlay-message.component.html
AMiKaH: I made the backdrop more opaque if a message is provided.

prime-angular-frontend/src/app/modules/enrolment/pages/overview/overview.component.ts
```diff
@@ -86,9 +91,14 @@ export class OverviewComponent extends BaseEnrolmentPage implements OnInit {
       message: 'When your enrolment is submitted for adjudication, it can no longer be updated. Are you ready to submit your enrolment?',
       actionText: 'Submit Enrolment'
     };
+
     this.busy = this.dialog.open(ConfirmDialogComponent, { data })
       .afterClosed()
       .pipe(
+        tap(() => {
+          this.ngBusyService.updateMessage(NG_BUSY_SUBMISSION_MESSAGE);
```
hogmann, line 99: In enrolment, currently not seeing any messages displayed on the screen.

prime-angular-frontend/src/app/modules/gis-enrolment/pages/ldap-information-page/ldap-information-page.component.html
```diff
@@ -3,19 +3,19 @@
         (ngSubmit)="onSubmit()"
         novalidate
         autocomplete="off">
-  <app-page-header>{{ title }}</app-page-header>
```
hogmann, line 6: I'm assuming this and a few other files are just auto-save from formatting. If you don't have code changes in the files try not to commit formatting like this with PR's takes time to decipher if anything changed as well as if they are not in the direct area that code was changed it an potentially not get tested and bugs can creep in if something goes wrong.

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.config.ts
```diff
@@ -8,5 +8,6 @@ export const busyConfig = {
   delay: 50,
   minDuration: 600,
   disableAnimation: true,
-  wrapperClass: 'ng-busy'
+  wrapperClass: 'ng-busy',
+  message: ''
```
hogmann, line 12: This looks to be left over code from another path taken.

prime-angular-frontend/src/app/lib/modules/ngx-busy/ng-busy.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+
+export class NgBusyService {
+  public message: string;
```
hogmann, line 8: I don't think these should be public.

prime-angular-frontend/src/app/modules/enrolment/pages/overview/overview.component.ts
```diff
@@ -45,6 +48,7 @@ export class OverviewComponent extends BaseEnrolmentPage implements OnInit {
   public IdentityProviderEnum = IdentityProviderEnum;
   public EnrolmentStatus = EnrolmentStatusEnum;
   public hasOboToRuAgreementTypeChange: boolean;
+  public readonly NG_BUSY_SUBMISSION_MESSAGE = NG_BUSY_SUBMISSION_MESSAGE;
```
hogmann, line 51: This line doesn't appear to be used.

hogmann (changes requested):

prime-angular-frontend/src/app/modules/gis-enrolment/pages/ldap-information-page/ldap-information-page.component.html
AMiKaH: I had code changes in all those files, but then reverted I will do that work in another ticket. I will try to remove those from the commit.

prime-angular-frontend/src/app/modules/enrolment/pages/overview/overview.component.ts
AMiKaH: Hmm I just submitted an enrolment and seems to work fine.

If you're testing locally you may not see it as the request will complete before the message is displayed (I added a 5seconds delay before the message shows up)
image

hogmann: image

hogmann: Does the text only show up after 5 sec?

prime-angular-frontend/src/app/lib/modules/ngx-busy/ng-busy.service.ts
```diff
@@ -0,0 +1,28 @@
...
+  providedIn: 'root',
+})
+
+export class NgBusyService {
+  private message: string;
+  private isShowSpinner: boolean;
+
+  constructor() { };
+
+  public setMessage(message: string): void {
```
hogmann, line 13: TS offers access to getters and setters 
https://www.typescripttutorial.net/typescript-tutorial/typescript-getters-setters/

prime-angular-frontend/src/app/modules/enrolment/pages/overview/overview.component.ts
AMiKaH: Yes

prime-angular-frontend/src/app/modules/gis-enrolment/pages/ldap-information-page/ldap-information-page.component.html
AMiKaH: Reverted the fgiles

prime-angular-frontend/src/app/lib/modules/ngx-busy/ng-busy.service.ts
AMiKaH: I used that earlier and decided not to afterwards for some reason. Now using default accessors

prime-angular-frontend/src/app/lib/constants.ts
```diff
@@ -22,3 +22,9 @@ export const DISPLAY_ID_OFFSET = 1000;
  * Paper enrollee GPID filter.
  */
 export const PAPER_ENROLLEE_GPID_PREFIX = 'NOBCSC';
+
+/**
+ * @description
+ * ngBusy submission message
+ */
+export const NG_BUSY_SUBMISSION_MESSAGE = 'Wait patiently, the submission can take up to a minute';
```
neophyte57, line 30: Maybe nitpicky but do we want to tie constants or variable names to an implementation?  e.g. this could just be `BUSY_SUBMISSION_MESSAGE`

prime-angular-frontend/src/app/lib/modules/ngx-busy/ng-busy.service.ts
```diff
@@ -0,0 +1,28 @@
...
+  private _message: string;
+  private _isShowSpinner: boolean;
+
+  constructor() { };
+
+  public set message(message: string) {
+    this._message = message;
+  }
+
+  public set isShowSpinner(showSpinner: boolean) {
```
hogmann, line 17: can simplify to showSpinner, since show is already a verb. Show or not show.

prime-angular-frontend/src/app/modules/enrolment/pages/overview/overview.component.ts
hogmann: Yeah i would stick with immediately

prime-angular-frontend/src/app/lib/constants.ts
AMiKaH: I was also thinking if whether we should make it a constant in the **constants.ts** file or in the module itself so that we can change what the text is based on our needs in the future.

hogmann (approved):
LGTM

johnnyl6 (approved):
LGTM, works well.

prime-angular-frontend/src/app/lib/constants.ts
neophyte57: I like how it is now, in `src\app\lib\modules\ngx-busy\busy.config.ts`

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.config.ts
```diff
@@ -8,5 +8,6 @@ export const busyConfig = {
   delay: 50,
   minDuration: 600,
   disableAnimation: true,
-  wrapperClass: 'ng-busy'
+  wrapperClass: 'ng-busy',
+  message: 'Wait patiently, the submission can take up to a minute.'
```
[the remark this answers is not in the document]
AMiKaH: If the message omitted, the default message would be 'Please wait...'.  from my understanding we do not want to use the word "Please". Also, the message  (and/or spinner) will never show up until we explicitly set their flags. The other way to work around this is to set the default message to an empty string and then pass a busy message whenever we want, not difficult, just adds an extra step for each submission.

prime-angular-frontend/src/app/modules/enrolment/pages/overview/overview.component.ts
```diff
@@ -86,9 +89,14 @@ export class OverviewComponent extends BaseEnrolmentPage implements OnInit {
       message: 'When your enrolment is submitted for adjudication, it can no longer be updated. Are you ready to submit your enrolment?',
       actionText: 'Submit Enrolment'
     };
+
     this.busy = this.dialog.open(ConfirmDialogComponent, { data })
       .afterClosed()
       .pipe(
+        tap(() => {
+          this.ngBusyService.showMessage = true;
```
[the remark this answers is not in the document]
AMiKaH: I went with the setter after deciding to use the default message, now that's not used, will refactor that.

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.config.ts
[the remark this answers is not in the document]
AMiKaH: So, I added the message here, because if we omit that then the default message is "Please wait...", from my understanding, during refinement it was mentioned not to use the word "Please". I could set the default message to an empty string and then we'd pass the message we want but I though why not just add the message in the config since we never use the loader message and if we do we can easily override it. To work around it, we can set the config message to an empty string.

mtpultz (approved):

PR #1973, opened 2022-01-29
Fix Keycloak Client for Users with no Attributes

james-hollinger:
Needed for GIS users, now that we are using the PHSA login. These users have no attributes, so are returning no attribute key on the user representation.


hogmann (approved):
LGTM, odd that is is just an attr called phone. "hey man what's your phone?"

itchison (approved):
Thanks for the quick turnaround

mtpultz (approved):

neophyte57 (approved):

PR #1974, opened 2022-01-31
PRIME-2135 Log requests to web-api

neophyte57:


prime-dotnet-webapi/Helpers/RequestLoggingMiddleware.cs
```diff
@@ -0,0 +1,31 @@
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using System.Threading.Tasks;
+
+namespace My.Service.Logging
```
james-hollinger, line 5: We should update the namespace if this code is intended to be merged

prime-dotnet-webapi/Helpers/RequestLoggingMiddleware.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            _logger.LogInformation(
+                "Incoming request:  {method} {url}",
+                context.Request?.Method,
+                context.Request?.Path.Value);
+
+            await _next(context);
```
johnnyl6, line 33: So the original AC writes: `All requests to the web-api are logged,` **`not just the responses`** `(if any)`
Does the response get logged anywhere in PRIME, if not should it be logged here as well?

neophyte57: So currently in the `develop` branch, the responses are already logged by some mechanism I'm not familiar with.  In OpenShift:
 
image

AMiKaH (approved):
🦞

prime-dotnet-webapi/Startup.cs
```diff
@@ -251,6 +251,8 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IHostApp
             // Matches request to an endpoint
             app.UseRouting();
 
+            app.UseMiddleware<RequestLoggingMiddleware>();
+
```
james-hollinger, line 255: Are you certain this is the correct location to register the middleware? Middleware are invoked in a pipeline in the order they are added to the app, so where you put `app.UseMiddleware` matters. If you look at the line below this, it mentions "Make sure to put this middleware right after `UseRouting()`." so I am unsure that this is the correct place for it.

PR #1996, opened 2022-02-07
PRIME-2130 Only 5 numbers for college of pharmacists license number

hogmann:
1. Log in as an enrollee
2. on the regulatory screen select college of pharmacists
3. enter a license number
4. validation asks for 5 numeric characters


prime-angular-frontend/src/app/shared/components/forms/college-certification-form/college-certification-form.component.html
```diff
@@ -80,6 +80,9 @@
             <mat-error *ngIf="licenseNumber.hasError('alphanumeric')">
               Must contain only alphanumeric characters
             </mat-error>
+            <mat-error *ngIf="licenseNumber.hasError('numeric') || licenseNumber.hasError('length')">
+              Must contain only 5 numeric characters
+            </mat-error>
```
AMiKaH, line 85: Maybe it's also worth putting a `maxLength="5"` for the input on line 76 of this file.

hogmann: Short answer no.

This input is used for many different license types and since this restriction is only for a certain college we don't want to hard code the field. Also we could then store it as a variable, but then we would be tracking validators in two places and it is not needed.

AMiKaH: Yep was just looking at it, commented too soon.

AMiKaH (approved):
Looks good!

prime-angular-frontend/src/app/shared/components/forms/college-certification-form/college-certification-form.component.html
johnnyl6, line 85: How about using a 2-in-1 pattern validator:
```HTML
<mat-error *ngIf="licenseNumber.hasError('pattern')">
```
and 
```typescript
licenseNumberValidators.push(Validators.pattern('^[0-9]{5}$'));
```

hogmann: What is the benefit of say doing it this way?

johnnyl6: I agree with the self descriptive part. However the validation message is triggered by `numeric` or `length`, the user does not really know which rule they violated. It generated the same outcome as the pattern validator unless there are two separate validation messages `Must contain numeric` and `Must contain 5 chars` which gives the user more insights.

johnnyl6 (approved):

prime-angular-frontend/src/app/shared/components/forms/college-certification-form/college-certification-form.component.html
hogmann: Yeah I agree having two messages would probably be more descriptive to the user, but we were asked to follow what we did for the CPSID or whatever the other one is, so the functionality was copied over from there and didn't change anything.

PR #2011, opened 2022-02-08
Fix routing when deleting site and org

hogmann:
1. Create an organization and site
2. navigate to the overview on that site in admin
3. select delete org from actions and choose yes
4. the org is deleted and you are brought back to the list of sites

Repeat for a site since the error was happening there too.


johnnyl6 (approved):
LGTM

AMiKaH (approved):
🚀

PR #2014, opened 2022-02-09
PRIME-2103 add loading indicator PATCH

AMiKaH:
Discovered a bug after merge for [PR1970](https://github.com/bcgov/moh-prime/pull/1970) where the message and spinner persisted for enrollees that continue to auto adjudication and sign their TOAs.

To test, use a test card 1-11 and use the corresponding licence number so they get auto adjudicated.

Submit (you should see the message and spinner). Sign TOA and submit (you shouldn't see anything new)


hogmann (approved):
LTGM

mtpultz (approved):

PR #2015, opened 2022-02-09
Simplify Agreement Engine Test Data

james-hollinger:
Used a generator to clean up test data for the Agreement Engine test rather than inline data.
Also, increased test cases from 42 to 108


AMiKaH (approved):
🚀

hogmann (approved):
I skirted around trying to do it this way but couldn't figure out how to get it to work. i sees now~

PR #2019, opened 2022-02-10
PRIME-2119 Add loading indicator to all submissions

AMiKaH:
[PRIME-2119](https://bcgovmoh.atlassian.net/browse/PRIME-2119)

### How to test:
Try all the different submissions we have (enrolment, site registration, health authority site registration). You should see the spinner with a message "Wait patiently, the submission could take up to a minute." only on the last, submission, page


prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.service.ts
```diff
@@ -18,13 +18,16 @@ export class BusyService {
...
    * @description
    * @param message the message to be displayed when a busy has a busy thing (subscription, promise, array of either... etc)
-   * @param request the observable that will run while the busy message and spinner are shown
+   * @param request the function or observable that will run while the busy message and spinner are shown
    * @returns OperatorFunction
    */
-  public showMessagePipe<T>(message: string, request: Observable<any>): OperatorFunction<T, Observable<any>> {
+  public showMessagePipe<T, R>(message: string, request: R): UnaryFunction<Observable<T>, Observable<any>> {
     return pipe(
       tap((_) => { this._message = message }),
```
mtpultz, line 26: Don't use single line with braces `tap((_) => this._message = message)` similarly when clearing the message

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.service.ts
```diff
@@ -18,13 +18,16 @@ export class BusyService {
   /**
    * @description
    * @param message the message to be displayed when a busy has a busy thing (subscription, promise, array of either... etc)
-   * @param request the observable that will run while the busy message and spinner are shown
+   * @param request the function or observable that will run while the busy message and spinner are shown
    * @returns OperatorFunction
    */
-  public showMessagePipe<T>(message: string, request: Observable<any>): OperatorFunction<T, Observable<any>> {
+  public showMessagePipe<T, R>(message: string, request: R): UnaryFunction<Observable<T>, Observable<any>> {
```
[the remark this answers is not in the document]
mtpultz: What's preventing me from passing a number, object literal, or anything else into this method, or passing a method that returns anything I want?  The typings don't indicate what is expected and the generic is wide open.  Only your code does a small check for a function and as it is your `exhaustMap` will fail since it expects an observable to be returned.  This is why I mentioned you should extend the signature of your parameters to include what you expect.

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.service.ts
```diff
@@ -18,13 +18,16 @@ export class BusyService {
...
     return pipe(
       tap((_) => { this._message = message }),
-      exhaustMap((_) => request),
+      exhaustMap((value: T) => (typeof request === 'function')
```
mtpultz, line 27: You can perform your check in the pipe, but it would be clearer to create const named operator, perform your checks outside the pipe, and then insert the const into the piped operators

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.service.ts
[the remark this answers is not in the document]
mtpultz: Also, is the UnaryFunction entering with a type of `Observable<T>`?

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.service.ts
AMiKaH: Removed, not sure why but the first time I tried without braces it complained, now it seems fine.

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.service.ts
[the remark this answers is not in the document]
AMiKaH: Return type was an oversight from before.

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.service.ts
```diff
@@ -18,14 +18,19 @@ export class BusyService {
   /**
    * @description
    * @param message the message to be displayed when a busy has a busy thing (subscription, promise, array of either... etc)
-   * @param request the observable that will run while the busy message and spinner are shown
+   * @param request the function or observable that will run while the busy message and spinner are shown
    * @returns OperatorFunction
    */
-  public showMessagePipe<T>(message: string, request: Observable<any>): OperatorFunction<T, Observable<any>> {
+  public showMessagePipe<T, R extends Function | Observable<any>>(message: string, request: R): UnaryFunction<Observable<any>, Observable<any>> {
+    const operator = exhaustMap((value: T) => (typeof request === 'function')
```
mtpultz, line 25: You're close.  You can simplify this down to:

```js
  public showMessagePipe<T>(
    message: string,
    observableOrFunction: Observable<T> | (() => Observable<T>)
  ): UnaryFunction<Observable<unknown>, Observable<T>> {
    const operator = isObservable(observableOrFunction)
      ? exhaustMap(() => observableOrFunction)
      : exhaustMap(observableOrFunction);

    return pipe(
      tap((_) => (this.message = message)),
      operator,
      tap((_) => (this.message = ''))
    );
  }
```

It has:
1) More descriptive parameter, 
2) Doesn't use `any`
3) Uses `unknown` for the incoming value into the UnaryFunction
4) Uses `T` on the outgoing value from the UnaryFunction since that's known based on the Observable
4) Don't need to invoke the method passed, and 
5) Makes it easy for the next person to unpackage when scanning then having the ternary is inside the `exhaustMap`

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.service.ts
```diff
@@ -18,14 +18,19 @@ export class BusyService {
   /**
    * @description
    * @param message the message to be displayed when a busy has a busy thing (subscription, promise, array of either... etc)
-   * @param request the observable that will run while the busy message and spinner are shown
+   * @param request the function or observable that will run while the busy message and spinner are shown
    * @returns OperatorFunction
    */
-  public showMessagePipe<T>(message: string, request: Observable<any>): OperatorFunction<T, Observable<any>> {
+  public showMessagePipe<T, R extends Function | Observable<any>>(message: string, request: R): UnaryFunction<Observable<any>, Observable<any>> {
```
mtpultz, line 24: Don't use `any` when possible it's a trap down out of the type system

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.service.ts
```diff
@@ -18,14 +18,19 @@ export class BusyService {
...
+    const operator = exhaustMap((value: T) => (typeof request === 'function')
+      ? request(value)
```
mtpultz, line 26: You don't need to invoke the method and pass the known parameter to a function that takes that parameter.

prime-angular-frontend/src/app/modules/phsa-eforms/pages/available-access/available-access.component.ts
```diff
@@ -44,7 +47,9 @@ export class AvailableAccessComponent implements OnInit {
   public onSubmit(): void {
     if (this.formUtilsService.checkValidity(this.form)) {
       const payload = this.phsaEformsFormStateService.json;
-      this.busy = this.phsaEformsResource.createEnrollee(payload)
+      this.busy = of(noop).pipe(
```
mtpultz, line 50: `noop` is a method with a void return type producing undefined so you should be using `noop()` which was in the example I added in the last comment.

AMiKaH: Wops, have to change its use somewhere else too!

prime-angular-frontend/src/app/modules/phsa-eforms/pages/available-access/available-access.component.ts
```diff
@@ -44,7 +47,10 @@ export class AvailableAccessComponent implements OnInit {
   public onSubmit(): void {
     if (this.formUtilsService.checkValidity(this.form)) {
       const payload = this.phsaEformsFormStateService.json;
-      this.busy = this.phsaEformsResource.createEnrollee(payload)
+      this.busy = of(noop())
```
mtpultz, line 50: GIS/PHSA are quick enrolment workflows with very little processing that we maintain, but aren't "owned" by us like our enrolment, paper enrolment, site reg, HA site reg, and admin.  I wouldn't update them until it's requested.

AMiKaH: Will remove PHSA and GIS

prime-angular-frontend/src/app/lib/modules/ngx-busy/busy.service.ts
```diff
@@ -18,14 +18,21 @@ export class BusyService {
   /**
    * @description
    * @param message the message to be displayed when a busy has a busy thing (subscription, promise, array of either... etc)
-   * @param request the observable that will run while the busy message and spinner are shown
+   * @param observableOrFunction the observable or function that will run while the busy message and spinner are shown
```
mtpultz, line 21: The name of the function and name/typing of the params explains their function.  A short description would be sufficient, and more likely to be read:

```
@description
Pipe for setting/clearing the busy message that wraps the execution of a customizeable operator function.
```

mtpultz (approved):

hogmann (approved):
LGTM

PR #2022, opened 2022-02-10
PRIME-2104 Inform user of changed requirements on license 

neophyte57:


hogmann (approved):
LGTM 👍

prime-angular-frontend/src/app/config/config.service.ts
```diff
@@ -127,6 +127,16 @@ export class ConfigService implements IConfigService {
...
+  public getPrescriberIdType(licenceCode: number): PrescriberIdTypeEnum {
+    const prescriberIdTypes = this.licenses
+      .filter(licenseConfig => licenseConfig.code === licenceCode)
+      .map(licenseConfig => licenseConfig.prescriberIdType);
+
+    return (prescriberIdTypes.length)
+      ? prescriberIdTypes[0]
+      : PrescriberIdTypeEnum.NA;
+  }
+
```
johnnyl6, line 139: compared to other methods in the file, this method seems more like belonging to a mapping or pipe rather than returning config.

prime-angular-frontend/src/app/modules/enrolment/pages/overview/overview.component.ts
```diff
@@ -218,7 +222,18 @@ export class OverviewComponent extends BaseEnrolmentPage implements OnInit {
...
+
+  private isMissingPharmaNetId(certifications: CollegeCertification[]): boolean {
+    return certifications.every((cert: CollegeCertification) => {
+      const prescriberIdType = this.configService.getPrescriberIdType(cert.licenseCode);
+      if (prescriberIdType === PrescriberIdTypeEnum.Mandatory) {
+        return cert.practitionerId === null;
+      }
+      else { return false; }
+    });
+  }
```
james-hollinger, line 238: Is it possible to run the existing validation that we have on the certifications page rather than checking for this _specific_ failure case?

neophyte57: Those validators work on a `AbstractControl` so AFAIK, they can't be repurposed for this ...

```
      this.formUtilsService.setValidators(this.practitionerId, [
        Validators.required,
        FormControlValidators.numeric,
        FormControlValidators.requiredLength(5)
      ]);      
```

prime-angular-frontend/src/app/config/config.service.ts
neophyte57: You're right, was thinking to put method into some utility class but `ConfigService` also has the necessary information, i.e. the `LicenseConfig` information

mtpultz: The ConfigService is supposed to be a "low-level" provider for Lookups.  It shouldn't have utility methods in it like @johnnyl6  mentions.  The `EnrolmentService` would be a good place to have this validation alongside canRequestRemoteAccess, hasAllowedRemoteAccessCareSetting, hasAllowedRemoteAccessLicences, and shouldShowCollegePrefix.

prime-angular-frontend/src/app/modules/enrolment/pages/overview/overview.component.ts
hogmann: I think I miss guided you a little, the only downside to going this route is currently we just display this toast

image

that doesn't really help to give the user and idea of where the error is, but we could intercept that. Im going to play around with something tonight.

mtpultz (approved):

PR #2023, opened 2022-02-11
PRIME-2113 can't delete enrollees

AMiKaH:
[PRIME-2113](https://bcgovmoh.atlassian.net/browse/PRIME-2113)

### How to test:

1. Create enrollee
2. Change status/add notes
3. Delete enrollee (should be deleted)


johnnyl6 (approved):
LGTM

EF's delete is silly to me. Need to load the object (potentially the entire object graph if there's foreign keys) to delete, which is not efficient at all. An alternative way is to configure the OnDelete behavior of the entity to Cascade of the particular relationship, and use a stub entity to delete. This could save the loading roundtrip before the delete.

hogmann:
> LGTM
> 
> EF's delete is silly to me. Need to load the object (potentially the entire object graph if there's foreign keys) to delete, which is not efficient at all. An alternative way is to configure the OnDelete behavior of the entity to Cascade of the particular relationship, and use a stub entity to delete. This could save the loading roundtrip before the delete.

So this is just a weird case where it isn't really ef failing but cascade-delete that fails because of a circular reference we have in our app. Since EnrolleeNotes (a 1st degree child of enrollee) cascade delete is triggered and an attempt to delete them occurs. This process fails since EnrolmentStatusReference ( a 2nd degree child of enrollee) also has a nullable  reference to those EnrolleeNotes. So if an enrollee ever had EnrolmentStatusReference's made with a foreign key to a note this cascade would fail.

This could be solved a few different ways:

1. once being potentially looking at seeing in the db config for these tables maybe being more explicit with what references to delete first.
2. Another is manually deleting or updating the references to remove the enrolleeNoteId foreign key
3. Including the troubled grandchildren which explicitly removes them first before triggering cascade delete

PR #2025, opened 2022-02-11
Fix: Enable Editing site reg handling

hogmann:
Looks like over the course of admin changes the handler for enable editing in site reg fell off, this adds it back in


james-hollinger (approved):
Grayson_Woops

PR #2044, opened 2022-02-15
PRIME-2140 Enrollee admin nav top and bottom

hogmann:
Testing

1. if you don't have more than 100 enrollees, go to ngx-material.modules.ts and change pageSize to something that works for your enrollee pool
2. Use top navigator, changes should be reflected in bottom
3. Use bottom navigator, changes should be reflected in top


AMiKaH (approved):
Looks 💎

johnnyl6 (approved):
LGTM

PR #2048, opened 2022-02-17
PRIME-2144 SPIKE: Dependabot Recommendations

AMiKaH:
Added same version ignore for FE to only create PR for non-breaking changes and patches. Reduced the limit of PRs opened by dependabot to up-to 4 from  up-to 10. 

Once the Sentry PR is merged, I will update this to ignore Sentry updates since we can't go any higher to work with our OC Sentry.


neophyte57:
Should we get rid of the `moh-prime\poc` folder from our repo as Dependabot sometimes makes suggestions for that code.  Also has everything of value been extracted from `moh-prime\pip-angular-frontend` and `moh-prime\pip-dotnet-webapi` @mtpultz @james-hollinger ?

neophyte57:
If we're only concerned about security updates, is this thread applicable:  https://stackoverflow.com/questions/64047526/how-to-get-dependabot-to-trigger-for-security-updates-only

hogmann (approved):
LGTM

AMiKaH:
> If we're only concerned about security updates, is this thread applicable: https://stackoverflow.com/questions/64047526/how-to-get-dependabot-to-trigger-for-security-updates-only

I think we still want to update non security packages, but with less spammy behaviour as in 10 PRs open on Tuesdays. That was my understanding. The security updates are also not affected by the new schedule

james-hollinger:
> Should we get rid of the `moh-prime\poc` folder from our repo as Dependabot sometimes makes suggestions for that code. Also has everything of value been extracted from `moh-prime\pip-angular-frontend` and `moh-prime\pip-dotnet-webapi` @mtpultz @james-hollinger ?

We cannot just drop the POC code. A better course of action would be to exclude it from dependabot.

PR #2060, opened 2022-02-23
PRIME-2147 logout issue

AMiKaH:
Currently if you login for a user enrolment then logout you will get an error like the one below:

image

After the fix you should be redirected to the landing page for PRIME.

How to test:

Login for a user enrolment then log out. You shouldn't see the error screen in the screenshot above.

For the logout part, we see the error in the client logs. the errors were related to trying to load the users' profiles after logging out, at the time they are no longer authenticated and their profile is not loaded. Hence the error "**The user profile was not loaded as the user is not logged in.**". After logging out we were redirecting the users to "https://<env>.oidc.gov.bc.ca/auth/**info**" but it should've been "<env.loginRedirectLink/**info**>" instead. So for local, instead of "https://dev.oidc.gov.bc.ca/auth/info" we should redirect to "http://localhost:4200/info"


neophyte57:
> Currently if you login for a user enrolment then logout you will get an error like the one below:
> 
> image
> 
> After the fix you should be redirected to the landing page for PRIME.
> 
> How to test:
> 
> Login for a user enrolment then log out. You shouldn't see the error screen in the screenshot above.

Would it be helpful to **summarize** the discussion about this ticket in https://bcgovmoh.atlassian.net/browse/PRIME-2147?

hogmann (approved):
Look like it works

johnnyl6 (approved):
Works well in both local and dev

PR #2088, opened 2022-03-04
PRIME-2148 Add reference number to subject line of email to hibc

johnnyl6:
- Includes siteId in subject line for email sent to hibc
  - site registration submission
  - site approval
- PRIME currently does not send email when site is updated


AMiKaH (approved):
LGTM. Just feedback on the ticket rather than the PR, we have the Site Id in the subject line, and in the body we have PEC both are 2 different numbers.

prime-dotnet-webapi/Services/EmailRenderingService.cs
```diff
@@ -161,7 +161,7 @@ public async Task<Email> RenderSiteRegistrationSubmissionEmailAsync(LinkedEmailV
                 from: PrimeEmail,
                 to: MohEmail,
                 cc: PrimeSupportEmail,
-                subject: "PRIME Site Registration Submission",
+                subject: $"[{siteId}] PRIME Site Registration Submission",
```
neophyte57, line 164: Just want to confirm that adding the `siteId` is **not** applicable to Community Pharmacy sites?

johnnyl6: > Just want to confirm that adding the `siteId` is **not** applicable to Community Pharmacy sites?

I believe the requirement was for emails sent to hibc, whereas Community Pharmacy sites emails are sent to PrimeSupport.

PR #2097, opened 2022-03-07
PRIME-2071 code climate diff changes

AMiKaH:
The update now runs code climate only when a PR is first created and then updated or when the PR is closed and merged to develop. This should prevent the code climate checks from running on the same commit SHA multiple more than once


johnnyl6 (approved):
🚀

AMiKaH:
Confirmed, new version. Updated new TOA to v2 with effective date of March 16, 2022

neophyte57 (approved):
Approving but @AMiKaH  can you please add some high-level comments re: the intent of the YAML configuration when you have this better figured out

PR #2103, opened 2022-03-15
PRIME-2186 Valid email address fails validation

johnnyl6:
allow single quote in localpart of email

### DIFF
--- a/prime-angular-frontend/src/app/lib/validators/form-control.validators.ts
+++ b/prime-angular-frontend/src/app/lib/validators/form-control.validators.ts
@@ -62,7 +62,7 @@ export class FormControlValidators {
    */
   public static email(control: AbstractControl): ValidationErrors | null {
     if (!control.value) { return null; }
-    const regExp = /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,}$/i;
+    const regExp = /^[a-z0-9._%'+-]+@[a-z0-9.-]+\.[a-z]{2,}$/i;
     // Affixed spaces does not invalidate the entry, and should
     // be sanitized by on submission and by the server
     const valid = (control.valid && regExp.test(control.value));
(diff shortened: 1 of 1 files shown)


prime-angular-frontend/src/app/lib/validators/form-control.validators.ts
```diff
@@ -62,7 +62,7 @@ export class FormControlValidators {
    */
   public static email(control: AbstractControl): ValidationErrors | null {
     if (!control.value) { return null; }
-    const regExp = /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,}$/i;
+    const regExp = /^[a-z0-9._%'+-]+@[a-z0-9.-]+\.[a-z]{2,}$/i;
```
neophyte57, line 65: I don't know if this Wikipedia article is definitive but even for unquoted local-parts of the email address, a lot of atypical characters are allowed **!#$%&'*+-/=?^_`{|}~** 😱  https://en.wikipedia.org/wiki/Email_address  Should they all be allowed?

AMiKaH (approved):
🚀

hogmann (approved):
LGTM

prime-angular-frontend/src/app/lib/validators/form-control.validators.ts
neophyte57: Note that @itchison is checking with CHES team 
image

johnnyl6: > Note that @itchison is checking with CHES team image

[CHES](https://github.com/bcgov/common-hosted-email-service) uses [validatorjs](https://github.com/validatorjs/validator.js) which allows single quote in emails.

neophyte57 (approved):
🎇

PR #2106, opened 2022-03-17
PRIME-2189 Resubmitting a declined site submission will not resubmit

johnnyl6:
- only allow site submission when site is in editable state


hogmann (approved):

neophyte57 (approved):
😊👍

PR #2111, opened 2022-03-23
PRIME-2179 allow addition of multiple addresses for paper enrolment

AMiKaH:
UI Has been confirmed with Pilar.

**How to test:**

1. Login as an admin and create a new paper enrolment
2. Enter an address then click the "+ Add additional address" button
3. Enter new addresses
4. Complete enrolment

**Expected result:**

1. You should be able to see the additional addresses being patched on the demographic page
2. You should see the additional addresses in the overview page
3. You should be able to see the additional addresses in under the "Enrolment" card for paper enrollees only, it can show either empty or with additional addresses. For non paper enrollees that section shouldn't be visible


prime-angular-frontend/src/app/modules/paper-enrolment/pages/demographic-page/demographic-page.component.ts
```diff
@@ -132,8 +133,4 @@ export class DemographicPageComponent extends AbstractEnrolmentPage implements O
 
     this.routeUtils.routeRelativeTo(nextRoutePath);
   }
-
-  protected onSubmitFormIsInvalid(): void {
-    this.showAddressFields = true;
-  }
```
johnnyl6, line 138: previously this is used to show validations when address is empty. it seems that now we lost the ability to show the validation errors if leaving the address field empty after clicking `save and continue` button.

AMiKaH: So I added a new boolean to force show when the form is invalid. I had to change the previous boolean to a getter since we now have more than one address and we  want to check each one for whether or not we should display the form fields. 

I don't think it's possible to have an additional address field be empty and its fields hidden, so in the method `onSubmitFormIsInvalid` I am only forcing the main address fields to be shown, instead of the main one and all additional addresses

prime-dotnet-webapi/Models/Addresses/Address.cs
```diff
@@ -9,7 +9,8 @@ public enum AddressType
     {
         Physical = 1,
         Mailing = 2,
-        Verified = 3
+        Verified = 3,
+        Additional = 4,
```
hogmann, line 13: We might want to use a more explicit name for these since additional is super vague and currently paper enrolment is the only place this feature is built for.

prime-angular-frontend/src/app/shared/components/enrollee/enrollee-review/enrollee-review.component.html
```diff
@@ -54,6 +54,11 @@
                   [showIfEmpty]="true"
                   (route)="onRoute(demographicRoutePath)"></app-address-view>
 
+<app-address-view title="Additional Addresses"
```
hogmann, line 57: address-view is a simple component that takes an address and displays it. 

What benefit do you get changing how that component functions, over using an NgFor in this component?

hogmann (changes requested):

prime-angular-frontend/src/app/modules/paper-enrolment/pages/demographic-page/demographic-page.component.html
```diff
@@ -84,11 +84,44 @@
 
       <app-address-form [form]="formState.physicalAddress"
                         [showManualButton]="true"
-                        [showAddressFields]="showAddressFields"></app-address-form>
+                        [showAddressFields]="showAddressFields(formState.physicalAddress.value)"></app-address-form>
+
+      <app-page-subheader2>
+        <ng-container appPageSubheader2Summary>Provide any additional addresses.</ng-container>
```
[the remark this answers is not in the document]
AMiKaH: - Will check the space. I also wasn't comfortable with the 2 '+Add' lines. 
- I can add a divider between the 2 buttons just so they don't look like they're grouped together
- I liked that as it shows its different but can put them on same width

With that being said after checking with Pilar and the fact that this will only be used by admins on rare occasions we decided not to spend more time on it, hence the aesthetics weren't a priority.

prime-angular-frontend/src/app/modules/paper-enrolment/pages/demographic-page/demographic-page.component.ts
johnnyl6: somehow after the fix it's not saving the additional address, only first address is saved.

prime-angular-frontend/src/app/modules/paper-enrolment/pages/demographic-page/demographic-page.component.html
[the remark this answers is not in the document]
hogmann: Yeah the buttons are not too bad having them like that since low use case.

Understand aesthetics are a priority but you added additional code/work and made it look not inline with how the rest of the page works.

Looking into how the address form works it uses row so having this external row just adds unnecessary spacing and margins 
image

There is no need to have these two divs and added styling, when a ng-container will suffice.

prime-angular-frontend/src/app/modules/paper-enrolment/pages/demographic-page/demographic-page.component.ts
AMiKaH: Fixed.

prime-angular-frontend/src/app/modules/paper-enrolment/pages/demographic-page/demographic-page.component.html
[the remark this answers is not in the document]
AMiKaH: Removed the divs,

- Space is now the same for enrollee's addresses and additional addresses titles and their corresponding fields.
- Added a divider between the 2 `Add` buttons
- Width now matches, minus the X mark width

prime-dotnet-webapi/Services/EnrolleePaperSubmissionService.cs
```diff
@@ -110,9 +110,28 @@ public async Task UpdateDemographicsAsync(int enrolleeId, PaperEnrolleeDemograph
                 .ThenInclude(a => a.Address)
                 .SingleOrDefaultAsync(e => e.Id == enrolleeId);
 
+            var additionalAddresses = enrollee.Addresses
```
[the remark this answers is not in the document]
AMiKaH: Yeah that was a bug on the `createPaperEnrollee` side, I didn't add the additional addresses part, but it worked on update only. Now fixed

prime-angular-frontend/src/app/shared/components/enrollee/enrollee-review/enrollee-review.component.html
AMiKaH: Reverted changes to the `AddressViewComponent`, using `*ngFor` instead

AMiKaH: The reason I did that was because we can have multiple additional addresses under one title. Now using a combination of ngFor and app-subheader. Reverted the work on the `AddressViewComponent`

prime-angular-frontend/src/app/shared/components/enrollee/enrollee-review/enrollee-review.component.html
```diff
@@ -54,6 +54,22 @@
...
+  </app-page-subheader>
+  <ng-container *ngFor="let additionalAddress of enrollee?.additionalAddresses; let index = index">
+    <app-page-subheader>
+      <ng-container appPageSubheaderSummary>
+        <strong> Address: {{ index + 1 }}</strong>
+      </ng-container>
+    </app-page-subheader>
+    <app-address-view [address]="additionalAddress"
+                      [showRedirect]="showEditRedirect"
+                      [showIfEmpty]="enrollee.gpid | contains: PAPER_ENROLLEE_GPID_PREFIX: 'startsWith'">
```
hogmann, line 69: This logic is no longer needed, since we only show if the address exists in the for

prime-angular-frontend/src/app/shared/components/enrollee/enrollee-review/enrollee-review.component.ts
```diff
@@ -33,6 +34,7 @@ export class EnrolleeReviewComponent {
   public CareSettingEnum = CareSettingEnum;
   public EnrolmentRoutes = EnrolmentRoutes;
   public AdjudicationRoutes = AdjudicationRoutes;
+  public readonly PAPER_ENROLLEE_GPID_PREFIX = PAPER_ENROLLEE_GPID_PREFIX;
```
hogmann, line 37: As above this will no longer be needed.

hogmann (approved):

johnnyl6 (approved):
LGTM

PR #2112, opened 2022-03-23
Fix/prime 2181 overview validation health auth enrollees

hogmann:
To Test:

1. Create an enrolment with Health Authority as it's care setting through to the overview
2. Go into database and remove EnrolleeHealthAuthorities related to that enrolment
3. refresh overview page
4. Error should be displayed and indicator should be on care setting

ps
- Have it now if any error is true it will display the error at the top as well as disable the button. So to add a new error add it to the object and then put it on the property in the enrollee-review component
- Also update the back buttons to properly navigate back to overview
- On the pharmanet-enrolment-summary after accepting toa if you had selected device provider as a care setting the page would break. Added device provider as an option to email so the page doesn't break


prime-angular-frontend/src/app/shared/components/enrollee/enrollee-review/enrollee-review.component.html
```diff
@@ -116,6 +116,8 @@
   <ng-container *ngFor="let careSetting of enrolment.careSettings">
 
     <app-enrollee-property title="Care Setting"
+                           [hasError]="enrolmentErrors?.missingHealthAuthorityCareSetting && careSetting.careSettingCode === CareSettingEnum.HEALTH_AUTHORITY"
+                           errorMessage="Must select health authorities when selecting health authority as your care setting."
```
AMiKaH, line 120: Just a suggestion, if we can sentence-case the "Health Authority" part since it's how it shows in the dropdown? as in 

`Must select health authorities when selecting Health Authority as your care setting.`

Also using **select** twice makes reading the error a little awkward, I'm just throwing this in for example to clarify for the users:

`Must specify the health authorities you work at when selecting Health Authority as your care setting.`

AMiKaH (approved):
LGTM, just that one comment.

prime-angular-frontend/src/app/shared/components/enrollee/enrollee-review/enrollee-review.component.html
hogmann: updated the test

johnnyl6 (approved):
LGTM

PR #2115, opened 2022-03-25
Fix/site business licence upload

hogmann:
To re-create the bug 

1. Create new site submission
2. save and submit the business licence page
3. refresh or logout/login
4. Continue through the workflow and submit your site

This can also be seen on first submission when you get to the overview page and go back to edit the business licence page.

The site will not have a document associated with the business licence.

Updated the overview and api to handle users coming back after having this bug affect them and need to add their business licence again.


itchison (approved):
lgtm

mtpultz (approved):

PR #2129, opened 2022-04-07
Fix/renewal enrolment workflow

hogmann:
Some updates to when a user comes back to renew, and are in the manual workflow. When they log back in after being adjudicated and approved they are brought to the overview page and it is not clear they need to sign a TOA.

Added some alerts with links.

Shoe horned in a couple small wordsmithing in as well


Paahn (approved):
LGTM

PR #2131, opened 2022-04-11
Feature/prime 2191 onboard dental college

hogmann:
On boarding for the dental surgeon and Optometrist licence classes. As well as overview validation that checks to see if your licence class / college combination is possible. 

Also fixed a bug where on going back to the certification page validation was wiped and not reset.

To test, local first save a dental surgeon + optometrist to the point of overview. Then apply these changes. You should see overview validation prompting cert errors. Once on that page the licence class will have options as well as it saying registration number.

https://bcgovmoh.atlassian.net/browse/PRIME-2196

https://bcgovmoh.atlassian.net/browse/PRIME-2191


Paahn (approved):
LGTM

PR #2141, opened 2022-04-25
Fix: Enrolment Overview Error Miss Reporting

hogmann:
Enrolment errors should only represent states that mean the overview is false, not just helper methods for deciding error messaging. Not having a cert is not an error but just used in the review page for message logic.


Paahn (approved):
Went through an enrolment without a cert and didn't get error on overview, LGTM unless there is anything else to test

PR #2171, opened 2022-07-06
PRIME-2236 Role for paper enrolment

neophyte57:
1. Setup your user profile in Keycloak to only have the "Paper Enrolment Clerk" group.  That is, do not be part of the "Manager" group e.g. 

image
	
2. Being only in this group should reduce your Roles to the following:

image
	
3. Now log into https://pr-2171.pharmanetenrolment.gov.bc.ca/admin and try to function as a "Paper Enrolment Clerk"

### DIFF
--- a/documentation/Keycloak settings/realm-export-dev.json
+++ b/documentation/Keycloak settings/realm-export-dev.json
@@ -63,6 +63,15 @@
         "attributes": {}
       },
       {
+        "id": "68702c75-bf68-4605-85b0-4b1e84aa894d",
+        "name": "paper_enrolment_only_view",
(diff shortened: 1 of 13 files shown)


neophyte57:
From Nick Mailhot:

image

bergomi02:
I believe the banner maintenance creation link is the one from Site Registrations.
image
However, the publish banner button is disabled. 
image
If I go to Maintenance section and try to view different items:
image
they are all empty, except Licence Classes - get 404 errors:
image
image
image

bergomi02:
The maintenance links are disabled now. 
image
image
Everything looks good to me.

bergomi02 (approved):

PR #2188, opened 2022-08-24
PRIME-2276 Update logging for non-prescriber check

bergomi02:
- To record the correct prefix when record not found.


neophyte57 (approved):
LGTM, tested with these cases:

[X] Nurse licence class enrollee that can't be found in Pharmanet
[X] Non-nurse licence class enrollee that can't be found in Pharmanet
[X] Nurse licence class enrollee that CAN be found in Pharmanet
[X] Non-nurse licence class enrollee that CAN be found in Pharmanet

PR #2195, opened 2022-09-06
PRIME-2275 API to return HPDIDs that have been updated

neophyte57:
To assist with testing, see 

image


bergomi02 (approved):
image
Tested against local database and works as expected.

PR #2199, opened 2022-09-08
PRIME-2285 TOA logic update

bergomi02:
Implement logic that highlighted in the red rectangle on the flow chart.
image

https://pr-2199.pharmanetenrolment.gov.bc.ca/info


neophyte57 (approved):
Nice work, I tested in both cases (Pharmanet ID provided and not)

PR #2202, opened 2022-09-15
PRIME-2292 Upper Case PharmaNet ID

bergomi02:
- Use CSS to capitalize PharmaNet ID
- Use onChange event to convert the input string to uppercase

Note: Please let me know if there is better implementation.

https://pr-2202.pharmanetenrolment.gov.bc.ca/info


neophyte57:
Might be overkill but can the back-end also enforce that the provided Pharmanet ID is converted to upper case before saving?

bergomi02:
@neophyte57 
Updated backend to capitalize the PharmaNet ID. Ready to review again.

PR #2217, opened 2022-11-02
PRIME-2328 add college name in provisioner link page

bergomi02:


prime-dotnet-webapi/Models/Api/EnrolleeCertDto.cs
```diff
@@ -8,6 +8,7 @@ public class EnrolleeCertDto
         public string PractRefId { get; set; }
         public string CollegeLicenceNumber { get; set; }
         public string PharmaNetId { get; set; }
+        public string CollegeName { get; set; }
```
neophyte57, line 11: I think this isn't necessary as the front-end has the data to do the translation from `CollegeCode`, e.g. see `CertificationViewModel`

bergomi02: I wish I noticed that before. Code is updated accordingly.

PR #2218, opened 2022-11-07
Updating postgress source repo

emuthiah:
Looks like postgress removed "stretch" repo. We are using mcr.microsoft.com/dotnet/aspnet:5.0 which is based on Debian "buster".


bergomi02 (approved):
LGTM

PR #2222, opened 2022-11-10
Feature/PRIME-2306 Store Org. Agreement and ToA in document manager

bergomi02:
Note:
- Organization Agreement PDF will be created and stored in Document Manager. Then it will be treated as uploaded agreement PDF that a record will be created in SignedAgreementDocument table as a reference. 
- Similar to Organization Agreement, TOA PDF will be created and stored in Document Manager, and a record will be created in SignedAgreementDocument table. 
- Default filename of the organization agreement is "Organization-Agreement.pdf", which matches the name in the existing frontend code.
- Default filename of the TOA is "Terms-of-Access.pdf", which matches the name in the existing frontend code.
- TOA PDF can be downloaded from Terms-of-Access page logged as Enrollee. However, it can't be downloaded by Admin at this point.
image

https://pr-2222.pharmanetenrolment.gov.bc.ca/info

### DIFF
--- a/prime-dotnet-webapi/Controllers/OrganizationsController.cs
+++ b/prime-dotnet-webapi/Controllers/OrganizationsController.cs
@@ -11,6 +11,8 @@ using Prime.Models;
 using Prime.Models.Api;
 using Prime.Services;
 using Prime.ViewModels;
+using Prime.HttpClients;
+using Prime.HttpClients.DocumentManagerApiDefinitions;
 
 namespace Prime.Controllers
 {
@@ -30,6 +32,7 @@ namespace Prime.Controllers
         private readonly IOrganizationService _organizationService;
         private readonly ISiteService _siteService;
         private readonly IPartyService _partyService;
+        private readonly IDocumentManagerClient _documentManagerClient;
 
 
         public OrganizationsController(
@@ -42,7 +45,8 @@ namespace Prime.Controllers
             IOrganizationClaimService organizationClaimService,
             IOrganizationService organizationService,
             ISiteService siteService,
-            IPartyService partyService)
+            IPartyService partyService,
+            IDocumentManagerClient documentManagerClient)
         {
             _adminService = adminService;
             _businessEventService = businessEventService;
@@ -54,6 +58,7 @@ namespace Prime.Controllers
             _organizationService = organizationService;
             _siteService = siteService;
             _partyService = partyService;
+            _documentManagerClient = documentManagerClient;
         }
 
         // GET: api/Organizations/5
@@ -484,6 +489,22 @@ namespace Prime.Controllers
             if (organization.PendingTransfer && await _organizationService.IsOrganizationTransferCompleteAsync(organizationId))
             {
                 await _organizationService.FinalizeTransferAsync(organizationId);
+            }
+
+
+            if (!organizationAgreementGuid.HasValue)
+            {
+                var filename = "Organization-Agreement.pdf";
+                // get the agreement
+                var agreement = await _organizationAgreementService.GetOrgAgreementAsync(organizationId, agreementId, true);
+                // store it into document manager
+                var documentGuid = await _documentManagerClient.SendFileAsync(new System.IO.MemoryStream(Convert.FromBase64String(agreement.AgreementContent)), filename, DestinationFolders.SignedOrgAgreements);
+                // add a record in signed organization agreement table, treat it as uploaded agreement
+                var signedAgreement = await _organizationService.AddSignedAgreementAsync(organizationId, agreementId, documentGuid, filename);
+                if (signedAgreement == null)
+                {
+                    return BadRequest("Signed Organization Agreement could not be created; network error or upload is already submitted");
+                }
             }
 
             return NoContent();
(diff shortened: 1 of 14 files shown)


prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -487,6 +492,22 @@ public async Task<ActionResult> AcceptOrganizationAgreement(int organizationId,
...
             }
 
+
+            if (!organizationAgreementGuid.HasValue)
+            {
+                var filename = "Organization-Agreement.pdf";
+                // get the agreement
+                var agreement = await _organizationAgreementService.GetOrgAgreementAsync(organizationId, agreementId, true);
+                // store it into document manager
+                var documentGuid = await _documentManagerClient.SendFileAsync(new System.IO.MemoryStream(Convert.FromBase64String(agreement.AgreementContent)), filename, DestinationFolders.SignedOrgAgreements);
```
neophyte57, line 502: Wouldn't the latest signed org agreement for **any** organization overwrite the previous files since the filename is identical and the folder is also constant?  I am still trying to confirm as I am not finding the files where I expected them ... where should they be saving to on the file system?

bergomi02: In the document manager docker console (my local environment), I can see the filename has been encoded. Therefore, the files with the same name will not be replaced or overwritten. 
image

prime-dotnet-webapi/Services/AgreementService.cs
```diff
@@ -66,12 +66,15 @@ public async Task<int> GetLatestAgreementVersionIdOfTypeAsync(AgreementType type
                 .FirstAsync();
         }
 
-        public async Task<SignedAgreementDocument> AddSignedAgreementDocumentAsync(int agreementId, Guid documentGuid)
+        public async Task<SignedAgreementDocument> AddSignedAgreementDocumentAsync(int agreementId, Guid documentGuid, string filename = "")
         {
-            var filename = await _documentClient.FinalizeUploadAsync(documentGuid, DestinationFolders.SignedAgreements);
-            if (string.IsNullOrWhiteSpace(filename))
+            if (filename == null)
```
neophyte57, line 71: Should this comparison be with empty string rather than `null` ... I'm not sure ... related to this, when is `SubmissionsController.AcceptToa` called?

prime-dotnet-webapi/Controllers/OrganizationsController.cs
```diff
@@ -487,6 +492,22 @@ public async Task<ActionResult> AcceptOrganizationAgreement(int organizationId,
...
+                // add a record in signed organization agreement table, treat it as uploaded agreement
+                var signedAgreement = await _organizationService.AddSignedAgreementAsync(organizationId, agreementId, documentGuid, filename);
+                if (signedAgreement == null)
+                {
+                    return BadRequest("Signed Organization Agreement could not be created; network error or upload is already submitted");
```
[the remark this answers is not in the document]
bergomi02: Yes. The error message is copied from Organization Agreement. There is a database constraint to ensure it is 1-to-1 relationship between Agreement and SignedAgreementDocument.

prime-dotnet-webapi/Services/AgreementService.cs
bergomi02: good catch. it is fixed.

prime-dotnet-webapi/Services/SubmissionService.cs
```diff
@@ -284,6 +297,22 @@ private async Task<bool> AcceptToaAsync(Enrollee enrollee, object additionalPara
...
             }
+            else
+            {
+                // for regular enrollee, get the pending agreement, create the PDF and store it in document manager
+                var pendingAgreementId = enrollee.Agreements.OrderByDescending(a => a.CreatedDate).Select(a => a.Id).First();
+                Agreement agreement = await _enrolleeAgreementService.GetEnrolleeAgreementAsync(enrollee.Id, pendingAgreementId, true);
+                var html = await _razorConverterService.RenderTemplateToStringAsync(RazorTemplates.Agreements.PdfNoSignature, agreement);
+                var pdfbinary = _pdfService.Generate(html);
+                var filename = "Terms-Of-Access.pdf";
+                var documentGuid = await _documentManagerClient.SendFileAsync(new System.IO.MemoryStream(pdfbinary), filename, DestinationFolders.SignedOrgAgreements);
```
neophyte57, line 308: Need a TODO: comment or something because it is somewhat misleading that we are saving enrollee TOAs into the `/app/document_uploads/signed_org_agreements` folder

prime-dotnet-webapi/Views/Agreements/AgreementPdfNoSignature.cshtml
```diff
@@ -0,0 +1,14 @@
...
+@model Agreement
+
+@{
+  Layout = "~/Views/Partials/_BaseLayout.cshtml";
+}
+
+@section Title { Terms of Access }
+
+@section Theme {
+@await Html.PartialAsync("~/Views/Partials/_AgreementTheme.cshtml")
```
neophyte57, line 11: Could use an indent at beginning of line

neophyte57 (approved):
😂

PR #2242, opened 2022-12-12
PRIME-2326 Health Authority Site Reg - By default don't ask for the siteID

yousiefc:


bergomi02:
@yousiefc 

The new toggle looks good and the Site ID saved properly, but I found couple things.

When clicking "Save and Continue" button without toggling the Site ID switch, a modal pops up. 
image
I believe we should remove the modal pop since a switch has been added for user to identify if there is a Site ID or not. 

When Site ID is provided and advances next step. 
image
Then go back, the Site ID switch is switched to off is incorrect.
image
Also, an error appear when toggle the switch back on. 
image

yousiefc:
Updated with the requested fixes

bergomi02:
@yousiefc 

Couple items:
- First, when the switch is being marked off, I think the system should clear the Site ID field. Since it hides the Site ID field, as a user, I expect to the Site ID will be removed. Do you agree?
- Second, when the Site ID field is filled and the switch is off, a server error pops up when I mark the switch to on. 
image

yousiefc:
@bergomi02

I agree that would make sense. Should be a quick fix.

And I will take another look at that error.

bergomi02 (approved):
LGTM, Well done!

PR #2247, opened 2022-12-19
Fix/PRIME-2338 enrollee should not access share gpid with expired toa

yousiefc:
Enrollees can no longer access the "Share GPID" page if their TOA is expired.


bergomi02 (approved):
Looks good. Works as expected. Also, I did not know the icon would change based on the status.

PR #2251, opened 2023-01-05
PRIME-2373 SendRemoteUsersUpdatedAsync failed

bergomi02:
There is a bug in getting the business license download link.


prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -117,7 +117,7 @@ public async Task SendSiteReviewedNotificationAsync(int siteId, string note)
 
         public async Task SendRemoteUsersUpdatedAsync(CommunitySite site)
         {
-            var downloadUrl = await _emailDocumentService.GetBusinessLicenceDownloadLink(site.Id);
+            var downloadUrl = await _emailDocumentService.GetBusinessLicenceDownloadLink(site.BusinessLicence.Id);
```
neophyte57, line 120: Question:  Is the erroneous behavior addressed by this PR that either the business licence would not be downloadable or that the wrong business licence would be linked to?

image

bergomi02: The only issue this PR fixing is that the SendRemoteUsersUpdatedAsync was throwing exception and the targeted email was failed to send out. I am not aware of other issues.

neophyte57: Obviously this fix makes sense, but how do you reproduce the error ... is it only if the value of the `site.Id` isn't in the `Id` field of the `BusinessLicence` table?

neophyte57 (approved):
Good catch!

PR #2255, opened 2023-01-09
PRIME-2340 do not email list

yousiefc:
- Added table DoNotEmail to database
- Send function now checks against this table before completing


prime-dotnet-webapi/Models/DoNotEmail.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Prime.Models
+{
+    [Table("DoNotEmail")]
+    public class DoNotEmail : BaseAuditable
+    {
+        [Key]
+        public int Id { get; set; }
+        public string Email { get; set; }
+        public DateTimeOffset ModifiedDate { get; set; }
```
bergomi02, line 13: Since the final table structure already include audit fields (including UpdatedTimeStamp), I think ModifiedDate is not needed.

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -390,6 +393,19 @@ public async Task SendEnrolleeAbsenceNotificationEmailAsync(int enrolleeId, Enro
...
+            var doNotEmail = await _context.DoNotEmail
+                .Where(e => e.Email == string.Join(",", email.To))
+                .Select(e => new {
+                    e.Email,
+                    e.Id
+                })
+                .SingleOrDefaultAsync();
+
+            if (doNotEmail != null) {
+                await _businessEventService.CreateEmailEventAsync(doNotEmail.Id, "Do not email enrollee");
```
bergomi02, line 405: The first parameter of the CreateEmailEventAsync function is for EnrolleeId. 
image
Since we don't have EnrolleeId in this case, we might just pass 0, or make the datatype nullable int and pass Null. I know that the column EnrolleeId in BusinessEvent table is nullable.

bergomi02 (commented):
Tested in PR environment and the feature works as described in the JIRA ticket. Just couple items in the code need to be addressed.

prime-dotnet-webapi/Services/EmailService.cs
yousiefc: Why would the enrolleeId not be added to the table? Maybe I am misunderstanding how the BusinessEvent is utilized but don't we need the ID to know which user to log the event for?

bergomi02: Good question. The core function "Send(email)" does not know where calling it. It just sends out email. It could be enrollee sending out email to the vendor to set up new account or Admin sending email to enrollee or the system sending out email to enrollees for renewal. I think using CreateEmailEventAsync() is the correct choice to log the event. However, we don't have an Enrollee ID for every situation. About auditing, the system should have the UserID who triggered the event to store in the CreatedUserId of the BusinessEvent Table. Hope that makes sense.

yousiefc: I understand now that makes sense. Thanks for clarifying:)

bergomi02:
@yousiefc 

After some testing, the business event function throws an exception because there is no enrollee ID zero in the system. For that, I think you better off create a new function to log the event without an enrollee ID. Couple more things.
 
I forgot to test this case at the first round of testing. I put "test@email.com" in the do-not-email list and tried sending email to "Test@email.com" (with upper case "T") and the email was able to send out. It should not. Probably check for space as well.  

Alan told me when I joined the project that we need to keep one database migration per PR. Sorry to point it out now. I believe you know how to reverse the database migrations locally and re-generate the migration again. (let me or Alan know if you don't). For PR database, you can remove the PR database and let GitHub to rebuild it when you commit your new code change. You might have trouble deleting the PR database. If so, you can try to close the PR and delete the PR database, then re-open the PR. That should work. It is a pain but good practice to follow. Please let me or Alan know if you have any question.

yousiefc:
@bergomi02 

Can you please take a look at this I am not sure why the build is failing. The error message is not telling me much, at least not that I understand.

Also I was able to reopen this PR please ignore the other one that is now closed.

bergomi02:
Just doing some testing and I got the following error when trying to send out an email:
image

yousiefc:
> Just doing some testing and I got the following error when trying to send out an email: image

hmm I didn't get this when I tried. I will check it out.

prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -390,6 +393,19 @@ public async Task SendEnrolleeAbsenceNotificationEmailAsync(int enrolleeId, Enro
...
+            var doNotEmail = await _context.DoNotEmail
+                .Where(e => String.Equals(e.Email, String.Join(",", email.To), StringComparison.OrdinalIgnoreCase))
+                .Select(e => new {
+                    e.Email,
+                    e.Id
+                })
+                .SingleOrDefaultAsync();
+
+            if (doNotEmail != null) {
+                await _businessEventService.CreateEmailEventAsync("Do not email enrollee");
```
bergomi02, line 405: Sorry, minor change. Can you change the text in the business event? since the email might not be send to enrolle, "Do not email enrollee" does not make sense. Also, just an idea, to push the to-email-address in the event and indicate that email to this email address has been blocked by do-not-email feature. If someone down the road investigates why an email is not sending out, the business event log can show what email address has been blocked. Hope that makes sense.

yousiefc: I think I see what you mean. I have also fixed the case insensitive comparison and was able to extensively test it locally, I will push that up with this change as well.

bergomi02:
@yousiefc 
In my testing the email address that being block was not recorded in the event log propertly.
image

yousiefc:
> @yousiefc In my testing the email address that being block was not recorded in the event log propertly. image

I have pushed the fix It appears to be working.
donotemailevent

bergomi02 (approved):
Finally, it is good!

PR #2256, opened 2023-01-19
Fix/prime 2340 do not email list

yousiefc:
Added table DoNotEmail to database
Send function now checks against this table before completing

*note: attempted to close pr to delete pr database but for some reason am unable to reopen it


yousiefc:
@bergomi02 
could you take a look at this? I am not sure why it continues to fail.
I tried to close the original pr to delete the db but it would not allow me to reopen it but I was having the same issue there.

PR #2259, opened 2023-01-24
PRIME-2383 server error upon enrolment

yousiefc:


bergomi02:
The code change seems to fix the issue, which is great. During the code review, noticed the wkhtmltopdf file has been changed, please revert the change on that file.

yousiefc:
> The code change seems to fix the issue, which is great. During the code review, noticed the wkhtmltopdf file has been changed, please revert the change on that file.  

Thanks for reviewing. I'll change the file back.

bergomi02 (approved):
LGTM! Good job!

PR #2261, opened 2023-01-26
PRIME-2332 Vendor selection limited by care type (Health Authority Site Registration)

neophyte57:
- `HealthAuthorityService.GetVendorsByCareTypeAsync` not needed at this time but did not remove 
- Several Health Authority pages on admin side are broken by this model change so relying on data load from Excel spreadsheet to populate database, for time being
- To test this PR, you will need to populate a couple of tables with data, e.g. 

image

image


bergomi02:
Got the following error when running the databaes update:
image
The error is due to the existing data in the database. Except Production, I believe other environments have some data already. It is very tricky to fix.

neophyte57:
> Got the following error when running the databaes update: image The error is due to the existing data in the database. Except Production, I believe other environments have some data already. It is very tricky to fix.

It seems any existing data in `HealthAuthorityVendor` needs to be re-created in some way as it isn't associated to the correct Care Type.  I could add a SQL statement to the `Up` method to truncate the table prior to other schema changes ...

bergomi02:
> > Got the following error when running the databaes update: image The error is due to the existing data in the database. Except Production, I believe other environments have some data already. It is very tricky to fix.
> 
> It seems any existing data in `HealthAuthorityVendor` needs to be re-created in some way as it isn't associated to the correct Care Type. I could add a SQL statement to the `Up` method to truncate the table prior to other schema changes ...

Hm, HealthAuthoritySite and HealthAuthorityTechicalSupportVendor tables have reference to HealthAuthoityVendor table. 
image

We might need to remove all site data, include some HA data (HealthAuthoityVendor & HealthAuthorityTechicalSupportVendor) in order to apply the schema change. Also, how do you deal with the mapping between technical support and vendor? if HealthAuthorityVendor.HealthAuthorityOrganizationId change to  HealthAuthorityVendor.HealthAuthorityCareTypeId, the technical support will map to multiple care-type-vendor records? The site data from Interior Health and Northern Health, they only have one technical support for their sites.

bergomi02:
Closing PR as PR-2276 replaced this one.

PR #2267, opened 2023-02-03
Updating build-develop git action to use docker cmd

emuthiah:
Updating build-develop git action to use docker cmd


neophyte57 (approved):
Let's give it a try for `dev`  👍

PR #2272, opened 2023-02-10
Using docker cmd in github actions

emuthiah:


neophyte57 (approved):
Looks good to try in production  🤞

PR #2278, opened 2023-02-16
Updating startup probe

emuthiah:


neophyte57 (approved):
Let's give it a try in `dev`  👍

PR #2283, opened 2023-02-27
PRIME-2395 preferred last name only A-Z

yousiefc:
- Added validations for preferred last name forms


bergomi02 (approved):
LGTM! Good job!

PR #2293, opened 2023-03-10
PRIME-2418 Backend Server Error when Remote User is added

bergomi02:
http://pr-2293.pharmanetenrolment.gov.bc.ca/site

Note:
- update condition before sending email. Site should be in completed state, meaning the organization agreement is signed. 
- add new debugger launch with Edge for angular frontend, alternative to Chrome
- update frontend validation to allow deferring the business licence for PCHP when re-submitting a site.


neophyte57 (approved):
Good fixes!

PR #2298, opened 2023-03-17
PRIME-2426 new API to update Party HPDID from KeyCloak

bergomi02:
To Test:
- used prime-application-service-account client ID to get the token
- then call http://localhost:5000/api/parties/signing-authorities/updatehpdid


prime-dotnet-webapi/Controllers/SingingAuthoritiesExtendedController.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using Prime.Configuration.Auth;
+using Prime.Services;
+
+namespace Prime.Controllers
+{
+    [Produces("application/json")]
+    [Route("api/parties/signing-authorities")]
+    [ApiController]
+    public class SigningAuthorityExtendedController : PrimeControllerBase
```
neophyte57, line 14: Typo in **filename**

prime-dotnet-webapi/Services/PartyService.cs
```diff
@@ -201,5 +206,40 @@ private IQueryable<Party> GetBasePartyQuery()
...
+                .Take(limit)
+                .ToListAsync();
+
+            var counter = 0;
+            foreach (var pId in partyIds)
+            {
+                //for each party ID, get the party record
+                var party = await _context.Parties.Where(p => p.Id == pId).FirstOrDefaultAsync();
+                //get the user object from keycloak
+                var party_user = await _keycloakClient.GetUser(party.UserId);
```
neophyte57, line 226: This needs to handle the case where cannot find user in Keycloak ... log message and skip to next?

Also variable names should use Camel case, as per rest of prime-dotnet-webapi code base

prime-dotnet-webapi/Controllers/SigningAuthoritiesExtendedController.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using Prime.Configuration.Auth;
+using Prime.Services;
+
+namespace Prime.Controllers
+{
+    [Produces("application/json")]
+    [Route("api/parties/signing-authorities")]
+    [ApiController]
+    public class SigningAuthoritiesExtendedController : PrimeControllerBase
```
neophyte57, line 14: I'm not sure the reason for the new controller?  If it has to do with `Authorize` attribute, an example with lots of different `Authorize` configurations is `ProvisionerAccessController`

prime-dotnet-webapi/Services/PartyService.cs
```diff
@@ -8,6 +8,8 @@
 
 using AutoMapper;
 
+using Prime.HttpClients;
+using Prime.HttpClients.KeycloakApiDefinitions;
```
neophyte57, line 12: `using` looks unused  😊

prime-dotnet-webapi/Services/PartyService.cs
```diff
@@ -201,5 +206,40 @@ private IQueryable<Party> GetBasePartyQuery()
...
                 .Include(p => p.PartyCertifications);
         }
+
+        public async Task<int> UpdatePartyHpdid(int limit)
+        {
+            //query party ID where HPDID is null
+            var partyIds = await _context.Parties
+                .Where(p => p.HPDID == null)
+                .OrderBy(p => p.Id)
+                .Select(p => p.Id)
```
neophyte57, line 216: Can't we just get `UserId` here directly?

prime-dotnet-webapi/Services/PartyService.cs
```diff
@@ -201,5 +206,40 @@ private IQueryable<Party> GetBasePartyQuery()
...
+                ++counter;
+            }
+
+            try
+            {
+                await _context.SaveChangesAsync();
+            }
+            catch (DbUpdateConcurrencyException)
+            {
+                return -1;
```
neophyte57, line 239: Would `0` be more accurate?

prime-dotnet-webapi/Controllers/SigningAuthoritiesExtendedController.cs
bergomi02: The existing SigningAuthoritiesController controls the access at the class level, unlike ProvisionerAccessController which at each action/function level. Instead of updating SigningAuthoritiesController `Authorize` to action/function level, adding a new one is better as we could remove the class all-together once it is no longer needed.

prime-dotnet-webapi/Services/PartyService.cs
bergomi02: reference removed.

prime-dotnet-webapi/Services/PartyService.cs
bergomi02: if exception was thrown, -1 is a better indication of something went wrong. Anyhow, updated the code to log the error as well.

prime-dotnet-webapi/Services/PartyService.cs
bergomi02: Good point. Updated the code accordingly.

prime-dotnet-webapi/Controllers/SigningAuthoritiesExtendedController.cs
neophyte57: I think the `Authorize` can be overridden at the action/function-level ... see `EnrolleesExtendedController` e.g.

bergomi02: hm.. I just tried to put the new function to the `SigningAuthoriesController`, without changing the class level `Authorize` attribute, and the API call got 403 error. Then I removed the class `Authorize` attribute and the call went through. It seems the class level `Authorize` attribute is the base which applied to all functions. The function level ones are the additional rules. looking at the `EnrolleeExtendedController` `Authorize` attribute, the rule doesn't make much sense. Let me dig a bit deeper...

bergomi02: this explains all:
image

bergomi02: Anyway, the current implementation is the best suit for this case in my opinion.

prime-dotnet-webapi/Services/PartyService.cs
```diff
@@ -201,5 +205,49 @@ private IQueryable<Party> GetBasePartyQuery()
...
+            var counter = 0;
+            foreach (var pId in partyIds)
+            {
+                //for each party ID, get the party record
+                var party = await _context.Parties.Where(p => p.Id == pId).FirstOrDefaultAsync();
+                //get the user object from keycloak
+                var partyUser = await _keycloakClient.GetUser(party.UserId);
+                if (partyUser == null)
+                {
+                    _logger.LogError($"Party user ID '{party.UserId}' not found in keycloak.");
```
neophyte57, line 228: Please capitalize "keycloak"

neophyte57 (approved):
LGTM, good idea to expose as API!

PR #2312, opened 2023-04-10
PRIME-2458 Error Popup on Provisioner Link page

bergomi02:
- refactor a bit on the way the code pull values from token
- update getAdmin side to match getUser


neophyte57 (approved):
Nice job, much more elegant ... Spring Cleaning!  🧹

PR #2327, opened 2023-05-11
PRIME-2481 Update Label from "Preferred [name]" to "Alternate [name]"

bergomi02:


neophyte57 (approved):
Nice job, not an easy one IMO

PR #2339, opened 2023-05-26
PRIME-2424 - update v1beta to v1 per openshift 4.12 upgrade

wilson496:


infrastructure/cron-jobs/rerun-enrollee-application-rules.cron.yml
```diff
@@ -8,83 +8,75 @@ metadata:
...
+    jobTemplate:
+      spec:
+        template:
+          spec:
+            containers:
+            - command:
+              - bash
+              - '-c'
+              - >-
+                echo -e "-------- STARTING CRON --------\n" #magic___^_^___line echo -e "-------- Getting access_token --------\n" #magic___^_^___line TOKEN=$(curl -X POST "${KEYCLOAK_URL}/realms/${KEYCLOAK_REALM}/protocol/openid-connect/token" \
```
neophyte57, line 26: What is this about `#magic___^_^___line`?

wilson496: I'm not sure, and it looks like this job isn't currently deployed. Let's leave this for the time being.

PR #2341, opened 2023-05-31
PRIME-2477 Add consent to share GPID screen

bergomi02:


bergomi02:
Requirement changed. Code change here is no longer meet the requirement.

PR #2346, opened 2023-06-09
PRIME-2472 Add Social Workers

bergomi02:


neophyte57:
Although forced to Manual Review, TOA should be automatically set to OBO 

image

bergomi02:
> @bergomi02 Although forced to Manual Review, TOA should be automatically set to OBO
> 
> image

As discussed in the refinement, Admin will select the proper TOA for now (There is other ticket for defaulting TOA for all license class).

neophyte57 (approved):
LGTM (some clean-up will be needed in another PR re: blank practRefIds)

PR #2364, opened 2023-07-24
PRIME-2526 Issue with OrgBook searches

neophyte57:


bergomi02 (approved):
LGTM!

FYI, in Product, I noticed that the organization records with registration ID have been updated after created (the updated date is greater than the created date). I tested locally the registration ID would be populated on update but not on creation, if that makes sense. I believe this PR should fix the problem.

neophyte57:
> LGTM!
> 
> FYI, in Product, I noticed that the organization records with registration ID have been updated after created (the updated date is greater than the created date). I tested locally the registration ID would be populated on update but not on creation, if that makes sense. I believe this PR should fix the problem.

I see, good observation and testing @bergomi02 ... didn't realize there was a small window to edit Organization (before submitting a site)

PR #2374, opened 2023-08-23
PRIME-2526 Remove debugging

neophyte57:
Debugging no longer needed as fix appears to be working.  Remove to reduce rows/noise in `ClientLog` table


bergomi02 (approved):
Good to go!

PR #2377, opened 2023-08-24
PRIME-2371 Add non-BC college licensing option to enrolment

yousiefc:
Included the feature of adding an out of province college license from paper enrolment to be included in the normal enrolment.


neophyte57:
FYI, (1) if this situation

image

Clicking "Save and Continue" should not take user to the Site Information screen.

(2) Going "Back", if out-of-province college licence information has been entered, the information should be shown, not hidden by "disabled" toggle state

neophyte57:
Good, "Other College" info saved 
image

But validation on Review screen needs to be adjusted 
image

image

neophyte57:
Not the solution we're looking for

PR #2380, opened 2023-09-07
PRIME-2564 Signed Org Agreement can use Legal Name not part of Site Submission

neophyte57:


bergomi02:
Getting server error when updating organization name. 
image

same error when saving organization name for a new organization.
image

neophyte57:
> Getting server error when updating organization name. image
> 
> same error when saving organization name for a new organization. image


For New organization, should be fixed.  Not sure about former scenario, but it works for me  🤔

bergomi02:
> > Getting server error when updating organization name. image
> > same error when saving organization name for a new organization. image
> 
> For New organization, should be fixed. Not sure about former scenario, but it works for me 🤔

Oh, for the first one, I was trying to reproduce the scenario where the first site is in review status and with org. agreement done then try to update organization name. I replicate the scenario in this PR with Test Card 1. If you login with Test Card 1 and try to update the org. name, you should see the error.

PR #2384, opened 2023-09-22
PRIME-2562 Resolve hostname `primedatasvc.hibc.gov.bc.ca`

neophyte57:
For some reason, the previously used image `public.ecr.aws/h0h9t7p1/farma-retriever:12` could not resolve `primedatasvc.hibc.gov.bc.ca`:  `curl` would fail with `Could not resolve host:  primedatasvc.hibc.gov.bc.ca`

Secondly, this image wasn't under the PRIME team's control.  

Therefore, the change is to use a new Docker image that resolves `primedatasvc.hibc.gov.bc.ca`

FYI, this is already deployed to Production.


bergomi02 (approved):
Approved as already moved to Prod.

rshourou (approved):

PR #2385, opened 2023-09-22
PRIME-2477 Add Enrollee Consent question for auto pull

bergomi02:


neophyte57:
Looking good so far.  Should review with the Business ASAP before we go further.  They may have some thoughts on the titles/wording on the Care Setting screen and Enrolment Review screen, workflow, etc.

bergomi02:
The requirement has changed and, the logic implemented here is no longer valid. Closing PR.

PR #2389, opened 2023-09-27
PRIME-2542 Update Organization Agreement (CP & PCHP) Page Break

bergomi02:


neophyte57:
I think the Business wants to avoid breaks like this where heading is on one page by itself and body is on following page:  

image

bergomi02:
> I think the Business wants to avoid breaks like this where heading is on one page by itself and body is on following page:
> 
> image

The page break is different than what in local environment. Use PR environment as a guide and adjust the page break, it should be good now. I signed in as PRIME001 and created two sites. You can look at them download version from them.

PR #2400, opened 2023-11-06
PRIME-2592 UI Update - Multiple Organization Support

bergomi02:
https://pr-2400.pharmanetenrolment.gov.bc.ca/site

### DIFF
--- a/prime-angular-frontend/src/app/modules/site-registration/pages/site-management-page/site-management-page.component.ts
+++ b/prime-angular-frontend/src/app/modules/site-registration/pages/site-management-page/site-management-page.component.ts
@@ -232,7 +232,7 @@ export class SiteManagementPageComponent implements OnInit {
                 : expiryDates;
             }, []);
           this.organizations = organizations;
-          return this.organization = organizations.find((org) => org.id === organization.id);
+          return this.organization = organizations.find((org) => org.id === this.organizationService.organization.id);
         }),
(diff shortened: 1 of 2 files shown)


prime-angular-frontend/src/app/modules/site-registration/pages/site-management-page/site-management-page.component.ts
```diff
@@ -232,7 +232,7 @@ export class SiteManagementPageComponent implements OnInit {
                 : expiryDates;
             }, []);
           this.organizations = organizations;
-          return this.organization = organizations.find((org) => org.id === organization.id);
+          return this.organization = organizations.find((org) => org.id === this.organizationService.organization.id);
```
neophyte57, line 235: Can you add a comment re: why this is used instead of previous code?  That is, about preserving selected Organization when navigating to Site Management page?

neophyte57 (approved):
Looks good

prime-angular-frontend/src/app/modules/site-registration/pages/site-management-page/site-management-page.component.ts
bergomi02: Yes, I was testing this section of code and it was not right. I updated it to get rid of the input org id and always use the organization in org-service. 
Also, I updated the UI for single organization. It was showing the empty dropdown menu.

PR #2414, opened 2023-12-09
PRIME-2386 User Preferred Names in Emails

bergomi02:


prime-dotnet-webapi/Services/EmailService.cs
```diff
@@ -64,8 +64,8 @@ public async Task SendProvisionerLinkAsync(string[] emails, EnrolmentCertificate
                 .Where(e => e.Id == token.EnrolleeId)
                 .Select(e => new
                 {
-                    e.FirstName,
-                    e.LastName,
+                    FirstName = e.PreferredFirstName ?? e.FirstName,
+                    LastName = e.PreferredLastName ?? e.LastName,
```
neophyte57, line 68: Should this be refactored into common functions or property?

bergomi02: Sure, but don't know sure what the property name should be. Any idea?

neophyte57: What do you think of `Current`?

bergomi02: The code is updated.

neophyte57 (approved):
LGTM!

PR #2435, opened 2024-01-18
PRIME-2628 Signing Authority can't upload Business Licence when the site business licence has been expired

bergomi02:


neophyte57:
This seems like it partially fixes the problem but if:
1. I upload a file 
2. Toggle "This business licence doesn’t have an expiry date" back and forth 
3. Will see the following

image

bergomi02:
> This seems like it partially fixes the problem but if:
> 
> 1. I upload a file
> 2. Toggle "This business licence doesn’t have an expiry date" back and forth
> 3. Will see the following
> 
> image
> 
> In `ClientLog` it says "NG0100: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value for 'ng-valid': 'true'. Current value: 'false'.. Find more at https://angular.io/errors/NG0100"

it is fixed.

bergomi02:
revisit the issue with new solution. closing this PR as no longer needed.

PR #2438, opened 2024-01-24
PRIME-2640 Fix scary-looking production `apply-oc-template` error

neophyte57:


bergomi02 (approved):
Lets try this to remove the error message.

PR #2441, opened 2024-01-30
PRIME-2647 Tweak current wording of "Only physicians and nurse practitioners are permitted to access PharmaNet remotely"

neophyte57:


bergomi02 (approved):
LGTM!!

PR #2445, opened 2024-02-08
PRIME-1620 Update Enrolment Remote User entry

bergomi02:


neophyte57:
Looks good so far, but there is a bug where Vendor information is missing on the Enrolment review screen.  Although technically outside scope of this ticket and unrelated to the changes, it is worth fixing with this PR

image

bergomi02:
> Looks good so far, but there is a bug where Vendor information is missing on the Enrolment review screen. Although technically outside scope of this ticket and unrelated to the changes, it is worth fixing with this PR
> 
> image

How to reproduce your screen. Here is what I see for this enrollee:
image

neophyte57:
> > Looks good so far, but there is a bug where Vendor information is missing on the Enrolment review screen. Although technically outside scope of this ticket and unrelated to the changes, it is worth fixing with this PR
> > image
> 
> How to reproduce your screen. Here is what I see for this enrollee: image

I think it only happens first-time through?  I reproduced for another enrollee

image

bergomi02:
> > > Looks good so far, but there is a bug where Vendor information is missing on the Enrolment review screen. Although technically outside scope of this ticket and unrelated to the changes, it is worth fixing with this PR
> > > image
> > 
> > 
> > How to reproduce your screen. Here is what I see for this enrollee: image
> 
> I think it only happens first-time through? I reproduced for another enrollee
> 
> image

it should be fixed.

neophyte57:
Can either College Licence or PharmaNet ID/Prescriber ID be required?

image

bergomi02:
> Can either College Licence or PharmaNet ID/Prescriber ID be required, at least on the front-end?
> 
> image

Good catch. it is fixed.

neophyte57:
In `dev` you can toggle the "Do you have a ..." control endlessly 

image

but in this PR, when you toggle on and then off, the form control disappears

bergomi02:
> In `dev` you can toggle the "Do you have a ..." control endlessly
> 
> image
> 
> but in this PR, when you toggle on and then off, the form control disappears

This should be fixed.

prime-angular-frontend/src/app/shared/components/forms/college-certification-form/college-certification-form.component.ts
```diff
@@ -278,8 +289,9 @@ export class CollegeCertificationFormComponent implements OnInit {
     } else {
       licenseNumberValidators.push(FormControlValidators.alphanumeric);
     }
-    this.formUtilsService.setValidators(this.licenseNumber, licenseNumberValidators);
-
+    if (this.collegeCode.value === CollegeLicenceClassEnum.CPSBC) {
+      this.formUtilsService.setValidators(this.licenseNumber, licenseNumberValidators);
+    }
```
neophyte57, line 294: Why are the validators set only when CPSBC?

bergomi02: Good point. The logic is updated so that it will add the validation to the license number control only if it is not in condensed mode and the college code is BCCNM - because the license number is not showing for BCCNM in remote access mode .

bergomi02:
@neophyte57 
As discussed before, the following changes have been implemented.
- add College to the remote access user card
- remove field that does not have a value in the remote access user section and the college license section 
- change the license number label of the remote access user section and the college license section depends on the license college - under individual enrolment, site registration, admin page.

neophyte57 (approved):
OMG 😅  That was not to be underestimated

PR #2453, opened 2024-02-23
PRIME-2645 fix enrolment overview page

bergomi02:


neophyte57 (approved):
Nicely done!

neophyte57:
Actually I tried something else, editing a change and when I return to Overview page, the change is **not** reflected there, e.g. Optional Extension Number

PR #2462, opened 2024-03-05
PRIME-2679 Email Template Update

bergomi02:
https://pr-2462.pharmanetenrolment.gov.bc.ca/admin


neophyte57:
Try to update database in PR environment

neophyte57:
Something strange re: "Last Modified" date on the email template cards.  In the PR environment, only the two updated templates have a date other than "16 Sep 2019".  In `dev`, all have "Last Modified" of "16 Sep 2019".  Whereas `test` and my local environment is actually as expected with a mix of "Last Modified" dates.  Why is that?

prime-dotnet-webapi/Migrations/20240318232909_EmailTemplateUpdate.cs
```diff
@@ -0,0 +1,72 @@
...
+using Npgsql.EntityFrameworkCore.PostgreSQL.Metadata;
+
+namespace Prime.Migrations
+{
+    public partial class EmailTemplateUpdate : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropTable(
+                name: "PharmanetTransactionLogTemp");
```
neophyte57, line 12: This snuck in?

bergomi02: Not sure why. I re-generated the migration and it did the same. For now, I removed the code manually for now.

prime-dotnet-webapi/Migrations/ApiDbContextModelSnapshot.cs
```diff
@@ -12059,60 +12059,6 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.ToTable("PharmanetTransactionLog");
                 });
 
-            modelBuilder.Entity("Prime.Models.PharmanetTransactionLogTemp", b =>
```
neophyte57, line 12062: Shouldn't `PharmanetTransactionLogTemp` remain in `ApiDbContextModelSnapshot`?

bergomi02: Good catch. Not sure how it got removed from the snapshot. I added it back manually. It should work fine now.

neophyte57 (approved):
LGTM

PR #2466, opened 2024-03-12
PRIME-2410 Add PharmanetTranscationLogTemp table

bergomi02:
This PR is used to create a temp table to store a smaller set of transaction log data for reporting. Hoping to fix the some report timeout issue.


prime-dotnet-webapi/Migrations/20240312233918_AddPharmanetTransactionLogTemp.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Prime.Migrations
+{
+    public partial class AddPharmanetTransactionLogTemp : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+
+        }
```
neophyte57, line 10: Shouldn't there be something in the `Up` method?

bergomi02: LOL, what am i doing? it is fixed. But the database is not up for this PR...hm

PR #2473, opened 2024-03-26
PRIME-2693 Problem viewing previous enrolment as admin

neophyte57:


bergomi02 (approved):
Good job fixing the issue. Look good!

PR #2479, opened 2024-03-28
pull mailhog-server image from OCP imageStream

rshourou:


neophyte57 (approved):
Thanks @rshourou for correcting this  🙏

PR #2480, opened 2024-03-28
PRIME-2410 Add index to Transaction Log temp table and add a new job to populate the table

bergomi02:


prime-dotnet-webapi/Controllers/JobsController.cs
```diff
@@ -52,5 +52,26 @@ public async Task<ActionResult> UpdatePractitionerTable()
...
+        [HttpPost("populate/transaction-log-temp", Name = nameof(PopulateTransactionLogTemp))]
+        [Authorize(Roles = Roles.PrimeApiServiceAccount)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        public async Task<ActionResult> PopulateTransactionLogTemp(int numberOfDays)
+        {
+            if (numberOfDays <= 0)
+            {
+                return Forbid();
```
neophyte57, line 70: `BadRequest()` is more appropriate?

prime-dotnet-webapi/Services/ReportingService.cs
```diff
@@ -121,5 +122,39 @@ public async Task<string> UpdatePractitionerTableAsync()
...
+            copySql.Append("\"TransactionSubType\", \"PractitionerId\", \"CollegePrefix\", \"TransactionOutcome\", \"ProviderSoftwareId\", ");
+            copySql.Append("\"ProviderSoftwareVersion\", \"LocationIpAddress\", \"SourceIpAddress\"");
+            copySql.Append("from \"PharmanetTransactionLog\" ptl ");
+            copySql.Append($"where \"TxDateTime\" > current_date + interval '-{numberInDays}' day ");
+
+            // if there is max transaction Id, use it in the where clause
+            if (maxTransactionId.HasValue)
+            {
+                copySql.Append($"and \"TransactionId\" > '{maxTransactionId}' ");
+            }
```
neophyte57, line 153: What's the intent of this?  What problem does it avoid/solve?

prime-dotnet-webapi/Controllers/JobsController.cs
bergomi02: Updated.

prime-dotnet-webapi/Services/ReportingService.cs
bergomi02: It needs a way to get the only missing records from the original log table - whatever missed from last job run. The best way (I can think of) is by comparing the last/max transaction ID from the temp. However, only if some records already in temp table. If not, skip that step and filter the records by date only.

neophyte57: Please add a comment re: the above, I guess for efficiency?
Also does the following help?
```
PTL  t1  XXXXXXXXXXXXXXXXXXXXX00000
     t2  XXXXXXXXXXXXXXXXXXXXXXX1111111

Temp t1                       00000
     t2                         0001111
```

neophyte57: Also it means to expand the contents of the PTL Temp table into the past, the PTL Temp table must be cleared first?  I think a comment re: that would be beneficial too

neophyte57 (approved):

prime-dotnet-webapi/Services/ReportingService.cs
bergomi02: More comment added.

neophyte57 (approved):
Improvement in `PopulateTransactionLogTemp` method comment

PR #2482, opened 2024-04-05
PRIME-2701 Expand Practitioner Table

bergomi02:
expand practitioner table

### DIFF
--- a/prime-dotnet-webapi/Models/Practitioner.cs
+++ b/prime-dotnet-webapi/Models/Practitioner.cs
@@ -16,9 +16,9 @@ namespace Prime.Models
         public string FirstName { get; set; }
         public string MiddleInitial { get; set; }
         public string LastName { get; set; }
-        public DateTime DateofBirth { get; set; }
+        public DateTime? DateofBirth { get; set; }
         public string Status { get; set; }
-        public DateTimeOffset EffectiveDate { get; set; }
+        public DateTimeOffset? EffectiveDate { get; set; }
         public DateTime? ProcessedDate { get; set; }
     }
 }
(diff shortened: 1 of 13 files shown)


prime-dotnet-webapi/Models/Practitioner.cs
```diff
@@ -11,15 +11,14 @@ public class Practitioner : BaseAuditable
...
         public string PracRefId { get; set; }
-
         public string FirstName { get; set; }
-
+        public string MiddleInitial { get; set; }
         public string LastName { get; set; }
-
+        public DateTime DateofBirth { get; set; }
+        public string Status { get; set; }
+        public DateTimeOffset EffectiveDate { get; set; }
```
neophyte57, line 21: Better as `DateTimeOffset?`

prime-dotnet-webapi/Models/Practitioner.cs
```diff
@@ -11,15 +11,14 @@ public class Practitioner : BaseAuditable
...
         public string CollegeId { get; set; }
-
         public string PracRefId { get; set; }
-
         public string FirstName { get; set; }
-
+        public string MiddleInitial { get; set; }
         public string LastName { get; set; }
-
+        public DateTime DateofBirth { get; set; }
```
neophyte57, line 19: Better as `DateTime?`

prime-dotnet-webapi/Models/Practitioner.cs
bergomi02: Good idea. It is updated.

prime-dotnet-webapi/Models/Practitioner.cs
bergomi02: It is updated.

neophyte57 (approved):
LGTM!

PR #2492, opened 2024-04-24
PRIME-2681 default upper case for PEC/Site ID

bergomi02:


neophyte57:
Looks like missing during HA Site Registration

image

bergomi02:
> Looks like missing during HA Site Registration
> 
> image

It is fixed.

neophyte57 (approved):
😊👍

PR #2502, opened 2024-05-06
PRIME-2528 Adjust what are acceptable characters in Alternative Name fields

neophyte57:


bergomi02 (approved):
LGTM. Thanks for simplifying the expression.

PR #2506, opened 2024-05-09
PRIME-2706 Mismatch Missing Business License flag in Admin

bergomi02:


neophyte57:
@bergomi02  I'm trying to understand what this PR fixes ... previously there was a bug that enabled CP Site Registrations to be created with No Business Licence but a Business Licence Expiry Date that was set?  Is there an instance of such data in the `dev` environment?

bergomi02:
> @bergomi02 I'm trying to understand what this PR fixes ... previously there was a bug that enabled CP Site Registrations to be created with No Business Licence but a Business Licence Expiry Date that was set? Is there an instance of such data in the `dev` environment?

This should fix any discrepancy between the two pages. Here is the example in DEV:
image
For "Site C", the missing business license is marked "Yes".
image
in the overview page, it marked as "No".

prime-angular-frontend/src/app/modules/adjudication/shared/components/site-registration-table/site-registration-table.component.ts
```diff
@@ -108,6 +108,22 @@ export class SiteRegistrationTableComponent implements OnInit, AfterViewInit {
     }
   }
 
+  public displayMissingBusinessLicence(row: SiteRegistrationListViewModel) {
```
neophyte57, line 111: Please add return type

neophyte57 (approved):
Looks good👍

prime-angular-frontend/src/app/modules/adjudication/shared/components/site-registration-table/site-registration-table.component.ts
bergomi02: return type added.

PR #2507, opened 2024-05-13
PRIME-2721 Remove note requirement when assigning adjudicator

bergomi02:


neophyte57:
This is still present 
image

bergomi02:
> This is still present image

Good catch. it is fixed

neophyte57 (approved):
👍

PR #2510, opened 2024-05-16
PRIME-2723 Send email to Connections for HA site approval

bergomi02:


neophyte57:
There should also be a Business Event when email is sent out (see https://bcgovmoh.atlassian.net/browse/PRIME-2177)

bergomi02:
> There should also be a Business Event when email is sent out (see https://bcgovmoh.atlassian.net/browse/PRIME-2177)

The business event is added. About the email content change, I will wait for the finalized version in the coming meeting.

neophyte57 (approved):
Nicely done @bergomi02 !

PR #2511, opened 2024-05-17
PRIME-2602 Add cron job template for refreshing materialized view

bergomi02:


neophyte57 (approved):
👍 

FYI @bergomi02 , I just adjusted where the image comes from after @rshourou put `python-curl-psql` into `9c33a9-tools` namespace.  Thanks both 😊

PR #2545, opened 2024-06-28
PRIME-2765 Show Licence Category in Licence Maintenance page

bergomi02:


prime-angular-frontend/src/app/modules/adjudication/pages/license-classes-maintenance-page/license-classes-maintenance-page.component.ts
```diff
@@ -69,9 +70,10 @@ export class LicenseClassesMaintenancePageComponent implements OnInit {
             collegeName: college.name,
             licenseCode: collegeLicense.licenseCode,
             discontinued: collegeLicense.discontinued,
+            collegeLicenseGroupingCode: license.collegeLicenses[0].collegeLicenseGroupingCode,
```
neophyte57, line 73: Is there a reason this can't be `collegeLicenseGroupingCode: collegeLicense.collegeLicenseGroupingCode,`?  Seems simpler

bergomi02: I just double checked. The "collegeLicense" is from college config which does not contain the collegeLicenseGroupingCode value (always null). Only license config contains the collegeLicenseGroupingCode value.

prime-angular-frontend/src/app/modules/adjudication/pages/license-classes-maintenance-page/license-classes-maintenance-page.component.ts
```diff
@@ -69,9 +70,10 @@ export class LicenseClassesMaintenancePageComponent implements OnInit {
             collegeName: college.name,
             licenseCode: collegeLicense.licenseCode,
             discontinued: collegeLicense.discontinued,
+            collegeLicenseGroupingCode: collegeLicense.collegeLicenseGroupingCode,
             ...license
           } as LicenseMaintenanceConfig;
-        }).sort(this.utilsService.sortByKey<LicenseMaintenanceConfig>('weight'))
+        }).sort(this.utilsService.sortByKey<LicenseMaintenanceConfig>('collegeName'))
```
neophyte57, line 76: It's unclear why there is a need to change ordering?

bergomi02: well, it is partially of my bad. The weight values are not set against whole list and some are all the same value, 10. Because of that, sorting by weight is wrong for sure. I tried to remove the sorting all together and it seems working alright. I will keep no sorting until business wait it to sort in a different way.

neophyte57: OK, yeah, the order has changed compared to `demo` but it is still grouped together logically 👍

neophyte57 (approved):
LGTM

prime-angular-frontend/src/app/modules/adjudication/pages/license-classes-maintenance-page/license-classes-maintenance-page.component.ts
neophyte57: Thanks for changing

PR #2564, opened 2024-08-12
PRIME-2767:  Stale TOA assignment if original SD questions

neophyte57:


bergomi02:
Tested, function as it intended. However, it is waiting for the last piece of content to be displayed.

bergomi02 (approved):
LGTM!

PR #2576, opened 2024-08-29
PRIME-2777 update Provider Team email 

bergomi02:


prime-dotnet-webapi/Configuration/PrimeConfigurationObjects.cs
```diff
@@ -108,4 +108,9 @@ public class SentryConfiguration
     {
         public string Dsn { get; set; }
     }
+
+    public class ProviderTeamConfiguration
```
neophyte57, line 112: I think this (and similar) should be "provider enrolment team"; less ambiguous, as precise as we know it

bergomi02: Updated as suggested.

neophyte57 (approved):
Nicely done!

PR #2596, opened 2024-09-27
Fix failure of build-deploy-backup-dev.yml

neophyte57:


bergomi02:
For some reason, this PR build hangs. I checked the build PR yml and there is reference to the delete yml file. 
image

Maybe it needs to be updated.

neophyte57:
> For some reason, this PR build hangs. I checked the build PR yml and there is reference to the delete yml file. image
> 
> Maybe it needs to be updated.

Good point!

bergomi02 (approved):
LGTM!

PR #2603, opened 2024-10-04
PRIME-2779 Do not sent "Site Active before Registration" email and do not create related event

bergomi02:


prime-dotnet-webapi/Controllers/SitesController.cs
```diff
@@ -945,8 +945,10 @@ public async Task<ActionResult> ApproveSite(int siteId)
 
                 if (communitySite.ActiveBeforeRegistration)
                 {
+                    /* do not send "Site Active Before registration email" to SA and dot not create related event
```
neophyte57, line 948: Please correct typo "dot"

bergomi02: it is fixed.

neophyte57 (approved):
LGTM

PR #2639, opened 2024-11-28
PRIME-2824 Site ID Duplicates icon did not show up correctly

bergomi02:


prime-dotnet-webapi/Services/CommunitySiteService.cs
```diff
@@ -108,6 +116,13 @@ private static void GroupSitesToOrgVisually(PaginatedList<CommunitySiteAdminList
             }
         }
 
+        private async Task<int> GetDuplicatePecCount(int? careSettingCode, string pec, int orgSiteId)
```
neophyte57, line 119: Why is this named `orgSiteId`?

bergomi02: oh, meant to be original site id (not PEC).

bergomi02: Updated the name to be clear

neophyte57 (approved):
Nice job!

PR #2671, opened 2025-02-06
PRIME-2877 bug fix - Licence class change should check and remove remote access when needed

bergomi02:


neophyte57:
In this scenario:
1. Start new enrollment  
2. Enter College Licence details matching Remote User from a site registration 
3. Select matching site after requesting Remote Access 
4. Continue past SD questions
5. At Overview, go back to edit College Licence, setting to None

Why do Remote Access details still remain?  

image

PR #2675, opened 2025-02-12
PRIME-2867 Use "promote" API to address CHES unreliability

neophyte57:
Added CHES_RETRY_WITHIN_HOURS to `ches` ConfigMap.

Switched CHES_ENABLED to `true` in `dev` environment.  Set `dev-send-renewal-emails` cron job to be suspended


bergomi02:
Since we have a way to resend the email, I wonder if we should improve the whole email status update process as well. For example, we let the email status update process to be executed first, however, make it only check the ones with empty status (currently it is checking any status other than "Completed", and some have been updated thousand+ times). After the email status update process is completed, run the resend process and make it to update the email log status to empty. so that the email status update process will pick them up to update the status in the next round. Just a thought. 

However, the current implementation works as expected.

bergomi02 (approved):
LGTM!

PR #2685, opened 2025-03-03
PRIME-2880 Fix able to remove OOP licence class when it should be required

neophyte57:


bergomi02 (approved):
Tested in local and fix the issue as expected. Good job!

PR #2735, opened 2025-06-25
OHID-344 Remote User Issues

bergomi02:
- add validation to avoid duplicate remote user entry in PCHP
- update API to skip duplicate remote user entry
- update the business logic that finds matching remote user from enrolment to use notified flag instead of site approved date. so that when signing authority tries to add more remote user and bring the site o review status, the enrollee is still able to add remote access.
- add the remote user created date in admin page and sort the remote user by created date in descending order


bergomi02:
Hi @rshourou, can you please approve this PR when you have a chance? going to push this change to Test for client to test the changes. Thanks~

rshourou (approved):
LGTM!

PR #2754, opened 2025-09-08
Remove code climate in CI/CD script

bergomi02:
As Code Climate API has been disabled back in July 2025, the existing Code Coverage script is no longer working and it blocks some PRs to move on to "develop" branch. This PR is to remove the Code Climate script in CI/CD workflow but we will lose the Code Coverage report. 

However, we are in process to add PRIME into SonarCloud which is another service for Code Coverage. Other PR will be created to add Code Coverage back to PRIME workflow.


rshourou (approved):
LGTM!

PR #2765, opened 2025-09-26
OHID-357 Improve business logic of HA Site Registration Passcode

bergomi02:
Use the passcode to default HA for new authorized user after the BCSC login.


rshourou (approved):
LGTM!

